PR #4, opened 2024-02-26
electrumx: fix race condition in client

joshuasing:
**Summary**
Fix a race condition in the electrumx client

Previously, the request ID stored in the client was incremented while the lock was held, however it was then read without the lock being held.

**Changes**
- Fix race condition in `electrumx` package


ClaytonNorthey92 (approved):

marcopeereboom (approved):
Are we sure we pre increment? seems odd.

joshuasing:
> Are we sure we pre increment? seems odd.

@marcopeereboom I am not sure, this was the previous behaviour except with the read in a lock.

PR #5, opened 2024-02-26
ci: fix go workflow succeeding even if a step fails

joshuasing:
**Summary**
Remove use of `continue-on-error` flag in Go workflow due to it causing the workflow to suceed even if a step failed.

**Changes**
- Replace `continue-on-error` with `if: (success() || failure()) && steps.deps.outcome == 'success'` to mimic behaviour without causing the job to succeed with steps failing.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -50,14 +50,14 @@ jobs:
           check-latest: true
 
       - name: "Download and verify dependencies"
+        id: deps
         run: make deps
 
       - name: "make race"
-        continue-on-error: true
         run: make race
 
       - name: "make"
-        continue-on-error: true
+        if: (success() || failure()) && steps.deps.outcome == 'success'
         env:
           PGTESTURI: "postgres://postgres:postgres@localhost:5432/postgres?sslmode=disable"
         run: |
@@ -65,6 +65,6 @@ jobs:
           git diff --exit-code
 
       - name: "make web popm"
-        continue-on-error: true
+        if: (success() || failure()) && steps.deps.outcome == 'success'
         run: |
           cd web && make


ClaytonNorthey92:
hey @joshuasing I am a little confused here; I _think_ I understand the issue we're trying to solve, however why not just leave it with the default? why do we need this `if` check? wouldn't the pipeline fail if `make deps` failed anyway?

joshuasing:
> hey @joshuasing I am a little confused here; I _think_ I understand the issue we're trying to solve, however why not just leave it with the default? why do we need this `if` check? wouldn't the pipeline fail if `make deps` failed anyway? as long as we didn't "continue on error"

@ClaytonNorthey92 I had added 'continue-on-error' so that we can collect error logs from each step instead of it exiting immediately when one step failed, aiming to allow us to collect extra information that could be useful for debugging.

I was unsure of whether continue-on-error would cause this bug at the time, as different parts of GitHub's documentation said completely different things.

ClaytonNorthey92:
> > hey @joshuasing I am a little confused here; I _think_ I understand the issue we're trying to solve, however why not just leave it with the default? why do we need this `if` check? wouldn't the pipeline fail if `make deps` failed anyway? as long as we didn't "continue on error"
> 
> @ClaytonNorthey92 I had added 'continue-on-error' so that we can collect error logs from each step instead of it exiting immediately when one step failed, aiming to allow us to collect extra information that could be useful for debugging.
> 
> I was unsure of whether continue-on-error would cause this bug at the time, as different parts of GitHub's documentation said completely different things.

@joshuasing OK that makes sense. Can we test this before merging? Perhaps push up a broken commit and make sure the pipeline fails correctly with this change? If so, then I approve ✅️

ClaytonNorthey92 (approved):
Let's drop the force-break test and merge!

marcopeereboom (approved):

PR #12, opened 2024-02-27
Automate Full Network E2E Acceptance Tests

ClaytonNorthey92:
**Summary**
Talking with @max-sanchez , we want to automate his acceptance test cases, these are:
* PoP transactions get published to Bitcoin, and each contains an OP_RETURN starting with 'HEMI"
* PoP transactions are using the correct fee (approx 1 sat/vB)
* PoP miner only creates one BTC transaction for each keystone
* PoP payouts are calculated for each keystone starting at 25, and payout address matches PoP miner ETH address
* Bitcoin Finality for blocks starts at -9 and progresses in step with new Bitcoin blocks
* Bitcoin Finality returns same result for last 10 blocks and querying for specific block in that list

**Changes**
Automates the above tests in a single "full network" test.  test cases are denoted by comment prefix: "acceptance test case"


service/bfg/bfg.go
```diff
@@ -271,6 +271,8 @@ func (s *Server) handleBitcoinBroadcast(ctx context.Context, bws *bfgWs, payload
 		tl2, err = pop.ParseTransactionL2FromOpReturn(v.PkScript)
 		if err != nil {
 			log.Errorf(err.Error()) // handle real error below
+		} else {
+			break
```
ClaytonNorthey92, line 275: this will likely be fixed in @marcopeereboom 's pr #10

jcvernaleo (changes requested):
Two things:

Could you rebase to address conflicts.

And you have tons of info in the github message, but none in the commit message.  The commit message should be the main source of info rather than the message on github.

ClaytonNorthey92:
> Two things:
> 
> Could you rebase to address conflicts.
> 
> And you have tons of info in the github message, but none in the commit message. The commit message should be the main source of info rather than the message on github.

hey @jcvernaleo - yup! will rebase, though I want to merge #18 first; it will simplify the e2e tests a bit.  And I can definitely make the commit message better 👍

ClaytonNorthey92:
@joshuasing @jcvernaleo I am actually going to close this in favor of #37 ; that's a better way to test, and we can write cleaner automated tests against the local network

PR #15, opened 2024-02-28
WIP: retry keystone on failure

### LINKED ISSUE #7: Implement "retry" when mining L2 Keystones in PoP Miner
### Confirmation

### Problem

As of right now, if pop miner fails to mine a keystone, it won't retry.  It's possible a one-off error occurs in this moment and a retry would succeed.

### Suggested solution

Implement some sort of retry for pop miner when "mining keystones".  This will require some discussion; the retry strategy, limit, etc. all need to be discussed.

### Additional Information

as of writing this, this refers to the `mineKeystone` function in `popm.go`

ClaytonNorthey92:
**Summary**
allowing keystone retries with a fifo queue

**Changes**
* if a keystone fails to mine, retry

* check if the keystone has already been mined

* allow a queue of 10 keystones, dropping the oldest if full and a new one comes in

fixes #7


service/popm/popm.go
```diff
@@ -50,6 +50,33 @@ func init() {
 	loggo.ConfigureLoggers(logLevel)
 }
 
+type CircularFifo struct {
+	buf chan hemi.L2Keystone
+	mtx sync.Mutex
```
ClaytonNorthey92, line 55: don't need this mutex

service/popm/popm.go
```diff
@@ -253,6 +279,33 @@ func createTx(l2Keystone *hemi.L2Keystone, btcHeight uint64, utxo *bfgapi.Bitcoi
 	return &btx, nil
 }
 
+func (m *Miner) shouldMineKeystone(ctx context.Context, ks *hemi.L2Keystone) (bool, error) {
+	serialized := hemi.L2KeystoneAbbreviate(*ks).Serialize()
+	response, err := m.callBFG(ctx, 5*time.Second, &bfgapi.PopTxsForL2BlockRequest{
```
ClaytonNorthey92, line 284: we can't do this from popm; it is on a different listener...find a different way

service/popm/popm.go
```diff
@@ -50,6 +50,63 @@ func init() {
...
+		requiresMine: true,
+	}
+
+	r.buf[r.write] = item
+	r.write = (r.write + 1) % uint(cap(r.buf))
+}
+
+func (r *CircularFifo) MineEach(cb func(ks hemi.L2Keystone) error) {
+	r.mtx.Lock()
+	copies := slices.Clone(r.buf)
```
ClaytonNorthey92, line 84: do oldest first

ClaytonNorthey92:
Closing until ready for review

PR #18, opened 2024-03-01
retry mine keystone on failure

### LINKED ISSUE #7: Implement "retry" when mining L2 Keystones in PoP Miner

ClaytonNorthey92:
**Summary**
We need a way to retry mining keystones if they fail, this provides a solution.

Fixes #7 

**Changes**
* keep buffer of 10 newest keystones known
* upon receiving a keystone, only insert into buffer if it is newer than the oldest keystone
* if buffer is full, drop oldest keystone upon inserting another
* when a keystone is mined, mark it as "processed" to not repeat mining same keystone
* generate fewer btc blocks in full network e2e test, since we have a retry strategy now 
* remove retry from full network e2e test

### DIFF
--- a/service/popm/popm.go
+++ b/service/popm/popm.go
@@ -127,7 +127,7 @@ func NewMiner(cfg *Config) (*Miner, error) {
 		holdoffTimeout: 5 * time.Second,
 		requestTimeout: 5 * time.Second,
 		mineNowCh:      make(chan struct{}, 1),
-		l2Keystones:    make(map[string]L2KeystoneProcessingContainer),
+		l2Keystones:    make(map[string]L2KeystoneProcessingContainer, l2KeystonesMaxSize),
 	}
 
 	switch strings.ToLower(cfg.BTCChainName) {
@@ -484,16 +484,32 @@ func (m *Miner) BitcoinUTXOs(ctx context.Context, scriptHash string) (*bfgapi.Bi
 func (m *Miner) mineKnownKeystones(ctx context.Context) {
 	keystonesFailed := false
 
-	m.ForEachL2Keystone(func(ks hemi.L2Keystone) error {
-		log.Infof("Received keystone for mining with height %v...", ks.L2BlockNumber)
-		if err := m.mineKeystone(ctx, &ks); err != nil {
+	copies := m.l2KeystonesForProcessing()
+
+	for _, e := range copies {
+		serialized := hemi.L2KeystoneAbbreviate(e).Serialize()
+		key := hex.EncodeToString(serialized[:])
+
+		log.Infof("Received keystone for mining with height %v...", e.L2BlockNumber)
+
+		err := m.mineKeystone(ctx, &e)
+		if err != nil {
 			log.Errorf("Failed to mine keystone: %v", err)
 			keystonesFailed = true
-			return err
 		}
 
-		return nil
-	})
+		m.mtx.Lock()
+
+		if v, ok := m.l2Keystones[key]; ok {
+			// if there is an error, mark keystone as "requires processing" so
+			// potentially gets retried, otherwise set this to false to
+			// nothing tries to process it
+			v.requiresProcessing = err != nil
+			m.l2Keystones[key] = v
+		}
+
+		m.mtx.Unlock()
+	}
 
 	// if no keystones failed, we're done
 	if !keystonesFailed {
@@ -905,8 +921,11 @@ func (m *Miner) AddL2Keystone(val hemi.L2Keystone) {
 	m.l2Keystones[key] = toInsert
 }
 
-func (m *Miner) ForEachL2Keystone(cb func(ks hemi.L2Keystone) error) {
+// l2KeystonesForProcessing creates copies of the l2 keystones, set them to
+// "processing", then returns the copies
+func (m *Miner) l2KeystonesForProcessing() []hemi.L2Keystone {
 	m.mtx.Lock()
+
 	copies := []hemi.L2Keystone{}
 	for i, v := range m.l2Keystones {
 
@@ -924,27 +943,9 @@ func (m *Miner) ForEachL2Keystone(cb func(ks hemi.L2Keystone) error) {
 	}
 	m.mtx.Unlock()
 
-	// mine the newest keystone first
 	slices.SortFunc(copies, func(a, b hemi.L2Keystone) int {
 		return int(b.L2BlockNumber) - int(a.L2BlockNumber)
 	})
 
-	for _, e := range copies {
-		serialized := hemi.L2KeystoneAbbreviate(e).Serialize()
-		key := hex.EncodeToString(serialized[:])
-
-		err := cb(e)
-
-		m.mtx.Lock()
-
-		if v, ok := m.l2Keystones[key]; ok {
-			// if there is an error, mark keystone as "requires processing" so
-			// potentially gets retried, otherwise set this to false to
-			// nothing tries to process it
-			v.requiresProcessing = err != nil
-			m.l2Keystones[key] = v
-		}
-
-		m.mtx.Unlock()
-	}
+	return copies
 }
--- a/service/popm/popm.go
+++ b/service/popm/popm.go
@@ -916,7 +916,6 @@ func (m *Miner) l2KeystonesForProcessing() []hemi.L2Keystone {
 	m.mtx.Lock()
 
 	for i, v := range m.l2Keystones {
-
 		// if we're currently processing, or we've already processed the keystone
 		// then don't process
 		if !v.requiresProcessing {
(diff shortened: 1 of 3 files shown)


service/popm/popm.go
```diff
@@ -50,6 +50,108 @@ func init() {
...
+	mtx     sync.Mutex
+	mapping map[string]*L2KeystonePriorityBufferElement
+	size    int
+}
+
+// L2KeystonePriorityBufferElement holds an L2Keystone and whether it has been
+// "processed" or not
+type L2KeystonePriorityBufferElement struct {
+	l2Keystone         hemi.L2Keystone
+	requiresProcessing bool
```
ClaytonNorthey92, line 66: hmm we might want to add a "currentlyProcessing" flag here as well; since this can be triggered by both timer and new keystone

service/popm/popm.go
```diff
@@ -50,6 +50,112 @@ func init() {
...
+func (r *L2KeystonePriorityBuffer) ForEach(cb func(ks hemi.L2Keystone) error) {
+	r.mtx.Lock()
+	copies := []hemi.L2Keystone{}
+	for _, v := range r.mapping {
+		if v.requiresProcessing {
+			copies = append(copies, v.l2Keystone)
+
+			// temporarily set this to false, so another goroutine doesn't
+			// try to process
+			v.requiresProcessing = false
```
marcopeereboom, line 134: This looks racy. You could have started traversing the map while the variable is not set yet. This is a logic race. Setting and unsetting the variable should really happen in the same lock.

service/popm/popm.go
```diff
@@ -50,6 +50,112 @@ func init() {
...
+	slices.SortFunc(copies, func(a, b hemi.L2Keystone) int {
+		return int(b.L2BlockNumber) - int(a.L2BlockNumber)
+	})
+
+	for _, e := range copies {
+		mined := hemi.L2KeystoneAbbreviate(e).Serialize()
+		key := hex.EncodeToString(mined[:])
+
+		err := cb(e)
+		r.mtx.Lock()
```
marcopeereboom, line 149: I didn't want to say this but this is probably where you want to serialize using a channel. I may be missing some subtlety so don't take this as gospel truth but once you start seeing locking/unlocking in the same loop it typically means the code should be rethought to prevent that from happening.

service/popm/popm_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	"fmt"
```
marcopeereboom, line 12: While I cannot put my finger on it, it seems not unlikely that there is a logic race/error somewhere in this code. I would create a test that has hundreds of go routines attacking the pipeline and see what `test -race` comes up with.

marcopeereboom (changes requested):

service/popm/popm.go
```diff
@@ -110,10 +123,11 @@ func NewMiner(cfg *Config) (*Miner, error) {
 
 	m := &Miner{
 		cfg:            cfg,
-		keystoneCh:     make(chan *hemi.L2Keystone, 3),
 		bfgCmdCh:       make(chan bfgCmd, 10),
 		holdoffTimeout: 5 * time.Second,
 		requestTimeout: 5 * time.Second,
+		mineNowCh:      make(chan struct{}, 1),
+		l2Keystones:    make(map[string]L2KeystoneProcessingContainer),
```
marcopeereboom, line 130: These really should have an initial size

service/popm/popm.go
```diff
@@ -435,24 +449,50 @@ func (m *Miner) BitcoinUTXOs(ctx context.Context, scriptHash string) (*bfgapi.Bi
 	return ir, nil
 }
 
+func (m *Miner) mineKnownKeystones(ctx context.Context) {
+	keystonesFailed := false
+
+	m.ForEachL2Keystone(func(ks hemi.L2Keystone) error {
```
marcopeereboom, line 455: Why is this not a proper for loop?

marcopeereboom (changes requested):
I may not be fully appreciating what is going on but let's spend a minute talking through this PR.

service/popm/popm.go
ClaytonNorthey92: I wanted the `ForEachL2Keystone` to handle the locking and copying, I could pull it out and make it a for loop if you want

service/popm/popm.go
ClaytonNorthey92: yeah I can do that, I can make it the size of the max keystones (10)

service/popm/popm.go
```diff
@@ -785,3 +841,75 @@ func (m *Miner) Run(pctx context.Context) error {
...
+
+	m.l2Keystones[key] = toInsert
+}
+
+// l2KeystonesForProcessing creates copies of the l2 keystones, set them to
+// "processing", then returns the copies with the newest first
+func (m *Miner) l2KeystonesForProcessing() []hemi.L2Keystone {
+	m.mtx.Lock()
+
+	copies := []hemi.L2Keystone{}
```
marcopeereboom, line 893: This should be allocated with make outside of the mutex.

service/popm/popm.go
ClaytonNorthey92: changed it to a `for` loop 👍

service/popm/popm.go
ClaytonNorthey92: done 👍

service/popm/popm.go
```diff
@@ -435,24 +449,66 @@ func (m *Miner) BitcoinUTXOs(ctx context.Context, scriptHash string) (*bfgapi.Bi
...
+	if !keystonesFailed {
+		return
+	}
+
+	// if a keystone failed, ensure that we retry
+	select {
+	case <-ctx.Done():
+		return
+	case <-time.After(l2KeystoneRetryTimeout):
+		go m.mineKnownKeystones(ctx)
```
marcopeereboom, line 492: you should not `go recurse calls. And can we get rid of the channel?

marcopeereboom (changes requested):

service/popm/popm.go
ClaytonNorthey92: sure 👍

service/popm/popm_test.go
```diff
@@ -643,6 +981,136 @@ func TestConnectToBFGAndPerformMine(t *testing.T) {
...
+		case <-ctx.Done():
+			if ctx.Err() != nil {
+				t.Fatal(ctx.Err())
+			}
+		}
+		missing := false
+		for m := range messagesExpected {
+			message := fmt.Sprintf("%s", m)
+			if messagesReceived[message] != messagesExpected[m] {
+				t.Logf("still missing message %v, found %d want %d", m, messagesReceived[message], messagesExpected[m])
```
joshuasing, line 1039: Will this test pass if there are missing messages? :thinking:

service/popm/popm_test.go
```diff
@@ -643,6 +981,136 @@ func TestConnectToBFGAndPerformMine(t *testing.T) {
...
+		case <-ctx.Done():
+			if ctx.Err() != nil {
+				t.Fatal(ctx.Err())
+			}
+		}
+		missing := false
+		for m := range messagesExpected {
+			message := fmt.Sprintf("%s", m)
+			if messagesReceived[message] < messagesExpected[m] {
+				t.Logf("still missing message %v, found %d want %d", m, messagesReceived[message], messagesExpected[m])
```
joshuasing, line 1104: Will this test pass if there are missing messages? :thinking:

service/popm/popm_test.go
ClaytonNorthey92: @joshuasing it won't pass, if there are missing messages it will time out because it won't break the `for` loop

service/popm/popm_test.go
ClaytonNorthey92: @joshuasing it won't pass, if there are missing messages it will time out because it won't break the `for` loop

service/popm/popm.go
```diff
@@ -821,3 +864,76 @@ func (m *Miner) Run(pctx context.Context) error {
...
+
+// l2KeystonesForProcessing creates copies of the l2 keystones, set them to
+// "processing", then returns the copies with the newest first
+func (m *Miner) l2KeystonesForProcessing() []hemi.L2Keystone {
+	copies := make([]hemi.L2Keystone, 0)
+
+	m.mtx.Lock()
+
+	for i, v := range m.l2Keystones {
+
```
joshuasing, line 919: nit: I think this extra newline could be removed to help improve readability.

joshuasing (approved):

marcopeereboom (approved):

PR #22, opened 2024-03-05
Network test start height + no panic

ClaytonNorthey92:
**Summary**
Upon testing with @joshuasing , I realized that we generate 5000 btc blocks in our test, but we tell bfg to start at block 100.  This causes bfg to take too long to catch up.

Also, don't panic in go routines if the context is done, this can cause tests to fail when they shouldn't

**Changes** 
* start bfg btc height high so we don't take too long to process
* don't panic in go routines if context is done


joshuasing (approved):

jcvernaleo (approved):
OK

PR #24, opened 2024-03-05
returning response errors if they exist from bfg -> popm

### LINKED ISSUE #23: Log response errors in popm from BFG
### Confirmation

### Problem

As of now, when popm calls bfg, bfg may respond with a _response_ but we don't check that response for errors. for example:

```go
res, err := m.callBFG(ctx, m.requestTimeout, bbr)
if err != nil {
	return nil, err
}

	
bbResp, ok := res.(*bfgapi.BitcoinBroadcastResponse)
if !ok {
	return nil, fmt.Errorf("not a bitcoin broadcast response %T", res)
}
```

should be:
```go
res, err := m.callBFG(ctx, m.requestTimeout, bbr)
if err != nil {
	return nil, err
}

	
bbResp, ok := res.(*bfgapi.BitcoinBroadcastResponse)
if !ok {
	return nil, fmt.Errorf("not a bitcoin broadcast response %T", res)
}

if bbResp.Error != nil {
   // handle here
}
```

### Suggested solution

make sure we handle bfg response errors as well

### Additional Information

_No response_

ClaytonNorthey92:
**Summary**
As of now, we're not returning response errors from bfg in popm if they occur, this makes it harder to debug.

**Changes**
return response errors as errors in popm 

fixes #23

### DIFF
--- a/api/protocol/protocol.go
+++ b/api/protocol/protocol.go
@@ -258,6 +258,10 @@ func (e Error) String() string {
 	return fmt.Sprintf("%v [%v:%v]", e.Message, e.Trace, e.Timestamp)
 }
 
+func (e Error) Error() string {
+	return e.String()
+}
+
 // RequestError wraps an error to create a protocol request error.
 //
 // Request errors are usually something caused by a client, e.g. validation or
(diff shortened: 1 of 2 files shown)


joshuasing (approved):

jcvernaleo (approved):
OK

PR #26, opened 2024-03-06
electrumx: add connection reuse and pooling

joshuasing:
**Summary**
Add a connection pool (and connection reuse) to the electrumx client.

Previously, the electrumx client was not persistent. Each request would result in a new TCP connection being established,
the request being written, a response being read and the connection being closed. This worked and was reliable, however it was not performant due to the overhead of establishing and closing a new TCP connection for each request.

**Changes**
 - Add `clientConn` which handles a single connection to an ElectrumX server. `clientConn` handles writing requests and reading responses on that connection, as well as pinging the server to keep the connection alive.

 - Add `connPool` which manages a pool of `clientConn`s. A connection can be acquired by calling `connPool.acquireConn()`, which will return a connection from the pool if there is a connection available, otherwise a new connection will be created. Once a connection has been used, it can be returned to the pool with `connPool.freeConn(conn)`, which will append the connection to the pool if the pool is smaller than the maximum size, otherwise the connection will be closed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,4 +92,4 @@ checksums: dist
 
 networktest:
 	HEMI_RUN_NETWORK_TEST=1	\
-		go test -count=1 -v -run TestFullNetwork -v ./e2e/network_test.go
+		go test -count=1 -v -run TestFullNetwork -v -timeout 25m ./e2e/network_test.go
--- a/Makefile
+++ b/Makefile
@@ -92,4 +92,4 @@ checksums: dist
 
 networktest:
 	HEMI_RUN_NETWORK_TEST=1	\
-		go test -count=1 -v -run TestFullNetwork -v -timeout 25m ./e2e/network_test.go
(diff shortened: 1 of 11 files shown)


joshuasing:
This should be functional, however `make networktest` currently fails due to https://github.com/hemilabs/heminetwork/pull/22.

ClaytonNorthey92 (commented):
hey @joshuasing just two questions. however I am starting to feel strongly about moving the connection pool to its own package; there is quite a bit of mutex locking/unlocking going on, and in your code I think we've avoided race conditions, however I fear that keeping this in the electrumx package could lead to unexpected locking/unlocking.  I also don't think there is any electrumx-specific logic in the connection pool 🤔 so I think it is worth separating. what do you think?

ClaytonNorthey92 (approved):

jcvernaleo (approved):
OK

PR #32, opened 2024-03-18
Use 'errors.Is' to compare errors

joshuasing:
**Summary**
Use `errors.Is` to compare errors, instead of using the `==` operator.
More information: https://pkg.go.dev/errors#Is

**Changes**
 - Replace use of `==` for error comparisons with `errors.Is`


jcvernaleo (approved):
OK

marcopeereboom (approved):

PR #33, opened 2024-03-18
Use `maps.Clone(m)` to copy returned map in `APICommands()`

joshuasing:
**Summary**
Copy the commands map with `maps.Clone` before returning in `{bssapi,bfgapi}.APICommands()`
More information: https://pkg.go.dev/maps#Clone

**Changes**
 - Use `maps.Clone(commands)` to clone the commands map before returning from `APICommands()`


jcvernaleo (approved):
OK

marcopeereboom (approved):

PR #37, opened 2024-03-20
localnet

### LINKED ISSUE #21: Create Optimism/Ethereum Side of Local Network
### Confirmation

### Problem

As of now, our local network only spins up the bitcoin+hemi services.  This gives us 100% control over out bitcoin environment.  We want to do the same for the ethereum/optimism side.

### Suggested solution

We need to run an ethereum node in the ethereum version of running bitcoind in regtest mode.  Do that then set up the rest of the optimism network to all run locally.  Outside of downloading dependencies, this should be totally isolated and not require internet connection.

add these services to `./e2e/docker-compose.yml`

### Additional Information

@max-sanchez to provide additional details

ClaytonNorthey92:
**Summary**
Allowing a user to run a local network, full e2e from eth <--> btc

fixes #21 

**Changes**
* updated e2e/docker-compose.yml to include optimism stack
* updated popm to allow retrying mining of keystones based on set threshold
* removing block creation rate configuration; two reasons: it wasn't working before (the docker compose config was removed) + geth L2+L1 is sensative to timing

***

See `README.md` for directions on building, breaking cache, and tearing down


ClaytonNorthey92:
fixes #21

e2e/optimism-stack.Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+
+WORKDIR /git/optimism/packages/contracts-bedrock
+RUN forge install
+RUN forge build
+
+WORKDIR /git/optimism
+
+# this is not great, but this command generates the needed files for devnet
+# and runs a failing test, ignore the failure but use the files
+RUN make devnet-allocs || :
```
ClaytonNorthey92, line 60: this is causing issues; need better solution

README.md
```diff
@@ -124,16 +124,33 @@ Prerequisites: `docker`
...
+For optimism cache break only:
+```
+docker compose -f ./e2e/docker-compose.yml build --build-arg OPTIMISM_CACHE_BREAK="$(date)"
+```
+
+**IMPORTANT:** make sure you run the following to tear down, this will remove
+data and give you a fresh start
+
+```
+docker-compose -f ./e2e/docker-compose.yml down -v --remove-orphans
```
jcvernaleo, line 151: I don't think we should mix `docker compose` and `docker-compose` in the instructions.  Should either use the plugin form or the legacy form consistantly.

jcvernaleo (changes requested):

e2e/optimism-stack.Dockerfile
jcvernaleo: You could wrap it in a script that runs the make command, expects the failure, checks for the generated files, does any cleanup, then reports success.

ClaytonNorthey92: That's a great idea. That would make this PR much smaller too since I wouldn't need to store the generates files in version controll. Will update 👍

ClaytonNorthey92: I'm just now remembering why I can't use this approach; I'll add some more detail in the AM

ClaytonNorthey92: hey @jcvernaleo ok so here is the pickle I am in with this:

with our current `hemi` branch of `optimism`, when running the `make devnet-allocs` command, I run into this issue:
```
/go/pkg/mod/github.com/cockroachdb/pebble@v0.0.0-20230928194634-aa077af62593/internal/cache/robin_hood.go:73:2: type [2147483647]byte larger than address space
```

this issue comes before the needed files are generated.  it's easily fixable, but it requires me to update `optimism` and `op-geth`, which is not ideal at this moment.

so, annoyingly, I think the safest thing to do here is generate the files from the updated `optimism` and `op-geth` branches, include them in version control, and not run the file generation command.  this means that we don't have to do anything for `optimism` and `op-geth` for this PR as well

e2e/docker-compose.yml
```diff
@@ -112,17 +128,154 @@ services:
...
+      - --num-confirmations=1
+      - --safe-abort-nonce-too-low-count=3
+      - --resubmission-timeout=30s
+      - --rpc.addr=0.0.0.0
+      - --rpc.port=8548 
+      - --rpc.enable-admin
+      - --max-channel-duration=1
+      - --max-pending-tx=1
+      - --l1-eth-rpc=http://op-geth-l1:8545
+      - --private-key=dfe61681b31b12b04f239bc0692965c61ffc79244ed9736ffa1a72d00a23a530
```
jcvernaleo, line 252: Can we move private keys and addresses to VARs?

e2e/docker-compose.yml
```diff
@@ -112,17 +128,154 @@ services:
...
+  op-proposer:
+    build:
+      dockerfile: optimism-stack.Dockerfile
+      context: .
+    command: 
+      - op-proposer/bin/op-proposer
+      - --poll-interval=1s
+      - --rpc.port=8560
+      - --rollup-rpc=http://op-node:8547
+      - --l2oo-address=0xD3292428DE99a781A4A5e503622E8c80E5383e20
```
jcvernaleo, line 269: same as above

jcvernaleo (changes requested):
Another minor thing.

e2e/docker-compose.yml
ClaytonNorthey92: ah yes good call 👍

cmd/popmd/popmd.go
```diff
@@ -60,6 +60,12 @@ var (
 			Help:         "address and port bssd prometheus listens on",
 			Print:        config.PrintAll,
 		},
+		"POPM_REMINE_THRESHOLD": config.Config{
```
marcopeereboom, line 63: This has got to go plese.

jcvernaleo (approved):
OK

PR #41, opened 2024-03-25
added tbcapi-btc-block-metadata-by-num

### LINKED ISSUE #38: tbcapi: implement tbcapi-btc-block-metadata-by-num tbcd call
### Confirmation

### Problem

need to implement `tbcapi-btc-block-metadata-by-num` call , this is defined as follows:

Returns the Metadata for a specific block in the canonical BTC chain processed by TBC.

Request Query:

```go
type BtcBlockMetadataByNumRequest struct {
	Height		uint32
}
```

Response Query:

```go
type BtcBlockMetadataByNumResponse struct {
	Block		BtcBlockMetadata
}
```

### Suggested solution

implement as defined in requirements

### Additional Information

_No response_

ClaytonNorthey92:
added tbcapi-btc-block-metadata-by-num endpoint to tbc, added tests but only run if you have docker because they depend on testcontainers to run a bitcoind instance

**Summary**
added tbcapi-btc-block-metadata-by-num endpoint to tbc

**Changes**
* added tbcapi-btc-block-metadata-by-num endpoint to tbc
* added HEMI_DOCKER_TESTS to only run tests that use docker when this is set
* using testcontainers + bitcoind to run tbc tests

fixes #38


api/tbcapi/tbcapi.go
```diff
@@ -27,11 +30,40 @@ var (
...
 	PingRequest  protocol.PingRequest
 	PingResponse protocol.PingResponse
+	BtcHeader    struct {
+		Version uint32 `json:"version"`
+
+		// hex encoded byte array
+		PrevHash string `json:"prev_hash"`
+
+		// hex encoded byte array
+		MerkleRoot string `json:"merkle_root"`
```
ClaytonNorthey92, line 40: @max-sanchez @joshuasing I used hex string encoding for the following fields:
* `merkle_root`
* `prev_hash`
* `bits`

the reason I do so is to match `bitcoin-cli`

ClaytonNorthey92: I know that strays from the docs a bit so I wanted to call it out here

ClaytonNorthey92: trying to match this response https://developer.bitcoin.org/reference/rpc/getblockheader.html

api/tbcapi/tbcapi.go
```diff
@@ -27,11 +30,40 @@ var (
...
+		Height uint32    `json:"height"`
+		NumTx  uint32    `json:"num_tx"`
+		Header BtcHeader `json:"header"`
+	}
+	BtcBlockMetadataByNumRequest struct {
+		Height uint32 `json:"height"`
+	}
+	BtcBlockMetadataByNumResponse struct {
+		Block BtcBlockMetadata `json:"block"`
+	}
```
joshuasing, line 59: I think we should each struct into a separate type block to help improve consistently with the rest of the code base.

ClaytonNorthey92: great idea, will update 👍

service/tbc/websocket.go
```diff
@@ -47,6 +50,63 @@ func (ws *tbcWs) handlePingRequest(ctx context.Context, payload any, id string)
 	return nil
 }
 
+func (ws *tbcWs) handleBtcBlockMetadataByNumRequest(ctx context.Context, payload any, id string, db tbcd.Database) error {
```
ClaytonNorthey92, line 53: handle better for error cases here; add an `Error` field for user and internal errors, handle accordingly, and write tests

service/tbc/websocket.go
```diff
@@ -47,6 +50,69 @@ func (ws *tbcWs) handlePingRequest(ctx context.Context, payload any, id string)
...
+
+		return nil
+	}
+
+	p, ok := payload.(*tbcapi.BtcBlockMetadataByNumRequest)
+	if !ok {
+		return fmt.Errorf("handleBtcBlockMetadataByNumRequest invalid payload type: %T", payload)
+	}
+
+	bh, err := db.BlockHeadersByHeight(ctx, uint64(p.Height))
```
ClaytonNorthey92, line 71: @max-sanchez @marcopeereboom what do we do if we have more than 1 block header at this height? 🤔

ClaytonNorthey92: resolved with this for now https://bloqexternal.slack.com/archives/C06JTUUS9RA/p1711470734212029

api/tbcapi/tbcapi.go
```diff
@@ -29,9 +32,43 @@ type (
...
+type BtcHeader struct {
+	Version uint32 `json:"version"`
+
+	// hex encoded byte array
+	PrevHash string `json:"prev_hash"`
+
+	// hex encoded byte array
+	MerkleRoot string `json:"merkle_root"`
+
+	Timestamp uint32 `json:"timestamp"`
```
joshuasing, line 44: I think we could use an `uint64` for the timestamp. `uint32` is okay, but I am generally against using 32-bit integers for storing time.

joshuasing (approved):

api/tbcapi/tbcapi.go
ClaytonNorthey92: I can update this, it's pretty quick 👍

jcvernaleo (approved):
OK

joshuasing (approved):

PR #51, opened 2024-04-05
Add common user-specific files to gitignore

joshuasing:
**Summary**
Add common user-specific files to the `.gitignore` file.

**Changes**
 - Add common Vim temporary files to `.gitignore`
 - Add common JetBrains temporary/user-specific files to `.gitignore`
 - Add common Visual Studio Codes temporary/user-specific files to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,32 @@
+### Hemi
 /bin/
 /dist/
 /pkg/
-.gocache
-*.sw?
+/.gocache/
+
+### Common editors
+## Vim
+# Swap
+[._]*.s[a-v][a-z]
+!*.svg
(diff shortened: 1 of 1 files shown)


jcvernaleo (changes requested):
If we're going to add those editor files, can we add eamcs *~ files too?

joshuasing:
> If we're going to add those editor files, can we add eamcs *~ files too?

Sure thing, I will do that now.

jcvernaleo (approved):
I've been using emacs for over a quarter of a century and those are more random files than even I know about :)

ClaytonNorthey92 (approved):

PR #63, opened 2024-04-08
tbcapi: use api.ByteSlice to encode bytes as hex

joshuasing:
**Summary**
Use `api.ByteSlice` instead of raw byte slices in `tbcapi` to encode bytes as hexadecimal strings.

**Changes**
 - Use `api.ByteSlice` instead of `[]byte` in `tbcapi` package

### DIFF
--- a/api/tbcapi/tbcapi.go
+++ b/api/tbcapi/tbcapi.go
@@ -6,8 +6,6 @@ package tbcapi
 
 import (
 	"context"
-	"encoding/hex"
-	"encoding/json"
 	"fmt"
 	"maps"
 	"reflect"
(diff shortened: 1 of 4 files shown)


api/tbcapi/tbcapi.go
```diff
@@ -104,16 +107,34 @@ type UtxosByAddressRequest struct {
...
+	if err != nil {
+		return err
+	}
+	*tx = [32]byte(s)
+	return nil
+}
+
 type TxByIdRequest struct {
-	TxId [32]byte `json:"tx_id"`
+	TxId TxId `json:"tx_id"`
```
ClaytonNorthey92, line 133: @joshuasing how do you feel about just making this an `api.ByteSlice`?

ClaytonNorthey92 (approved):

PR #65, opened 2024-04-08
specifiying errors for tbc, tx-by-id, utxos-by-address

ClaytonNorthey92:
**Summary**
specifiying errors for tbc, tx-by-id, utxos-by-address

**Changes**
* added an error code to error responses
* better error handling in tbc tx-by-id, utxos-by-address
* updated tests to test against codes instead of text

### DIFF
--- a/api/protocol/protocol.go
+++ b/api/protocol/protocol.go
@@ -28,6 +28,11 @@ const (
 	logLevel           = "protocol=INFO"
 	WSConnectTimeout   = 20 * time.Second
 	WSHandshakeTimeout = 15 * time.Second
+
+	ErrCodeUnknown    = 0
+	ErrCodeNotFound   = 1
+	ErrCodeBadRequest = 2
+	ErrInternalError  = 3
 )
 
 const (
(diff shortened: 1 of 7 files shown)


ClaytonNorthey92:
discussed with @joshuasing on slack about adding error codes to protocol errors to determine what kind of error they are, rather than relying on text

api/protocol/protocol.go
```diff
@@ -28,6 +28,11 @@ const (
 	logLevel           = "protocol=INFO"
 	WSConnectTimeout   = 20 * time.Second
 	WSHandshakeTimeout = 15 * time.Second
+
+	ErrCodeUnknown    = 0
+	ErrCodeNotFound   = 1
+	ErrCodeBadRequest = 2
+	ErrCodeInternal   = 3
```
joshuasing, line 35: I think moving these to a separate `const` block would improve readability.
Also, I think `iota` could be used here.

joshuasing (approved):

marcopeereboom (approved):

PR #67, opened 2024-04-09
tbc: remove use of math/rand from rpc.go

joshuasing:
**Summary**
Remove use of `math/rand` package in `service/tbc/rpc.go`.

I believe the original code in `service/tbc/rpc.go` used `crypto/rand`, however it appears to have been changed in #46 and committed as bd5f7bc5bdd3fb475ae291f0cc10c34984dd2d76.

The `math/rand` package is deterministic and should never be used where security matters.

Go 1.22 introduced [`math/rand/v2`](https://pkg.go.dev/math/rand/v2), which aims to improve the current `math/rand` package. I think it would be nice if we could update the required version in `go.mod` to Go 1.22 so that we can use this package instead of `math/rand`.

**Changes**
- Use `crypto/rand` instead of `math/rand` package in `service/tbc/rpc.go`.

### DIFF
--- a/service/tbc/rpc.go
+++ b/service/tbc/rpc.go
@@ -6,10 +6,10 @@ package tbc
 
 import (
 	"context"
+	"crypto/rand"
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"math/rand"
 	"net/http"
 	"sync"
 	"time"


ClaytonNorthey92 (approved):

marcopeereboom:
Line 84
```
        rand.Seed(time.Now().UnixNano()) // used for seeding, ok to be math.rand
```

I don't mind if we use crypto rand but it isn't a bug. The comment does need to be removed if you commit this.

marcopeereboom (approved):
Do remove comment that talks about math rand

PR #74, opened 2024-04-11
utxos-by-address raw vs cooked + tx-by-id cleanup

### LINKED ISSUE #60: TBC: utxos-by-addresssr raw vs cooked
### Confirmation

### Problem

We want to split this endpoint into two: one that returns encoded raw bytes as the response, and one that returns a "cooked" response, i.e. the fields parsed out into a json structure.

### Suggested solution

We want to split this endpoint into two: one that returns encoded raw bytes as the response, and one that returns a "cooked" response, i.e. the fields parsed out into a json structure.

### Additional Information

_No response_

ClaytonNorthey92:
**Summary**

utxos-by-address raw vs cooked + tx-by-id cleanup

**Changes**

this commit addresses two things:
* splitting up utxos-by-address into raw-vs-cooked
* fixing tx-by-id to return the error that will also be in the response; this allows us to print the error when responding

fixes #60

### DIFF
--- a/api/tbcapi/tbcapi.go
+++ b/api/tbcapi/tbcapi.go
@@ -112,7 +112,7 @@ type UtxosByAddressRawRequest struct {
 
 type UtxosByAddressRawResponse struct {
 	Utxos []api.ByteSlice `json:"utxos"`
-	Error *protocol.Error `json:"error"`
+	Error *protocol.Error `json:"error,omitempty"`
 }
 
 type UtxosByAddressRequest struct {
@@ -129,7 +129,7 @@ type Utxo struct {
 
 type UtxosByAddressResponse struct {
 	Utxos []Utxo          `json:"utxos"`
-	Error *protocol.Error `json:"error"`
+	Error *protocol.Error `json:"error,omitempty"`
 }
 
 type TxByIdRawRequest struct {
@@ -138,7 +138,7 @@ type TxByIdRawRequest struct {
 
 type TxByIdRawResponse struct {
 	Tx    api.ByteSlice   `json:"tx"`
-	Error *protocol.Error `json:"error"`
+	Error *protocol.Error `json:"error,omitempty"`
 }
 
 type TxByIdRequest struct {
@@ -147,7 +147,7 @@ type TxByIdRequest struct {
 
 type TxByIdResponse struct {
 	Tx    Tx              `json:"tx"`
-	Error *protocol.Error `json:"error"`
+	Error *protocol.Error `json:"error,omitempty"`
 }
 
 type OutPoint struct {
--- a/api/tbcapi/tbcapi.go
+++ b/api/tbcapi/tbcapi.go
@@ -29,6 +29,9 @@ const (
 	CmdBalanceByAddressRequest  = "tbcapi-balance-by-address-request"
 	CmdBalanceByAddressResponse = "tbcapi-balance-by-address-response"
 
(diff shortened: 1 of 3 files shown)


api/tbcapi/tbcapi.go
```diff
@@ -101,14 +104,31 @@ type BalanceByAddressResponse struct {
...
 
+type UtxosByAddressRawRequest struct {
+	Address string `json:"address"`
+	Start   uint   `json:"start"`
+	Count   uint   `json:"count"`
+}
+
+type UtxosByAddressRawResponse struct {
+	Utxos []api.ByteSlice `json:"utxos"`
+	Error *protocol.Error `json:"error"`
```
marcopeereboom, line 115: Error shoud be optional. This goes for all other Error' in JSON as well

marcopeereboom (changes requested):

api/tbcapi/tbcapi.go
ClaytonNorthey92: ah via `omit_empty`?

joshuasing: Yes, please add `omitempty` to this error to be consistent with all other api packages.

service/tbc/rpc.go
```diff
@@ -259,18 +294,21 @@ func (s *Server) handleUtxosByAddressRequest(ctx context.Context, req *tbcapi.Ut
 			e := protocol.NewInternalError(err)
 			return &tbcapi.UtxosByAddressResponse{
 				Error: e.ProtocolError(),
-			}, e
+			}, err
```
joshuasing, line 297: The internal error (`e`) should be returned as the error here, as it contains the unique trace for the error.

service/tbc/rpc.go
```diff
@@ -259,18 +294,21 @@ func (s *Server) handleUtxosByAddressRequest(ctx context.Context, req *tbcapi.Ut
...
 		}
 
-		// TODO(joshuasing): do not include raw err in errors sent to client
 		return &tbcapi.UtxosByAddressResponse{
-			Error: protocol.RequestErrorf("error getting utxos for address: %s", err),
-		}, nil
+			Error: protocol.RequestErrorf("error getting utxos for address: %s", req.Address),
+		}, err
```
joshuasing, line 302: I think we could return `nil` as the error here, as this error is likely caused by the client.
What do you think?

service/tbc/rpc.go
```diff
@@ -282,13 +320,28 @@ func (s *Server) handleTxByIdRawRequest(ctx context.Context, req *tbcapi.TxByIdR
...
+		responseErr := protocol.RequestErrorf("invalid tx id")
 		return &tbcapi.TxByIdRawResponse{
 			Error: responseErr,
 		}, responseErr
 	}
 
+	tx, err := s.TxById(ctx, [32]byte(req.TxId))
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			responseErr := protocol.RequestErrorf("not found: %s", hex.EncodeToString(req.TxId))
```
joshuasing, line 333: I think this error could be "tx not found: %s" to help improve consistency.

service/tbc/rpc.go
ClaytonNorthey92: ah good catch, thanks

service/tbc/rpc.go
ClaytonNorthey92: ah yes, I'll update

marcopeereboom (approved):

joshuasing (approved):

PR #75, opened 2024-04-11
tbc: set peersWanted = 1 when running in localnet mode

joshuasing:
**Summary**
Set `peersWanted` to `1` when running in localnet mode, as there will only be one peer.
This helps to prevent thousands of extra go routines being created in the peerManager when running in localnet mode.

**Changes**
 - Add `peersWanted int` to the `tbc.Server` struct, defaulting to the `defaultPeersWanted` const.
 - Set `peersWanted` to `1` when running in localnet mode.


joshuasing:
This does not sufficiently resolve the issue causing this.

PR #82, opened 2024-04-16
Add aggressive ping/pong to cull questionable connections.

marcopeereboom:
The ttl package can be used to conveniently timeout events. Essentially it is a map that has an expiration per key.

**Summary**
`tbc` needs to schedule events in the future such as ping responses. This pr adds a convenient package to timeout or cancel events. The pings are needed to more aggressively kill misbehaving connections.

**Changes**

### DIFF
--- a/ttl/ttl.go
+++ b/ttl/ttl.go
@@ -38,40 +38,34 @@ func New(capacity int) (*TTL, error) {
 
 func (tm *TTL) ttl(ctx context.Context, key any) {
 	<-ctx.Done()
-	switch err := ctx.Err(); {
-	case err == nil:
-		// not yet closed
+	err := ctx.Err()
+	if err == nil {
+		return
+	}
 
+	tm.mtx.Lock()
+	defer tm.mtx.Unlock()
+	v, ok := tm.m[key]
+	if !ok {
+		return
+	}
+
+	switch {
 	case errors.Is(err, context.DeadlineExceeded):
 		// expired
-		tm.mtx.Lock()
-		defer tm.mtx.Unlock()
-		v, ok := tm.m[key]
-		if !ok {
-			return
-		}
 		if v.expired != nil {
 			go v.expired(key, v.value)
 		}
 
-		// For now assume we always want to remove key
-		delete(tm.m, key)
-
 	case errors.Is(err, context.Canceled):
 		// This is the caller calling cancel
-		tm.mtx.Lock()
-		defer tm.mtx.Unlock()
-		v, ok := tm.m[key]
-		if !ok {
-			return
-		}
 		if v.remove != nil {
 			go v.remove(key, v.value)
 		}
-
-		// For now assume we always want to remove key
-		delete(tm.m, key)
 	}
+
+	// For now assume we always want to remove key
+	delete(tm.m, key)
 }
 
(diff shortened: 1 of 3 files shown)


ClaytonNorthey92 (approved):

ttl/ttl.go
```diff
@@ -0,0 +1,115 @@
...
+
+func New(capacity int) (*TTL, error) {
+	return &TTL{
+		m: make(map[any]value, capacity),
+	}, nil
+}
+
+func (tm *TTL) ttl(ctx context.Context, key any) {
+	select {
+	case <-ctx.Done():
```
joshuasing, line 41: I don't think this `select` block is necessary as there is only one case, and could be simplified:
```suggestion
<-ctx.Done()
```

ttl/ttl.go
```diff
@@ -0,0 +1,115 @@
...
+				go v.remove(key, v.value)
+			}
+
+			// For now assume we aways want to remove key
+			delete(tm.m, key)
+		}
+	}
+}
+
+func (tm *TTL) Put(pctx context.Context, ttl time.Duration, key any, val any, expired func(any, any), remove func(any, any)) {
```
joshuasing, line 79: Could this function ever be passed a `ttl` of zero? If so, should we handle this when creating the context in the stored value? 🤔

ttl/ttl_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	var wg sync.WaitGroup
+	for i := 0; i < count; i++ {
+		wg.Add(1)
+		tm.Put(ctx, time.Second, strconv.Itoa(i), &wg, callback, nil)
+	}
+	for i := 0; i < count; i++ {
+		key := strconv.Itoa(i)
+		_, err := tm.Get(key)
+		if err != nil {
```
joshuasing, line 48: nit: I think this error check could be inlined:
```suggestion
		if _, err := tm.Get(key); err != nil {
```

ttl/ttl_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	var wg sync.WaitGroup
+	for i := 0; i < count; i++ {
+		wg.Add(1)
+		tm.Put(ctx, time.Second, strconv.Itoa(i), &wg, callbackPanic, callback)
+	}
+	for i := 0; i < count; i++ {
+		key := strconv.Itoa(i)
+		_, err := tm.Get(key)
+		if err != nil {
```
joshuasing, line 81: nit: I think this error check could be inlined:
```suggestion
		if _, err := tm.Get(key); err != nil {
```

ttl/ttl_test.go
```diff
@@ -0,0 +1,98 @@
...
+			t.Fatalf("%v: %v", key, err)
+		}
+		err = tm.Cancel(key)
+		if err != nil {
```
joshuasing, line 85: nit: I think this error check could be inlined:
```suggestion
		if err = tm.Cancel(key); err != nil {
```

ttl/ttl.go
marcopeereboom: Yes, don't do that.

joshuasing (approved):

PR #87, opened 2024-04-18
Replace os.Kill with syscall.SIGTERM in signal.Notify calls

joshuasing:
**Summary**
Replace `os.Kill` with `syscall.SIGTERM` in `signal.Notify` calls.

[`os.Kill` is equal to `syscall.SIGKILL`](https://pkg.go.dev/os#pkg-constants:~:text=Kill%20%20%20%20%20%20Signal%20%3D%20syscall.SIGKILL)

SIGKILL cannot be caught on Unix-like systems.
SIGTERM is often used to request that an application terminate gracefully.

https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html
https://man.openbsd.org/signal

**Changes**
 - Replace `os.Kill` (syscall.SIGKILL) with `syscall.SIGTERM` in signal handlers

### DIFF
--- a/cmd/bfgd/bfgd.go
+++ b/cmd/bfgd/bfgd.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"syscall"
 
 	"github.com/juju/loggo"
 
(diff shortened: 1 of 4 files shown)


jcvernaleo (approved):
OK

ClaytonNorthey92 (approved):

PR #90, opened 2024-04-18
Track pending blocks with ttl package

### LINKED ISSUE #84: Replace Server.blocks with `ttl` package.
### Confirmation

### Problem

The blocks map has an implicit timeout that get's episodically scanned for timeouts. Instead of scanning we should rely on the `ttl` package that provides convenient time-to-live handling.

### Suggested solution

From the looks of the code it is possible to move all the block insert logic into the `Cancel` function and the timeout of a block download in the `remove` callback. This should drastically simplify the block download cache.

### Additional Information

_No response_

marcopeereboom:
**Summary**
The pending block cache is scanned in various locations. Instead of scanning this makes the block expire using the ttl package and thus the code is more even driven and it locks less.

**Changes**
- Cope with the new ttl package for pings.
- Replace blockPeer map with ttl.

Fixes #84

### DIFF
--- a/service/tbc/tbc.go
+++ b/service/tbc/tbc.go
@@ -156,7 +156,6 @@ type blockPeer struct {
 type Config struct {
 	AutoIndex               bool
 	BlockSanity             bool
-	RegtestPort             string
 	LevelDBHome             string
 	ListenAddress           string
 	LogLevel                string
@@ -192,8 +191,8 @@ type Server struct {
 	port        string
 	seeds       []string
 
-	peers  map[string]*peer      // active but not necessarily connected
-	blocks map[string]*blockPeer // outstanding block downloads [hash]when/where
+	peers  map[string]*peer // active but not necessarily connected
+	blocks *ttl.TTL         // outstanding block downloads [hash]when/where
 	pings  *ttl.TTL
 
 	// IBD hints
@@ -226,11 +225,15 @@ func NewServer(cfg *Config) (*Server, error) {
 	if err != nil {
 		return nil, err
 	}
+	blocks, err := ttl.New(defaultPendingBlocks)
+	if err != nil {
+		return nil, err
+	}
 	defaultRequestTimeout := 10 * time.Second // XXX: make config option?
 	s := &Server{
 		cfg:            cfg,
 		printTime:      time.Now().Add(10 * time.Second),
-		blocks:         make(map[string]*blockPeer, defaultPendingBlocks),
+		blocks:         blocks,
 		peers:          make(map[string]*peer, defaultPeersWanted),
 		pings:          pings,
 		blocksInserted: make(map[string]struct{}, 8192), // stats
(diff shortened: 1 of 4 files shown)


ttl/ttl.go
```diff
@@ -13,29 +13,40 @@ import (
...
+	remove  func(any, any) // called when removed from map
+
+	timeoutExpired bool // set when this value has expired
 
 	// Value context
 	ctx    context.Context
 	cancel context.CancelFunc
 }
 
+// TTL is an opaque structure that wraps the TTL key/value map.
```
joshuasing, line 31: I think this comment could be made more descriptive. I don't think "wraps the TTL key/value map" really makes sense, because TTL is basically the key/value store from a user's perspective.

ttl/ttl.go
```diff
@@ -13,29 +13,40 @@ import (
...
-	m map[any]value
+	autoDelete bool
+	m          map[any]*value
 }
 
-func New(capacity int) (*TTL, error) {
+// New creates a new TTL map with the provided capacity. If autoDelete is set
+// then the key will be deleted from the map on either the cancel or timeout
+// event.
+func New(capacity int, autoDelete bool) (*TTL, error) {
```
joshuasing, line 42: Will more parameters like be added to this function in the future? If so, would it be worth using either an option pattern or creating a `ttl.Config` struct?

service/tbc/tbc.go
```diff
@@ -1015,96 +993,80 @@ func (s *Server) downloadBlock(ctx context.Context, p *peer, ch *chainhash.Hash)
...
 		log.Errorf("blocks missing: %v", err)
 		return
 	}
+
 	for k := range bm {
 		bi := bm[k]
 		hash, _ := chainhash.NewHash(bi.Hash[:])
 		hashS := hash.String()
-		if _, ok := s.blocks[hashS]; ok {
+		if _, _, err := s.blocks.Get(hashS); err == nil {
```
joshuasing, line 1055: Would it be worth adding a `Contains(key)` function to the TTL map? :thinking:

ttl/ttl.go
marcopeereboom: IMO two is already too magical so I don't think/hope so.

service/tbc/tbc.go
marcopeereboom: I wasn't sold on having it just yet.

ttl/ttl.go
```diff
@@ -13,29 +13,42 @@ import (
...
 
 	// Value context
 	ctx    context.Context
 	cancel context.CancelFunc
 }
 
+// TTL is an opaque structure that stores key/values in an internal map. These
+// values have a time-to-live callback functioned associated with them.
+// Depending on configuration eiher these values are automatically deleted from
+// the map on expiration.
```
joshuasing, line 33:
```suggestion
// TTL is an opaque structure that stores key/values in an internal map. These
// values have a time-to-live callback functions associated with them.
// Depending on configuration either these values are automatically deleted from
// the map on expiration.
```

joshuasing (approved):

PR #91, opened 2024-04-19
Added configurable fee-per-vB to PoP Miner

max-sanchez:
This adds the environment variable POPM_STATIC_FEE which can be used to configure a specific fee level for the PoP Miner to use.

Longer-term, BFG needs an RPC method for getting current network fee levels.


ClaytonNorthey92 (approved):

joshuasing (approved):

marcopeereboom:
Guys, this really needs to not be a knob and be calculated.

PR #100, opened 2024-04-23
Added README for generating forks in BTC regtest for TBC fork resolution testing

max-sanchez:
Readme for how to create a forked chain in regtest and save the data to replay the fork scenario in the future.


joshuasing (changes requested):
I think this file name could be lowercased, e.g. `regtest-fork-generation.md`. It is common to use uppercase for readme files, but I don't think it is needed for other files.

joshuasing (approved):

ClaytonNorthey92 (approved):

PR #104, opened 2024-04-26
tbcapi: use reverse byte order for hashes in serialised, tidy up

joshuasing:
**Summary**
Use reverse byte order for hashes (tx ids, block ids and merkle root) serialised tbcapi request/responses.

Additionally, change `tbcapi` to consistently use pointers to structs for data in responses, preventing unwanted allocations and causes `null` to be returned instead of an empty object.

**Changes**
- Use reverse byte order for hashes in `tbcapi`
- Tidy up `wire.MsgTx` to `tbcapi.Tx` conversion function
- Consistently use pointers for data structs in `tbcapi` responses

### DIFF
--- a/api/tbcapi/tbcapi.go
+++ b/api/tbcapi/tbcapi.go
@@ -62,12 +62,12 @@ type (
 )
 
 type BlockHeader struct {
-	Version    int32  `json:"version"`
-	PrevHash   string `json:"prev_hash"`
-	MerkleRoot string `json:"merkle_root"`
-	Timestamp  int64  `json:"timestamp"`
-	Bits       string `json:"bits"`
-	Nonce      uint32 `json:"nonce"`
+	Version    int32         `json:"version"`
+	PrevHash   api.ByteSlice `json:"prev_hash"`   // reverse order
+	MerkleRoot api.ByteSlice `json:"merkle_root"` // reverse order
+	Timestamp  int64         `json:"timestamp"`
+	Bits       string        `json:"bits"`
+	Nonce      uint32        `json:"nonce"`
 }
 
 type BlockHeadersByHeightRawRequest struct {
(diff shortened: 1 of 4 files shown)


api/tbcapi/tbcapi.go
```diff
@@ -131,18 +131,18 @@ type UtxosByAddressRequest struct {
...
 type Utxo struct {
-	TxId     api.ByteSlice `json:"tx_id"`
+	TxId     api.ByteSlice `json:"tx_id"` // reverse order
 	Value    uint64        `json:"value"`
 	OutIndex uint32        `json:"out_index"`
 }
 
 type UtxosByAddressResponse struct {
-	Utxos []Utxo          `json:"utxos"`
+	Utxos []*Utxo         `json:"utxos"`
```
marcopeereboom, line 140: I don't think we want pointers here.

api/tbcapi/tbcapi.go
```diff
@@ -151,16 +151,16 @@ type TxByIdRawResponse struct {
 }
 
 type TxByIdRequest struct {
-	TxId api.ByteSlice `json:"tx_id"`
+	TxId api.ByteSlice `json:"tx_id"` // reverse order
 }
 
 type TxByIdResponse struct {
-	Tx    Tx              `json:"tx"`
+	Tx    *Tx             `json:"tx"`
```
marcopeereboom, line 158: Same here.

marcopeereboom (approved):

api/tbcapi/tbcapi.go
joshuasing: If we don't use pointers here, it results in an empty Tx being returned, which is confusing for clients.
With a pointer, the JSON will be `{"tx": null}` if the tx is empty, whereas without a pointer it becomes:
```json
{"tx": {"version": 0, "lock_time": 0, "tx_in": null, "tx_out": null}}
```

PR #105, opened 2024-04-30
Add pprof http server to daemons

### LINKED ISSUE #94: Add pprof to all daemons
### Confirmation

### Problem

_No response_

### Suggested solution

Add pprof to all daemons. This did not happen in the past because of crazy schedules but the time has come.

Simply listen on locahost if an environment variable is set with a listen address and port.

### Additional Information

_No response_

joshuasing:
**Summary**
Add a configuration option that enables a HTTP server that serves `net/http/pprof` handlers in each daemon.
This configuration option is disabled by default and can be enabled by setting an environment variable, e.g. TBC: `TBC_PPROF_ADDRESS=localhost:6060`
https://pkg.go.dev/net/http/pprof

Resolves #94 

**Changes**
- Add `hemi/pprof` package that contains a reusable pprof http server, similar to Deucalion.
- Run a `pprof.Server` in each daemon

*Note: I am unsure whether this is worth moving to a new package that can then be used by each daemon, if you have thoughts on this, please let me know.*


marcopeereboom (approved):
Not sure I love having several muxes but this can go in as is. At some point it would be neat to reuse the same mux.

service/bfg/bfg.go
```diff
@@ -1427,6 +1429,37 @@ func (s *Server) Run(pctx context.Context) error {
 		publicHttpErrCh <- publicHttpServer.ListenAndServe()
 	}()
 
+	if s.cfg.PprofListenAddress != "" {
+		pprofMux := http.NewServeMux()
```
ClaytonNorthey92, line 1433: hey @joshuasing , I am sure you thought about this, but if I am not mistaken, this code is repeated across the services, could this code be importable and callable instead? as to not repeat the logic?

joshuasing: Yes, it could be and I would like to (there's a note in the pr message)

I was unsure of where to put a package for this, and what to name it.

ClaytonNorthey92: ah I missed the note 🤦 I think we should put this in a package.  what to name it, I am unsure.  perhaps `profile`?

ClaytonNorthey92 (approved):

marcopeereboom (changes requested):
I think this is overkill but fine. Please move pprof into service dir.

marcopeereboom (approved):

PR #108, opened 2024-05-07
bfgd: fix loops unconditionally exited after one interation (SA4004)

joshuasing:
**Summary**
Add a target to the Makefile to run [Staticcheck](https://staticcheck.io/), a "state of the art linter for the Go programming language" that uses static analysis.

Additionally, fix several issues detected by Staticcheck in the bfgd postgres database.

**Changes**
- Add `staticcheck` and `staticcheck-deps` make tasks.
- Fix several loops that were unconditionally exited after a single interation in `database/bfgd/postgres/postgres.go` (SA4004).
- Fix bug in `AccessPublicKeyDelete` implementation that caused `database.NotFoundError` to be returned even when a key was found and deleted.

**Notes**
It would be nice to add Staticcheck to the `lint` target in the future, however it currently reports a lot of issues, thus doing this would cause builds to fail.

*Passes tests and E2E networktest.*

### DIFF
--- a/database/bfgd/postgres/postgres.go
+++ b/database/bfgd/postgres/postgres.go
@@ -83,25 +83,25 @@ func New(ctx context.Context, uri string) (*pgdb, error) {
 	return p, nil
 }
 
-func (pg *pgdb) Version(ctx context.Context) (int, error) {
+func (p *pgdb) Version(ctx context.Context) (int, error) {
 	log.Tracef("Version")
 	defer log.Tracef("Version exit")
 
 	const selectVersion = `SELECT * FROM version LIMIT 1;`
 	var dbVersion int
-	if err := pg.db.QueryRowContext(ctx, selectVersion).Scan(&dbVersion); err != nil {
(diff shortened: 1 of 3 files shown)


database/bfgd/postgres/postgres.go
```diff
@@ -946,64 +927,40 @@ func (p *pgdb) AccessPublicKeyExists(ctx context.Context, publicKey *bfgd.Access
...
 	}
 
-	for rows.Next() {
-		var count int
-		if err := rows.Scan(&count); err != nil {
-			return err
-		}
-
-		return database.NotFoundError("public key not found")
-	}
```
joshuasing, line 1002: This behaviour is rather confusing, but I believe it is supposed to return a `database.NotFoundError` if nothing was deleted? The refactored code should have the proper behaviour, but this appears to always return a `database.NotFoundError`?

ClaytonNorthey92: yes I think you're right @joshuasing ; we were missing a check against the count 🤦

ClaytonNorthey92 (approved):

PR #116, opened 2024-05-13
popm: simplify receivedKeystones variable in tests

joshuasing:
**Summary**
Simplify `receivedKeystones` variable in PoP miner tests

**Changes**
- Directly assign `receivedKeystones` variable in `popm_test.go` instead of creating the variable and then adding each element in a `for range` loop.


jcvernaleo (approved):
OK

ClaytonNorthey92 (approved):

PR #119, opened 2024-05-13
Add configuration option for static fees to web PoP miner

max-sanchez:
Added a configuration field for setting a static fee, so web PoP miner has configuration parity with current downloadable PoP miner daemon.

### DIFF
--- a/web/popminer/popminer.go
+++ b/web/popminer/popminer.go
@@ -123,7 +123,7 @@ func runPopMiner(this js.Value, args []js.Value) (any, error) {
 	cfg.BTCChainName = args[0].Get("network").String()
 	cfg.BTCPrivateKey = args[0].Get("privateKey").String()
 	cfg.LogLevel = args[0].Get("logLevel").String() // "popm=TRACE:protocol=TRACE"
-	cfg.StaticFee = (uint)(args[0].Get("staticFee").Int())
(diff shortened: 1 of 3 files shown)


ClaytonNorthey92 (approved):

web/popminer/popminer.go
```diff
@@ -123,6 +123,7 @@ func runPopMiner(this js.Value, args []js.Value) (any, error) {
 	cfg.BTCChainName = args[0].Get("network").String()
 	cfg.BTCPrivateKey = args[0].Get("privateKey").String()
 	cfg.LogLevel = args[0].Get("logLevel").String() // "popm=TRACE:protocol=TRACE"
+	cfg.StaticFee = (uint)(args[0].Get("staticFee").Int())
```
joshuasing, line 126: I think we could simplify this to:
```suggestion
	cfg.StaticFee = uint(args[0].Get("staticFee").Int())
```

joshuasing (changes requested):

web/popminer/popminer.go
max-sanchez: Fixed

joshuasing (approved):

web/popminer/popminer.go
joshuasing: Thanks

PR #125, opened 2024-05-17
ci: fix bug that makes version type always 'unstable'

joshuasing:
**Summary**
Adds a missing `$` character that causes the version type to always be set to 'unstable' in the release workflow.

**Changes**
- Fix version type check in release workflow prepare step


jcvernaleo (approved):
OK

ClaytonNorthey92 (approved):

PR #129, opened 2024-05-20
remove bug (and ineffiency) from MATERIALIZED VIEW

ClaytonNorthey92:
**Summary**
previously, in our canonical chain MATERIALIZED VIEW, we would fall back to height if we couldn't find a previous block by hash

**Changes**
the aforementioned is incorrect (and inefficient), so remove it


ClaytonNorthey92:
hold on...seeing one more bug 🐛

ClaytonNorthey92:
there might be a larger issue going on here, converting to draft

PR #131, opened 2024-05-21
Stopgap to re-fetch blocks with no children

### LINKED ISSUE #130: BFG has incorrect blocks
### Confirmation

### Describe the bug

There are cases where BFG may have only 1 block at a certain height, but that block is not in the "best" chain.

An example:
```sql
bfgdb=> select count(*) from btc_blocks where height = 2816533;
-[ RECORD 1 ]
count | 1

bfgdb=> select * from btc_blocks where height = 2816533;
-[ RECORD 1 ]------------------------------------------------------------------------------------------------------------------------------------------------------------------
hash       | \x393c616fd93a81de3b678a5c81020a89e726d227f2c669dc566b180000000000
header     | \x00c00522b4bd1ee0e973e5761a258cb7e968c4f309a4c2f1ac1d93dd0700000000000000d7b38c2373a86a80c854f7cc8509dad4004c6e6404c04086aae10ea770f58eaa27434b66ffff001dcc02ab9c
height     | 2816533
created_at | 2024-05-20 12:33:50.138647
updated_at | 
```

and you can see there is a different block at that height on the "best chain"
https://blockstream.info/testnet/block/0000000000002b2c603cc62223bda9440d304d4d3afe4e7c3723614ba6b061f0

furthermore, it would (obviously) seem that that block is a parent to no others:
```sql
bfgdb=> select * from btc_blocks where substr(header, 5, 32) = '\x393c616fd93a81de3b678a5c81020a89e726d227f2c669dc566b180000000000';
(0 rows)
```

this would indicate there was likely a fork which invalidated our block.  in BFG we have no way go "go back" and get the correct block, as we're always iterating up the chain.

### Expected behaviour

if another block occurs at the same height and is on the "best chain" btc needs to recognize that

### Environment

Version: 
Operating System: 
Architecture: 

### Additional Information

_No response_

ClaytonNorthey92:
**Summary**
in BFG, we only increment block height to get each block in the chain; if there is a fork, we don't handle for it.  

**Changes**
this introduces a check for blocks with no children, it will then "re-process" the block at that height from electrumx

fixes #130

### DIFF
--- a/database/bfgd/database_ext_test.go
+++ b/database/bfgd/database_ext_test.go
@@ -1687,6 +1687,7 @@ func TestBtcHeightsNoChildren(t *testing.T) {
 
 	createBlocksWithNoChildren := func(ctx context.Context, count int, db bfgd.Database) []int64 {
 		heights := make([]int64, count)
+		var prevHash []byte
 		for i := range count {
 			height := mathrand.Int64()
 			hash := make([]byte, 32)
@@ -1694,8 +1695,14 @@ func TestBtcHeightsNoChildren(t *testing.T) {
 				t.Fatal(err)
 			}
 			header := make([]byte, 80)
-			if _, err := rand.Read(header); err != nil {
-				t.Fatal(err)
+			for {
+				if _, err := rand.Read(header); err != nil {
+					t.Fatal(err)
+				}
+
+				if !slices.Equal(prevHash, header[5:37]) {
+					break
+				}
 			}
 
 			btcBlock := bfgd.BtcBlock{
@@ -1709,6 +1716,7 @@ func TestBtcHeightsNoChildren(t *testing.T) {
 			}
 
 			heights[i] = height
+			prevHash = hash
 		}
 
 		return heights
@@ -1736,16 +1744,9 @@ func TestBtcHeightsNoChildren(t *testing.T) {
 			if _, err := rand.Read(hash); err != nil {
 				t.Fatal(err)
 			}
-
 			header := make([]byte, 80)
-			for {
-				if _, err := rand.Read(header); err != nil {
-					t.Fatal(err)
-				}
-
-				if !slices.Equal(prevHash, header[5:37]) {
-					break
-				}
+			if _, err := rand.Read(header); err != nil {
+				t.Fatal(err)
 			}
 
 			if len(prevHash) > 0 {
@@ -1810,7 +1811,7 @@ func TestBtcHeightsNoChildren(t *testing.T) {
 			slices.Sort(heights)
 			slices.Sort(toCmp)
 
-			// we return a nil slice if empty, change that here for deep.Equal
+			// we return a nil slice if emtpy, change that here for deep.Equal
 			if len(heights) == 0 {
 				heights = []uint64{}
 			}
--- a/database/bfgd/database_ext_test.go
+++ b/database/bfgd/database_ext_test.go
@@ -1743,7 +1743,7 @@ func TestBtcHeightsNoChildren(t *testing.T) {
 					t.Fatal(err)
 				}
 
-				if !slices.Equal(hash, header[5:37]) {
+				if !slices.Equal(prevHash, header[5:37]) {
 					break
 				}
 			}
--- a/database/bfgd/database_ext_test.go
+++ b/database/bfgd/database_ext_test.go
@@ -1653,6 +1653,168 @@ func TestL2BtcFinalitiesByL2KeystoneNotPublishedHeight(t *testing.T) {
 	}
 }
 
+func TestBtcHeightsNoChildren(t *testing.T) {
+	type testTableItem struct {
+		name                         string
+		numberToCreateWithChildren   int
+		numberToCreateWithNoChildren int
+		overlapCount                 int
+	}
+
+	testTable := []testTableItem{
+		{
+			name:                         "0",
+			numberToCreateWithNoChildren: 0,
+			numberToCreateWithChildren:   43,
+		},
+		{
+			name:                         "less than 100",
+			numberToCreateWithNoChildren: 76,
+			numberToCreateWithChildren:   4,
+		},
+		{
+			name:                         "more than 100",
+			numberToCreateWithNoChildren: 126,
+			numberToCreateWithChildren:   333,
+		},
+		{
+			name:                         "more than 100 and overlap",
+			numberToCreateWithNoChildren: 126,
+			numberToCreateWithChildren:   333,
+			overlapCount:                 98,
+		},
+	}
+
+	createBlocksWithNoChildren := func(ctx context.Context, count int, db bfgd.Database) []int64 {
+		heights := make([]int64, count)
+		for i := range count {
+			height := mathrand.Int64()
+			hash := make([]byte, 32)
+			if _, err := rand.Read(hash); err != nil {
+				t.Fatal(err)
+			}
+			header := make([]byte, 80)
+			if _, err := rand.Read(header); err != nil {
+				t.Fatal(err)
+			}
+
+			btcBlock := bfgd.BtcBlock{
+				Height: uint64(height),
+				Hash:   hash,
+				Header: header,
+			}
+
+			if err := db.BtcBlockInsert(ctx, &btcBlock); err != nil {
+				t.Fatal(err)
+			}
+
+			heights[i] = height
+		}
+
+		return heights
+	}
+
+	createBlocksWithChildren := func(ctx context.Context, count int, db bfgd.Database, avoidHeights []int64, overlapHeights []int64) []int64 {
+		prevHash := []byte{}
+		overlapHeightI := 0
+		heights := make([]int64, count)
+		for i := range count {
+			var height int64
+			for {
+				if overlapHeightI < len(overlapHeights) {
+					height = overlapHeights[overlapHeightI]
+					overlapHeightI++
+					break
+				}
+
+				height = mathrand.Int64()
+				if !slices.Contains(avoidHeights, height) {
+					break
+				}
+			}
+			hash := make([]byte, 32)
+			if _, err := rand.Read(hash); err != nil {
+				t.Fatal(err)
+			}
+			header := make([]byte, 80)
+			if _, err := rand.Read(header); err != nil {
+				t.Fatal(err)
+			}
+
+			if len(prevHash) > 0 {
+				for k := range 32 {
+					header[k+4] = prevHash[k]
+				}
+			}
+
+			btcBlock := bfgd.BtcBlock{
+				Height: uint64(height),
+				Hash:   hash,
+				Header: header,
+			}
+
+			if err := db.BtcBlockInsert(ctx, &btcBlock); err != nil {
+				t.Fatal(err)
+			}
+			prevHash = hash
+			heights[i] = height
+		}
+		return heights
+	}
+
+	for _, tti := range testTable {
+		t.Run(tti.name, func(t *testing.T) {
+			ctx, cancel := defaultTestContext()
+			defer cancel()
+
+			db, sdb, cleanup := createTestDB(ctx, t)
+			defer func() {
+				db.Close()
+				sdb.Close()
+				cleanup()
+			}()
+
+			var overlapHeights []int64
+			noChildrenHeights := createBlocksWithNoChildren(ctx, tti.numberToCreateWithNoChildren, db)
+
+			childrenHeights := createBlocksWithChildren(ctx, tti.numberToCreateWithChildren, db, nil, overlapHeights)
+
+			if tti.overlapCount > 0 {
+				overlapHeights = noChildrenHeights[:tti.overlapCount]
+				oldChildrenHeights := childrenHeights
+				for _, o := range oldChildrenHeights {
+					if !slices.Contains(overlapHeights, o) {
+						childrenHeights = append(childrenHeights, o)
+					}
+				}
+			}
+
+			heights, err := db.BtcBlocksHeightsWithNoChildren(ctx)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			toCmp := make([]uint64, len(noChildrenHeights)+1)
+			for i, c := range noChildrenHeights {
+				toCmp[i] = uint64(c)
+			}
+			toCmp[len(toCmp)-1] = uint64(childrenHeights[len(childrenHeights)-1])
+
+			slices.Sort(heights)
+			slices.Sort(toCmp)
+
+			// we return a nil slice if emtpy, change that here for deep.Equal
+			if len(heights) == 0 {
+				heights = []uint64{}
+			}
+
+			if diff := deep.Equal(toCmp[:len(toCmp)-1], heights); len(diff) != 0 {
+				t.Fatalf("unexpected diff %s", diff)
+			}
+		})
+	}
+}
+
 func createBtcBlock(ctx context.Context, t *testing.T, db bfgd.Database, count int, chain bool, height int, lastHash []byte, l2BlockNumber uint32) bfgd.BtcBlock {
 	header := make([]byte, 80)
 	hash := make([]byte, 32)
--- a/database/bfgd/database_ext_test.go
+++ b/database/bfgd/database_ext_test.go
@@ -1715,7 +1715,7 @@ func TestBtcHeightsNoChildren(t *testing.T) {
 	}
 
 	createBlocksWithChildren := func(ctx context.Context, count int, db bfgd.Database, avoidHeights []int64, overlapHeights []int64) []int64 {
-		prevHash := []byte{}
+		var prevHash []byte
 		overlapHeightI := 0
 		heights := make([]int64, count)
 		for i := range count {
--- a/database/bfgd/database_ext_test.go
+++ b/database/bfgd/database_ext_test.go
@@ -1687,7 +1687,6 @@ func TestBtcHeightsNoChildren(t *testing.T) {
 
 	createBlocksWithNoChildren := func(ctx context.Context, count int, db bfgd.Database) []int64 {
 		heights := make([]int64, count)
-		var prevHash []byte
 		for i := range count {
 			height := mathrand.Int64()
 			hash := make([]byte, 32)
@@ -1695,14 +1694,8 @@ func TestBtcHeightsNoChildren(t *testing.T) {
 				t.Fatal(err)
 			}
 			header := make([]byte, 80)
-			for {
-				if _, err := rand.Read(header); err != nil {
-					t.Fatal(err)
-				}
-
-				if !slices.Equal(prevHash, header[5:37]) {
-					break
-				}
+			if _, err := rand.Read(header); err != nil {
+				t.Fatal(err)
 			}
 
 			btcBlock := bfgd.BtcBlock{
@@ -1716,7 +1709,6 @@ func TestBtcHeightsNoChildren(t *testing.T) {
 			}
 
 			heights[i] = height
-			prevHash = hash
 		}
 
 		return heights
--- a/database/bfgd/database_ext_test.go
+++ b/database/bfgd/database_ext_test.go
@@ -1736,9 +1736,16 @@ func TestBtcHeightsNoChildren(t *testing.T) {
 			if _, err := rand.Read(hash); err != nil {
 				t.Fatal(err)
 			}
+
 			header := make([]byte, 80)
-			if _, err := rand.Read(header); err != nil {
-				t.Fatal(err)
+			for {
+				if _, err := rand.Read(header); err != nil {
+					t.Fatal(err)
+				}
+
+				if !slices.Equal(hash, header[5:37]) {
+					break
+				}
 			}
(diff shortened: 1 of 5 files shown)


service/bfg/bfg.go
```diff
@@ -192,6 +195,35 @@ func NewServer(cfg *Config) (*Server, error) {
...
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-s.checkForInvalidBlocks:
+			heights, err := s.db.BtcBlocksHeightsWithNoChildren(ctx)
+			if err != nil {
+				log.Errorf("error trying to get heights for btc blocks: %s", err)
+			} else {
+				log.Infof("received %d heights with no children, will re-check", len(heights))
```
ClaytonNorthey92, line 215: probably should ignore the tip here as to not blow up the logs

database/bfgd/postgres/postgres.go
```diff
@@ -965,6 +965,64 @@ func (p *pgdb) AccessPublicKeyDelete(ctx context.Context, publicKey *bfgd.Access
...
+		AND EXISTS(
+			SELECT * FROM btc_blocks bb4 WHERE substr(bb4.header, 5, 32) = bb3.hash
+		)
+	)
+	ORDER BY HEIGHT DESC
+	OFFSET $1 + 1
+	LIMIT 100
+	`
+
+	heights := []uint64{}
```
joshuasing, line 997: nit: use nil slice
```suggestion
	var heights []uint64
```

database/bfgd/postgres/postgres.go
```diff
@@ -965,6 +965,64 @@ func (p *pgdb) AccessPublicKeyDelete(ctx context.Context, publicKey *bfgd.Access
...
+
+	WHERE NOT EXISTS (SELECT * FROM btc_blocks bb2 WHERE SUBSTR(bb2.header, 5, 32) = bb1.hash)
+	
+	AND NOT EXISTS (
+		SELECT * FROM btc_blocks bb3 WHERE bb1.height = bb3.height 
+		AND EXISTS(
+			SELECT * FROM btc_blocks bb4 WHERE substr(bb4.header, 5, 32) = bb3.hash
+		)
+	)
+	ORDER BY HEIGHT DESC
```
ClaytonNorthey92, line 992: `HEIGHT` --> `height`

database/bfgd/postgres/postgres.go
ClaytonNorthey92: hey @joshuasing I did this on purpose; if there are no results, I just wanted to return an empty slice, as opposed to `nil` (so you could check `if len(result/heights) == 0` for example), do you feel strongly that this should be `nil`?

joshuasing (changes requested):

database/bfgd/database_ext_test.go
```diff
@@ -1653,6 +1653,168 @@ func TestL2BtcFinalitiesByL2KeystoneNotPublishedHeight(t *testing.T) {
...
+			}
+
+			heights[i] = height
+		}
+
+		return heights
+	}
+
+	createBlocksWithChildren := func(ctx context.Context, count int, db bfgd.Database, avoidHeights []int64, overlapHeights []int64) []int64 {
+		prevHash := []byte{}
```
joshuasing, line 1718: I think this could be changed to use a nil slice:
```suggestion
		var prevHash []byte
```

database/bfgd/postgres/postgres.go
```diff
@@ -965,6 +965,62 @@ func (p *pgdb) AccessPublicKeyDelete(ctx context.Context, publicKey *bfgd.Access
...
+func (p *pgdb) BtcBlocksHeightsWithNoChildren(ctx context.Context) ([]uint64, error) {
+	log.Tracef("BtcBlocksHeightsWithNoChildren")
+	defer log.Tracef("BtcBlocksHeightsWithNoChildren exit")
+
+	/*
+		give me all heights for btc blocks
+		where there does not exist a child of this block
+		and there are no other blocks at this height with children
+		exclude the tip because it won't have children by definition
+	*/
```
joshuasing, line 980: I think this comment could be improved, additionally I think `//` should be used instead of C-style comments:
```suggestion
	// Query all heights from btc_blocks where the block does not have any
	// children and there are no other blocks at the same height with children.
	// Excludes the tip because it will not have any children.
```

database/bfgd/postgres/postgres.go
```diff
@@ -965,6 +965,62 @@ func (p *pgdb) AccessPublicKeyDelete(ctx context.Context, publicKey *bfgd.Access
...
+		
+		AND NOT EXISTS (
+			SELECT * FROM btc_blocks bb3 WHERE bb1.height = bb3.height 
+			AND EXISTS(
+				SELECT * FROM btc_blocks bb4 WHERE substr(bb4.header, 5, 32) = bb3.hash
+			)
+		)
+		ORDER BY height DESC
+		OFFSET $1 + 1
+		LIMIT 100
```
joshuasing, line 994: Is there anyway we could further optimise this query? It appears to select from `btc_blocks` four times.

Also there's a few inconsistencies that I think we could fix:
```suggestion
		SELECT height FROM btc_blocks bb1
		WHERE NOT EXISTS (SELECT * FROM btc_blocks bb2 WHERE substr(bb2.header, 5, 32) = bb1.hash)
		AND NOT EXISTS (
			SELECT * FROM btc_blocks bb3 WHERE bb1.height = bb3.height 
			AND EXISTS (
				SELECT * FROM btc_blocks bb4 WHERE substr(bb4.header, 5, 32) = bb3.hash
			)
		)
		ORDER BY height DESC
		OFFSET $1 + 1
		LIMIT 100
```

database/bfgd/postgres/postgres.go
```diff
@@ -965,6 +965,62 @@ func (p *pgdb) AccessPublicKeyDelete(ctx context.Context, publicKey *bfgd.Access
 	return nil
 }
 
+// BtcBlocksHeightsWithNoChildren returns the heights of blocks that have no
+// child blocks in our database, these represent possible forks that we have
+// not handled yet
```
joshuasing, line 970: I think we could improve the clarity of this comment a bit, what do you think of the following?
```suggestion
// BtcBlocksHeightsWithNoChildren returns the heights of blocks stored in the
// database that do not have any children, these represent possible forks that
// have not been handled yet.
```

database/bfgd/postgres/postgres.go
ClaytonNorthey92: hey @joshuasing ok so I did some digging into whether we can optimize this.

I tested this locally, so I ran `EXPLAIN` with both `set enable_seqscan = off` and `set enable_seqscan = on`.

With that `on`, meaning we don't discourage the use of seq scans, the planner doesn't help much, as I have so few rows in my local db that it determines the best way to query is to use seq scans.
```sql
                                                 QUERY PLAN                                                 
------------------------------------------------------------------------------------------------------------
 Limit  (cost=112.47..112.72 rows=100 width=8)
   ->  Sort  (cost=112.47..112.90 rows=175 width=8)
         Sort Key: bb1.height DESC
         ->  Hash Anti Join  (cost=76.59..105.95 rows=175 width=8)
               Hash Cond: (bb1.height = bb3.height)
               ->  Hash Right Anti Join  (cost=25.75..48.96 rows=350 width=8)
                     Hash Cond: (substr(bb2.header, 5, 32) = bb1.hash)
                     ->  Seq Scan on btc_blocks bb2  (cost=0.00..17.00 rows=700 width=32)
                     ->  Hash  (cost=17.00..17.00 rows=700 width=40)
                           ->  Seq Scan on btc_blocks bb1  (cost=0.00..17.00 rows=700 width=40)
               ->  Hash  (cost=46.46..46.46 rows=350 width=8)
                     ->  Hash Join  (cost=23.25..46.46 rows=350 width=8)
                           Hash Cond: (bb3.hash = substr(bb4.header, 5, 32))
                           ->  Seq Scan on btc_blocks bb3  (cost=0.00..17.00 rows=700 width=40)
                           ->  Hash  (cost=20.75..20.75 rows=200 width=32)
                                 ->  HashAggregate  (cost=18.75..20.75 rows=200 width=32)
                                       Group Key: substr(bb4.header, 5, 32)
                                       ->  Seq Scan on btc_blocks bb4  (cost=0.00..17.00 rows=700 width=32)
(18 rows)
```

so, if I _discourage_ the use of seq scans, I can see postgres uses the following indexes:
```sql
                                                         QUERY PLAN                                                         
----------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=203.42..203.67 rows=100 width=8)
   ->  Sort  (cost=203.42..203.85 rows=175 width=8)
         Sort Key: bb1.height DESC
         ->  Hash Anti Join  (cost=167.54..196.90 rows=175 width=8)
               Hash Cond: (bb1.height = bb3.height)
               ->  Hash Right Anti Join  (cost=71.22..94.44 rows=350 width=8)
                     Hash Cond: (substr(bb2.header, 5, 32) = bb1.hash)
                     ->  Bitmap Heap Scan on btc_blocks bb2  (cost=7.83..24.82 rows=700 width=32)
                           ->  Bitmap Index Scan on btc_blocks_hash  (cost=0.00..7.65 rows=700 width=0)
                     ->  Hash  (cost=54.65..54.65 rows=700 width=40)
                           ->  Index Scan using height_idx on btc_blocks bb1  (cost=0.15..54.65 rows=700 width=40)
               ->  Hash  (cost=91.94..91.94 rows=350 width=8)
                     ->  Hash Join  (cost=31.22..91.94 rows=350 width=8)
                           Hash Cond: (bb3.hash = substr(bb4.header, 5, 32))
                           ->  Index Scan Backward using height_idx on btc_blocks bb3  (cost=0.15..54.65 rows=700 width=40)
                           ->  Hash  (cost=28.57..28.57 rows=200 width=32)
                                 ->  HashAggregate  (cost=26.57..28.57 rows=200 width=32)
                                       Group Key: substr(bb4.header, 5, 32)
                                       ->  Bitmap Heap Scan on btc_blocks bb4  (cost=7.83..24.82 rows=700 width=32)
                                             ->  Bitmap Index Scan on btc_blocks_hash  (cost=0.00..7.65 rows=700 width=0)
(20 rows)
```

whilst we do `SELECT FROM btc_blocks` quite a bit, the query plan is given the _option_ to use the indexes  `height_idx` and `btc_blocks_hash` to optimize the query.


so, given there are no sql changes, I can run this on testnet to see what we would do there:
```sql
                                                                                       QUERY PLAN                                                                                        
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=76187.25..76198.91 rows=100 width=8) (actual time=508.866..520.173 rows=74 loops=1)
   ->  Gather Merge  (cost=76187.13..82027.40 rows=50056 width=8) (actual time=508.863..520.165 rows=75 loops=1)
         Workers Planned: 2
         Workers Launched: 2
         ->  Sort  (cost=75187.11..75249.68 rows=25028 width=8) (actual time=505.763..505.771 rows=25 loops=3)
               Sort Key: bb1.height DESC
               Sort Method: quicksort  Memory: 25kB
               Worker 0:  Sort Method: quicksort  Memory: 25kB
               Worker 1:  Sort Method: quicksort  Memory: 25kB
               ->  Parallel Hash Right Anti Join  (cost=67606.48..74228.76 rows=25028 width=8) (actual time=505.535..505.713 rows=25 loops=3)
                     Hash Cond: (substr(bb2.header, 5, 32) = bb1.hash)
                     ->  Parallel Seq Scan on btc_blocks bb2  (cost=0.00..5913.14 rows=100114 width=81) (actual time=0.005..6.826 rows=80091 loops=3)
                     ->  Parallel Hash  (cost=66980.77..66980.77 rows=50057 width=41) (actual time=484.837..484.843 rows=25 loops=3)
                           Buckets: 131072  Batches: 1  Memory Usage: 1120kB
                           ->  Parallel Hash Anti Join  (cost=60247.95..66980.77 rows=50057 width=41) (actual time=450.641..484.612 rows=25 loops=3)
                                 Hash Cond: (bb1.height = bb3.height)
                                 ->  Parallel Seq Scan on btc_blocks bb1  (cost=0.00..5913.14 rows=100114 width=41) (actual time=0.011..8.728 rows=80091 loops=3)
                                 ->  Parallel Hash  (cost=59622.23..59622.23 rows=50057 width=8) (actual time=450.053..450.057 rows=80064 loops=3)
                                       Buckets: 262144 (originally 131072)  Batches: 1 (originally 1)  Memory Usage: 12544kB
                                       ->  Hash Join  (cost=47775.58..59622.23 rows=50057 width=8) (actual time=308.685..395.255 rows=80064 loops=3)
                                             Hash Cond: (bb3.hash = (substr(bb4.header, 5, 32)))
                                             ->  Parallel Seq Scan on btc_blocks bb3  (cost=0.00..5913.14 rows=100114 width=41) (actual time=0.011..9.082 rows=80091 loops=3)
                                             ->  Hash  (cost=41487.17..41487.17 rows=240273 width=81) (actual time=308.364..308.367 rows=240269 loops=3)
                                                   Buckets: 65536 (originally 65536)  Batches: 8 (originally 4)  Memory Usage: 7681kB
                                                   ->  Unique  (cost=40285.81..41487.17 rows=240273 width=81) (actual time=153.986..244.353 rows=240269 loops=3)
                                                         ->  Sort  (cost=40285.81..40886.49 rows=240273 width=81) (actual time=153.985..207.116 rows=240273 loops=3)
                                                               Sort Key: (substr(bb4.header, 5, 32))
                                                               Sort Method: external merge  Disk: 29176kB
                                                               Worker 0:  Sort Method: external merge  Disk: 29176kB
                                                               Worker 1:  Sort Method: external merge  Disk: 29176kB
                                                               ->  Seq Scan on btc_blocks bb4  (cost=0.00..7314.73 rows=240273 width=81) (actual time=0.016..47.489 rows=240273 loops=3)
 Planning Time: 0.397 ms
 Execution Time: 526.158 ms
(33 rows)
```

that's not good; that cost is too high, so if I try to add and functional index on testnet (don't worry I'll remove it)

```sql
CREATE INDEX btc_blocks_header_prev_hash_idx ON btc_blocks (substr(header, 5, 32));
```

```sql
                                                                                    QUERY PLAN                                                                                    
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=1051.40..6021.14 rows=100 width=8) (actual time=321.047..325.646 rows=74 loops=1)
   ->  Nested Loop Anti Join  (cost=1001.70..2986224.09 rows=60068 width=8) (actual time=319.796..325.638 rows=75 loops=1)
         ->  Gather Merge  (cost=1000.86..1175528.49 rows=120136 width=8) (actual time=319.736..325.059 rows=80 loops=1)
               Workers Planned: 2
               Workers Launched: 2
               ->  Nested Loop Anti Join  (cost=0.84..1160661.79 rows=50057 width=8) (actual time=1.452..211.113 rows=27 loops=3)
                     ->  Parallel Index Scan using height_idx on btc_blocks bb1  (cost=0.42..11868.81 rows=100114 width=41) (actual time=0.013..24.873 rows=80091 loops=3)
                     ->  Index Scan using btc_blocks_header_prev_hash_idx on btc_blocks bb2  (cost=0.42..21.55 rows=1201 width=81) (actual time=0.002..0.002 rows=1 loops=240273)
                           Index Cond: (substr(header, 5, 32) = bb1.hash)
         ->  Nested Loop Semi Join  (cost=0.84..15.06 rows=1 width=8) (actual time=0.007..0.007 rows=0 loops=80)
               ->  Index Scan using height_idx on btc_blocks bb3  (cost=0.42..0.54 rows=1 width=41) (actual time=0.003..0.003 rows=1 loops=80)
                     Index Cond: (height = bb1.height)
               ->  Index Scan using btc_blocks_header_prev_hash_idx on btc_blocks bb4  (cost=0.42..21.55 rows=1201 width=81) (actual time=0.003..0.003 rows=0 loops=84)
                     Index Cond: (substr(header, 5, 32) = bb3.hash)
 Planning Time: 0.590 ms
 Execution Time: 325.682 ms
(16 rows)
```

I get 1/75 of the cost! wow! ok so I am going to add that index to the pr, good call 👍 🤯

database/bfgd/postgres/postgres.go
ClaytonNorthey92: ah so @joshuasing I was trying to align each line in the comment with a line in the query, perhaps I should add comments within the query itself? 🤔

joshuasing: I think the a comment like this above the query is sufficient and makes it easy to understand what the query is doing without the need to read the query.

I think with the consistency changes suggested in the following comment, the query is also readable enough for people to understand without the need for extra comments. https://github.com/hemilabs/heminetwork/pull/131#discussion_r1618894550

service/bfg/bfg.go
```diff
@@ -162,6 +165,37 @@ func NewServer(cfg *Config) (*Server, error) {
...
+			}
+
+			log.Infof("received %d heights with no children, will re-check", len(heights))
+			for _, height := range heights {
+				log.Infof("reprocessing block at height %d", height)
+				if err := s.processBitcoinBlock(ctx, height); err != nil {
+					log.Errorf("error processing bitcoin block: %s", err)
+				}
+			}
+
```
joshuasing, line 194: nit: could we remove this extra newline?

ClaytonNorthey92: yup!

database/bfgd/database_ext_test.go
```diff
@@ -1653,6 +1653,168 @@ func TestL2BtcFinalitiesByL2KeystoneNotPublishedHeight(t *testing.T) {
...
+		for i := range count {
+			height := mathrand.Int64()
+			hash := make([]byte, 32)
+			if _, err := rand.Read(hash); err != nil {
+				t.Fatal(err)
+			}
+			header := make([]byte, 80)
+			if _, err := rand.Read(header); err != nil {
+				t.Fatal(err)
+			}
```
joshuasing, line 1699: As more of a note for later, as these are completely random there is no guarantee that the header won't have a previous hash that matches an existing hash, which could cause unexpected behaviour.

This would be extremely rare, but could cause test failures so I think it's worth noting.

database/bfgd/database_ext_test.go
```diff
@@ -1653,6 +1653,168 @@ func TestL2BtcFinalitiesByL2KeystoneNotPublishedHeight(t *testing.T) {
...
+			toCmp := make([]uint64, len(noChildrenHeights)+1)
+			for i, c := range noChildrenHeights {
+				toCmp[i] = uint64(c)
+			}
+			toCmp[len(toCmp)-1] = uint64(childrenHeights[len(childrenHeights)-1])
+
+			slices.Sort(heights)
+			slices.Sort(toCmp)
+
+			// we return a nil slice if emtpy, change that here for deep.Equal
```
joshuasing, line 1806: There's a typo here, oops:
```suggestion
			// we return a nil slice if empty, change that here for deep.Equal
```

database/bfgd/postgres/postgres.go
```diff
@@ -965,6 +965,57 @@ func (p *pgdb) AccessPublicKeyDelete(ctx context.Context, publicKey *bfgd.Access
...
+				SELECT * FROM btc_blocks bb4 WHERE substr(bb4.header, 5, 32) = bb3.hash
+			)
+		)
+		ORDER BY height DESC
+		OFFSET $1 + 1
+		LIMIT 100
+	`
+
+	var heights []uint64
+	for offset := 0; ; offset += 100 {
```
joshuasing, line 993: Does querying this data in groups of 100 have any impact on performance, or is it for another reason? 🤔

joshuasing (approved):

database/bfgd/postgres/postgres.go
ClaytonNorthey92: meh it was just a way to implement pagination, the number was somewhat arbitrary; the last time I looked at testnet, we're actually at about 80 total blocks affected by this, my thinking was that we will probably never get more than 100 except for possible the first time we deploy this, so just paginate by that (and expect to not have more than a page after the first deploy, I'd guess we'd be returning 1 or 2 results but need to test)

database/bfgd/database_ext_test.go
ClaytonNorthey92: that's true, let me see if I can handle 🤔

joshuasing: I think the fix you added checks whether a block's header contains its own "hash" as the previous hash, correct me if I am wrong.

I don't think this will prevent the issue I mentioned above, where another block's header will have the "previous hash" of another generated block "hash", which would make the block appear as so it is a child of another generated block.

ClaytonNorthey92: 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦

max-sanchez (approved):

PR #142, opened 2024-06-05
ignore ulimits in tbc when on localnet

ClaytonNorthey92:
**Summary**
localnet is small, and is easy for tbc to crawl and index.  with localnet, we often run it in CI environments where we have limited or no control over ulimits.  there is no need for this in localnet, so let's ignore it if we're on localnet.

**Changes**
if network set to localnet, ignore ulimits


jcvernaleo (approved):
Seems reasonable.  If localnet ever gets that big, we are using it wrong.

OK

joshuasing (approved):

PR #144, opened 2024-06-07
popm/wasm: improve and tidy up Go code

joshuasing:
**Summary**
Initial pass to improve and tidy up the Go code for the WASM PoP Miner.
This is the first of many pull requests to improve the WASM and create a TypeScript (NPM) package that can be used to run the PoP Miner.

**Changes**
This splits the `web/popminer/popminer.go` file into 6 files:
 - `api.go`: Contains all API types
 - `dispatch.go`: Contains all methods that can be dispatched from JS.
     The majority of the code in this file was moved from `popminer.go`.
 - `network.go`: Contains Hemi/Bitcoin network settings
 - `network_local.go`: Contains localnet settings (behind build tag, build with `-tags "hemi_localnet"` to enable)
 - `popminer.go`: Contains the main function and core
 - `util.go`: Contains utility functions for working with JS.

This contains a lot of miscellaneous improvements, however the notable ones are:
 - Invoke Promise resolve function with `js.Value` instead of marshalled JSON, meaning that the result can be used directly without the need to encode/decode JSON.

 - Make all errors use a common object (created by the `jsError` function) that contains the error message, a timestamp, and the debug stack, allowing for easier debugging.

 - Replace `wasmPing` method with `version` - `wasmPing` is not very useful and was only used for testing the connection between JS and WASM. `version` returns useful version information and allows the same testing.

 - Use structs for all return types. The `jsValueOf` function supports converting structs to JavaScript Objects via reflection.

 - Move network configurations into a separate file and use constants. I think this is cleaner than the previous solution, and allows localnet to only be available when built with the 'hemi_localnet' build tag. *Note: I have found the networks to be slightly confusing, as we need to interact with Hemi and Bitcoin. This "standardises" the networks in the WASM PoP miner and uses the **Hemi network name**, not the Bitcoin network name. If this is not useful, then this can be reverted.*
 
  - Move dispatch function to under a global object to prevent collisions and other issues. - The dispatch function is now placed in a `@hemilabs/pop-miner` global object. An example of accessing the dispatch function would be `globalThis['@hemilabs/pop-miner'].dispatch(args)`. **The name `@hemilabs/pop-miner` is used, as it will be the name of the NPM package.**
 
 Other miscellaneous changes:
  - Fix typo in `cmd/popmd/popmd.go`
  - Added versioning to and cleaned up `web/Makefile`
  - Reorganised code
  - Changed default logging level to `WARN` and added ability to change log level with linker flag (e.g. `-X main.logLevel=TRACE`)
  - Renamed `runpopminer` to `startPoPMiner`
  - Use `chan struct{}` instead of `chan bool` to block application exit
  - Consistently use lowerCamelCase for method names
  - Improve some variable and function naming

 Please ask any questions and provide any feedback you have.
 *Note: The `wasm_exec.js` file is from Go. I added it to Git due to build issues I was having with certain environments, and it allows us to update the file when we want instead of it being whatever is included the user's installed Go version.*
 
 **The next pull request will contain the initial TypeScript types and their documentation.**

### DIFF
--- a/web/popminer/dispatch.go
+++ b/web/popminer/dispatch.go
@@ -274,7 +274,7 @@ func startPoPMiner(_ js.Value, args []js.Value) (any, error) {
 			// TODO(joshuasing): dispatch event on failure
 			pmMtx.Lock()
 			defer pmMtx.Unlock()
-			pm.err = err // Theoretically this can logic race unless we unset om
+			pm.err = err // Theoretically this can logic race unless we unset pm
 		}
 	}()
 
@@ -401,7 +401,7 @@ func bitcoinInfo(_ js.Value, _ []js.Value) (any, error) {
 	}, nil
 }
 
-func bitcoinUTXOs(this js.Value, args []js.Value) (any, error) {
+func bitcoinUTXOs(_ js.Value, args []js.Value) (any, error) {
 	log.Tracef("bitcoinUTXOs")
 	defer log.Tracef("bitcoinUTXOs exit")
 
--- a/web/popminer/dispatch.go
+++ b/web/popminer/dispatch.go
@@ -1,0 +1,438 @@
+// Copyright (c) 2024 Hemi Labs, Inc.
+// Use of this source code is governed by the MIT License,
+// which can be found in the LICENSE file.
+
+//go:build js && wasm
+
+package main
+
+import (
+	"context"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"runtime/debug"
+	"syscall/js"
+	"time"
+
+	"github.com/btcsuite/btcd/btcutil"
+	btcchaincfg "github.com/btcsuite/btcd/chaincfg"
+	dcrsecpk256k1 "github.com/decred/dcrd/dcrec/secp256k1/v4"
+	"github.com/juju/loggo"
+
+	"github.com/hemilabs/heminetwork/ethereum"
+	"github.com/hemilabs/heminetwork/service/popm"
+)
+
+const (
+	// The following can be dispatched at any time.
+	MethodVersion       = "version"       // Retrieve WASM version information
+	MethodGenerateKey   = "generateKey"   // Generate secp256k1 key pair
+	MethodStartPoPMiner = "startPoPMiner" // Start PoP Miner
+	MethodStopPoPMiner  = "stopPoPMiner"  // Stop PoP Miner
+
+	// The following can only be dispatched after the PoP Miner is running.
+	MethodPing           = "ping"           // Ping BFG
+	MethodL2Keystones    = "l2Keystones"    // Retrieve L2 keystones
+	MethodBitcoinBalance = "bitcoinBalance" // Retrieve bitcoin balance
+	MethodBitcoinInfo    = "bitcoinInfo"    // Retrieve bitcoin information
+	MethodBitcoinUTXOs   = "bitcoinUTXOs"   // Retrieve bitcoin UTXOs
+)
+
+// Dispatcher maps methods to the relative dispatch.
+var Dispatcher = map[string]*Dispatch{
+	MethodVersion: {
+		Handler: wasmVersion,
+	},
+	MethodGenerateKey: {
+		Handler: generateKey,
+		Required: []DispatchArgs{
+			{Name: "network", Type: js.TypeString},
+		},
+	},
+	MethodStartPoPMiner: {
+		Handler: startPoPMiner,
+		Required: []DispatchArgs{
+			{Name: "logLevel", Type: js.TypeString},
+			{Name: "network", Type: js.TypeString},
+			{Name: "privateKey", Type: js.TypeString},
+		},
+	},
+	MethodStopPoPMiner: {
+		Handler: stopPopMiner,
+	},
+
+	// The following can only be dispatched after the PoP Miner is running.
+	MethodPing: {
+		Handler: ping,
+	},
+	MethodL2Keystones: {
+		Handler: l2Keystones,
+		Required: []DispatchArgs{
+			{Name: "numL2Keystones", Type: js.TypeNumber},
+		},
+	},
+	MethodBitcoinBalance: {
+		Handler: bitcoinBalance,
+		Required: []DispatchArgs{
+			{Name: "scriptHash", Type: js.TypeString},
+		},
+	},
+	MethodBitcoinInfo: {
+		Handler: bitcoinInfo,
+	},
+	MethodBitcoinUTXOs: {
+		Handler: bitcoinUTXOs,
+		Required: []DispatchArgs{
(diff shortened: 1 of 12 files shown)


web/www/wasm_exec.js
joshuasing, line 1: This file is from Go and has not been modified in anyway, other than adding license link comment.

marcopeereboom: I am not sure I love this. This will easily get out of date. I am of the opinion that we should use the wasm_exec.js from the build machine instead. This way if the tests fail we know right away if something went awry.

web/popminer/dispatch.go
```diff
@@ -0,0 +1,438 @@
...
+		btcChainParams = &btcchaincfg.TestNet3Params
+		netNormalized = "testnet3"
+	case "mainnet":
+		btcChainParams = &btcchaincfg.MainNetParams
+		netNormalized = "mainnet"
+	default:
+		return js.Null(), fmt.Errorf("invalid network: %v", net)
+	}
+
+	privKey, err := dcrsecpk256k1.GeneratePrivateKey()
```
marcopeereboom, line 215: Not sure we want to pull in all this stuff, including some btcuti/ethereum stuff. For now this is fine but we need to see if we can pull it out and save some valuable network bandwidth.

web/popminer/dispatch.go
```diff
@@ -0,0 +1,438 @@
...
+	defer log.Tracef("startPoPMiner exit")
+
+	pmMtx.Lock()
+	defer pmMtx.Unlock()
+	if pm != nil {
+		return js.Null(), errors.New("pop miner already started")
+	}
+
+	pm = new(Miner)
+	pm.ctx, pm.cancel = context.WithCancel(context.Background())
```
marcopeereboom, line 252: Is this too late to create the context?

web/popminer/dispatch.go
```diff
@@ -0,0 +1,438 @@
...
+
+	cfg := popm.NewDefaultConfig()
+	cfg.BTCPrivateKey = args[0].Get("privateKey").String()
+	cfg.StaticFee = uint(args[0].Get("staticFee").Int())
+
+	cfg.LogLevel = args[0].Get("logLevel").String()
+	if cfg.LogLevel == "" {
+		cfg.LogLevel = "popm=ERROR"
+	}
+	loggo.ConfigureLoggers(cfg.LogLevel)
```
marcopeereboom, line 262: Shouldn't this be the very first thing to be done globally so that you can log very early on?

web/popminer/dispatch.go
```diff
@@ -0,0 +1,438 @@
...
+			pmMtx.Lock()
+			defer pmMtx.Unlock()
+			pm.err = err // Theoretically this can logic race unless we unset om
+		}
+	}()
+
+	return js.Undefined(), nil
+}
+
+func stopPopMiner(_ js.Value, _ []js.Value) (js.Value, error) {
```
marcopeereboom, line 293: shouldn't this cancel the context and wg.Wait?

web/popminer/dispatch.go
```diff
@@ -0,0 +1,438 @@
...
+	log.Tracef("l2Keystones")
+	defer log.Tracef("l2Keystones exit")
+
+	c := args[0].Get("numL2Keystones").Int()
+	if c < 0 || c > 10 {
+		c = 2
+	}
+	count := uint64(c)
+
+	activePM, err := activeMiner()
```
marcopeereboom, line 345: I am not sure I love this construct. This is silly unless we actually have named miners that are passed in via args. Which is weird in its own special way.

This is very raceable.

web/popminer/util.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+var (
+	promiseConstructor = js.Global().Get("Promise")
+	objectConstructor  = js.Global().Get("Object")
+	arrayConstructor   = js.Global().Get("Array")
+)
+
+// Object represents a JavaScript Object.
+type Object map[string]any
```
marcopeereboom, line 24: Cute. But you can directly return a map as well if that's what you are trying to achieve. I don;t think there is need for these gymnastics.

The point of having predefined return structures was that you could test them at the go level. Now this moves it up. I am ok with this but I am not seeing any advantage besides taste.

marcopeereboom (changes requested):
I am overall ok with this but would like to understand why returning a map is preferable over predefined structures.

web/popminer/dispatch.go
joshuasing: This is copied from the previous code, where would be a better place for it?
I think this is the earliest that it can be stored in the global miner.

We could also initialise the Miner object at init and create the context in the main function? 🤔

web/popminer/dispatch.go
joshuasing: Yes, it is called in the init function:
https://github.com/hemilabs/heminetwork/blob/joshua/web-popm/improve-go/web/popminer/popminer.go#L49

This behaviour is the same as before.

web/popminer/dispatch.go
joshuasing: I think it does? https://github.com/hemilabs/heminetwork/blob/joshua/web-popm/improve-go/web/popminer/dispatch.go#L306-L307

web/popminer/dispatch.go
joshuasing: This behaviour is from the previous code, do you want it changed?

web/popminer/util.go
joshuasing: Yes, a map could be used, however I did it like this so that it is both obvious that this will be represented in JS as an object (and as such the keys have meaning), and also easy to add utility functions for.

The dispatch handlers returning a `js.Value` also means that it is guaranteed that a `js.Value` can be returned when an error isn't.

web/popminer/dispatch.go
joshuasing: Added a comment.

web/popminer/util.go
joshuasing: TODO: Use structs instead.

web/popminer/dispatch.go
```diff
@@ -0,0 +1,431 @@
...
+	if err != nil {
+		return js.Null(), err
+	}
+
+	return BitcoinInfoResult{
+		Height: pr.Height,
+	}, nil
+}
+
+func bitcoinUTXOs(this js.Value, args []js.Value) (any, error) {
```
marcopeereboom, line 404: this *sigh*

marcopeereboom (approved):
I do really want some sort of detection for wasm_exec.js being out of date with the version that lives in the repo. I am even ok with storing the wasm_exec.js version in the repo and comparing it agains the build machine version. But we need some non-human detection of that file's content.

PR #152, opened 2024-06-20
Remove height as the terminal condition for indexers.

### LINKED ISSUE #141: tbc: replace various height based tables and calls with a hash
### Confirmation

### Problem

_No response_

### Suggested solution

We are using various height api and keyed tables. While this mostly works it really should be hash based.

E.g.
```
func (s *Server) SyncIndexersToHeight(ctx context.Context, height uint64) error {

var (
        UtxoIndexHeightKey = []byte("utxoindexheight") // last indexed utxo height key
        TxIndexHeightKey   = []byte("txindexheight")   // last indexed tx height key
)

func (s *Server) indexUtxosInBlocks(ctx context.Context, startHeight, maxHeight uint64, utxos map[tbcd.Outpoint]tbcd.CacheOutput) (int, error) {
```

Etc.

This is a bit of a messy change but should happen ASAP.

### Additional Information

_No response_

marcopeereboom:
**Summary**
Currently height is being used as the terminal condition for indexers. This cannot be the case since this breaks during forks.

**Changes**
This PR adds a hash as the terminal condition for the indexers and catches potential forks.

Fixes #141

### DIFF
--- a/database/tbcd/database.go
+++ b/database/tbcd/database.go
@@ -114,6 +114,19 @@ func (bh BlockHeader) Wire() (*wire.BlockHeader, error) {
 	return &wbh, nil
 }
 
+func (bh BlockHeader) BlockHash() *chainhash.Hash {
+	ch, _ := chainhash.NewHash(bh.Hash)
+	return ch
+}
+
+func (bh BlockHeader) ParentHash() *chainhash.Hash {
+	wh, err := bh.Wire()
+	if err != nil {
+		panic(err)
+	}
+	return &wh.PrevBlock
+}
+
 // Block contains a raw bitcoin block and its corresponding hash.
 type Block struct {
 	Hash  database.ByteArray
@@ -396,4 +409,34 @@ func NewTxMapping(txId, blockHash *chainhash.Hash) (txKey TxKey) {
 	copy(txKey[33:], blockHash[:])
 
 	return txKey
+}
+
+// Helper functions
+
+// B2H converts a raw bloc header to a wire block header structure.
+func B2H(header []byte) (*wire.BlockHeader, error) {
+	var bh wire.BlockHeader
+	if err := bh.Deserialize(bytes.NewReader(header)); err != nil {
+		return nil, fmt.Errorf("deserialize block header: %w", err)
+	}
+	return &bh, nil
+}
+
+// HeaderHash return the block hash from a raw block header.
+func HeaderHash(header []byte) *chainhash.Hash {
+	h, err := B2H(header)
+	if err != nil {
+		panic(err)
+	}
+	hash := h.BlockHash()
+	return &hash
+}
+
+// HeaderHash return the parent block hash from a raw block header.
+func HeaderParentHash(header []byte) *chainhash.Hash {
+	h, err := B2H(header)
+	if err != nil {
+		panic(err)
+	}
+	return &h.PrevBlock
 }
--- a/service/tbc/crawler.go
+++ b/service/tbc/crawler.go
@@ -7,7 +7,6 @@ package tbc
 import (
 	"context"
 	"crypto/sha256"
-	"encoding/binary"
 	"errors"
 	"fmt"
 	"runtime"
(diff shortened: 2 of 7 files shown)


database/tbcd/database.go
```diff
@@ -397,3 +410,33 @@ func NewTxMapping(txId, blockHash *chainhash.Hash) (txKey TxKey) {
 
 	return txKey
 }
+
+// Helper functions
+
+// B2H converts a raw bloc header to a wire block header structure.
```
gndelia, line 416: minor spelling mistake? 😅 
```suggestion
// B2H converts a raw block header to a wire block header structure.
```

ClaytonNorthey92 (changes requested):
hey @marcopeereboom the only thing I would ask that we change is checking for that error that's ignored

database/tbcd/database.go
```diff
@@ -397,3 +410,33 @@ func NewTxMapping(txId, blockHash *chainhash.Hash) (txKey TxKey) {
...
+func B2H(header []byte) (*wire.BlockHeader, error) {
```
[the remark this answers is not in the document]
marcopeereboom: indeed.

database/tbcd/database.go
```diff
@@ -114,6 +114,19 @@ func (bh BlockHeader) Wire() (*wire.BlockHeader, error) {
 	return &wbh, nil
 }
 
+func (bh BlockHeader) BlockHash() *chainhash.Hash {
+	ch, _ := chainhash.NewHash(bh.Hash)
```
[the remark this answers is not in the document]
marcopeereboom: Sure.

service/tbc/crawler.go
```diff
@@ -124,94 +161,149 @@ func (s *Server) fixupCache(ctx context.Context, b *btcutil.Block, utxos map[tbc
...
+	// forced for.
+	// XXX check cumulative? check fork?
 
 	// Allocate here so that we don't waste space when not indexing.
 	utxos := make(map[tbcd.Outpoint]tbcd.CacheOutput, s.cfg.MaxCachedTxs)
 	defer clear(utxos)
 
-	log.Infof("Start indexing UTxos at height %v count %v", height, count)
+	log.Infof("Start indexing UTxos at hash %v height %v", utxoHH.Hash, utxoHH.Height)
+	log.Infof("End indexing UTxos at hash %v height %v", endHash, endHeight)
```
joshuasing, line 303: Should these logs be at the `info` level? If they are primarily for debugging, I think the `debug` level would be better.

service/tbc/crawler.go
```diff
@@ -274,92 +361,142 @@ func processTxs(cp *chaincfg.Params, blockHash *chainhash.Hash, txs []*btcutil.T
...
+
 	// Allocate here so that we don't waste space when not indexing.
 	txs := make(map[tbcd.TxKey]*tbcd.TxValue, s.cfg.MaxCachedTxs)
-	// log.Infof("max %v %v", s.cfg.MaxCachedTxs, s.cfg.MaxCachedTxs*(105))
-	// return nil
 	defer clear(txs)
 
-	log.Infof("Start indexing transactions at height %v count %v", height, count)
+	log.Infof("Start indexing Txs at hash %v height %v", txHH.Hash, txHH.Height)
+	log.Infof("End indexing Txs at hash %v height %v", endHash, endHeight)
```
joshuasing, line 496: Should these logs be at the `info` level? If they are primarily for debugging, I think the `debug` level would be better.

joshuasing (approved):

service/tbc/crawler.go
marcopeereboom: I have actually gone back and forth on this. If we don't print it we have absolutely no idea how long it'll run or how much needs to be done. In principle I agree with you but we need a start and end indication the way the code works now.

service/tbc/crawler.go
marcopeereboom: I just typed a few sentence here explaining why but github ate them.

Basically, we need to know how long we will be running.

service/tbc/crawler.go
joshuasing: Ah okay, that's fine. I just don't want to end up with random unneeded Info logs :)

service/tbc/crawler.go
joshuasing: See response: https://github.com/hemilabs/heminetwork/pull/152#discussion_r1650879693

ClaytonNorthey92 (approved):

PR #153, opened 2024-06-20
Use hemilabs/websocket fork of nhooyr.io/websocket

joshuasing:
**Summary**
Use https://github.com/hemilabs/websocket fork of `nhooyr.io/websocket` library

The WASM implementation of the WebSocket connection in `nhooyr.io/websocket` contains a bug that causes `net.ErrClosed` to always be returned when the connection is closed, causing the underlying `websocket.CloseError` to never be returned. This fork contains a fix that returns the close error correctly. This patch will hopefully be upstreamed at a later date, once we are able to confirm that this fix is working correctly and doesn't change anything else.

Tested by running the `web/integrationtest` and providing a private key that is not authorised on the BFG server.
The exit behaviour now matches the `popmd` daemon.

**Changes**
 - Use `github.com/hemilabs/websocket` instead of `nhooyr.io/websocket` (likely temporarily)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,10 @@ module github.com/hemilabs/heminetwork
 
 go 1.22.2
 
+toolchain go1.22.3
+
+replace nhooyr.io/websocket v1.8.11 => github.com/hemilabs/websocket v0.0.0-20240620132401-b5109a38f904
+
 require (
 	github.com/btcsuite/btcd v0.24.0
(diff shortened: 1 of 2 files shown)


jcvernaleo (approved):
OK

marcopeereboom (approved):
I really would like to use upstream code but for now this makes sense.

PR #154, opened 2024-06-21
popm/wasm: add error codes

joshuasing:
**Summary**
Add error codes to the WebAssembly PoP Miner.
These error codes will be used to programatically handle errors and display useful information to users.

The default error code is ErrorCodeInternal ("internal").

**Changes**
 - Add `ErrorCode` type and add `Code` field to the existing `Error` type
 - Add `ErrorCodeInternal` for use with internal errors (and is the default error code)
 - Add `ErrorCodeInvalidValue` for use when an invalid value is provided for an argument, currently only used in `generateKey`
 - Refactor how some elements are displayed in integration test web page to help with testing
 - Add `codedError` type, used to include an error code in an error that is extracted when creating an `Error` that will be converted to a JavaScript Object

More error codes will be added as needed in future pull requests.

### DIFF
--- a/web/popminer/api.go
+++ b/web/popminer/api.go
@@ -24,23 +24,35 @@ const (
 	MethodBitcoinUTXOs   Method = "bitcoinUTXOs"   // Retrieve bitcoin UTXOs
 )
 
-// ErrorCode is a unique identifier used to differentiate between error types.
-type ErrorCode string
+// ErrorCode is used to differentiate between error types.
+type ErrorCode uint32
 
 const (
 	// ErrorCodeInternal is used when the error is internal, either due to an
 	// invalid dispatch or a panic.
-	ErrorCodeInternal ErrorCode = "internal"
+	ErrorCodeInternal ErrorCode = 0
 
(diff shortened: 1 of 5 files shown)


web/popminer/api.go
```diff
@@ -24,8 +24,24 @@ const (
 	MethodBitcoinUTXOs   Method = "bitcoinUTXOs"   // Retrieve bitcoin UTXOs
 )
 
+// ErrorCode is a unique identifier used to differentiate between error types.
+type ErrorCode string
```
marcopeereboom, line 28: I don't think we should use strings as the key. I feel like an int with a textual representation is a better option. In my experience string compares as error values tends to end in tears.

marcopeereboom (changes requested):

ClaytonNorthey92 (approved):

ArturDolzan (commented):
LGTM

marcopeereboom (changes requested):
We need to make error code 0 "invalid". As in "it means nothing and someone did something wrong in their code" Then error codes that mean something must be != 0.

marcopeereboom (approved):

PR #158, opened 2024-06-24
tbc: allow seeds to be overwritten

### LINKED ISSUE #157: Allow TBC seeds to be overridden
### Confirmation

### Problem

As of now, there are certain situations where we will want to override which "seeds" tbc will connect to.  The best example for use of this would be localnet, where we want to connect to a docker container via a docker network.

### Suggested solution

Allow seeds to be set in a configuration option for TBC.

Follow this logic:

if seeds are specified in config (i.e. overriden) use that list, otherwise use default seeds for configured network.

### Additional Information

_No response_

ClaytonNorthey92:
**Summary**
allow tbc seeds to be overwritten

**Changes**
if cfg seeds are set, use those, otherwise use the defaults for the network chosen

fixes #157

### DIFF
--- a/cmd/tbcd/tbcd.go
+++ b/cmd/tbcd/tbcd.go
@@ -96,7 +96,7 @@ var (
 		"TBC_SEEDS": config.Config{
 			Value:        &cfg.Seeds,
 			DefaultValue: []string{},
-			Help:         "list of tbc seeds, format of <host>:<port>",
+			Help:         "list of seed domains for Bitcoin P2P, in the format '<host>:<port>' (for localnet, must be a single host:port)",
 			Print:        config.PrintAll,
 		},
 	}
--- a/cmd/tbcd/tbcd.go
+++ b/cmd/tbcd/tbcd.go
@@ -93,6 +93,12 @@ var (
 			Help:         "address and port tbcd pprof listens on (open <address>/debug/pprof to see available profiles)",
 			Print:        config.PrintAll,
 		},
+		"TBC_SEEDS": config.Config{
+			Value:        &cfg.Seeds,
+			DefaultValue: []string{},
+			Help:         "list of tbc seeds, format of <host>:<port>",
+			Print:        config.PrintAll,
+		},
 	}
 )
 
@@ -139,6 +145,7 @@ func _main() error {
 	if err != nil {
 		return fmt.Errorf("create tbc server: %w", err)
 	}
+
 	// XXX remove, this is an illustration of calling the direct API of server
 	// go func() {
 	//	for {
--- a/config/config.go
+++ b/config/config.go
@@ -12,6 +12,7 @@ import (
 	"reflect"
 	"sort"
 	"strconv"
+	"strings"
 )
 
 type PrintMode int
@@ -88,6 +89,8 @@ func Parse(c CfgMap) error {
 				}
 
 				reflect.ValueOf(v.Value).Elem().SetBool(val)
+			case reflect.Slice:
+				*(v.Value.(*[]string)) = strings.Split(envValue, ",")
 
 			default:
 				return fmt.Errorf("unsuported type for %v: %v",
--- a/config/config.go
+++ b/config/config.go
@@ -90,7 +90,8 @@ func Parse(c CfgMap) error {
 
 				reflect.ValueOf(v.Value).Elem().SetBool(val)
 			case reflect.Slice:
-				*(v.Value.(*[]string)) = strings.Split(envValue, ",")
+				// we assume that the slice here can be appended to (ex. an empty slice)
+				reflect.AppendSlice(reflect.ValueOf(v.Value), reflect.ValueOf(strings.Split(envValue, ",")))
 
 			default:
 				return fmt.Errorf("unsuported type for %v: %v",
--- a/config/config.go
+++ b/config/config.go
@@ -91,7 +91,8 @@ func Parse(c CfgMap) error {
 				reflect.ValueOf(v.Value).Elem().SetBool(val)
 			case reflect.Slice:
 				// we assume that the slice here can be appended to (ex. an empty slice)
-				reflect.AppendSlice(reflect.ValueOf(v.Value), reflect.ValueOf(strings.Split(envValue, ",")))
+				value := reflect.ValueOf(v.Value).Elem()
+				value.Set(reflect.AppendSlice(value, reflect.ValueOf(strings.Split(envValue, ","))))
 
 			default:
 				return fmt.Errorf("unsuported type for %v: %v",
--- a/config/config.go
+++ b/config/config.go
@@ -90,7 +90,6 @@ func Parse(c CfgMap) error {
 
 				reflect.ValueOf(v.Value).Elem().SetBool(val)
 			case reflect.Slice:
-				// we assume that the slice here can be appended to (ex. an empty slice)
 				value := reflect.ValueOf(v.Value).Elem()
 				value.Set(reflect.AppendSlice(value, reflect.ValueOf(strings.Split(envValue, ","))))
 
--- a/service/tbc/tbc.go
+++ b/service/tbc/tbc.go
@@ -67,6 +67,9 @@ const (
 var (
 	zeroHash = new(chainhash.Hash) // used to check if a hash is invalid
 
+	localnetSeeds = []string{
+		"bitcoind",
+	}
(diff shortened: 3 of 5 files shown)


service/tbc/tbc.go
```diff
@@ -146,6 +182,11 @@ func sliceChainHash(ch chainhash.Hash) []byte {
 	return ch[:]
 }
 
+type Seed struct {
+	Host string
+	Port string
+}
```
joshuasing, line 188: I think instead of using a struct like this, we could keep the host and port in one string, e.g. "localhost:43111" and split it using `net.SplitHostPort`: https://pkg.go.dev/net#SplitHostPort

This would make the seeds easier to pass, and makes the configuration variable that needs to be added to `cmd/tbcd` easier to work with.

What do you think about this?

marcopeereboom: Yes.

marcopeereboom (changes requested):

service/tbc/tbc.go
ClaytonNorthey92: Yeah I was on the fence about doing that here or at the consumer (op-geth, for example) level. I can change it here 👍

joshuasing (changes requested):

service/tbc/tbc.go
```diff
@@ -306,26 +311,36 @@ func (s *Server) seed(pctx context.Context, peersWanted int) ([]tbcd.Peer, error
...
-		ips, err := resolver.LookupIP(ctx, "ip", s.seeds[k])
+	var moreSeeds []tbcd.Peer
+	for _, v := range s.seeds {
+		_, port, err := net.SplitHostPort(v)
+		if err != nil {
+			log.Errorf("SplitHostPort: %v", err)
+			errorsSeen++
+			continue
+		}
+		ips, err := resolver.LookupIP(ctx, "ip", v)
```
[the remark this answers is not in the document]
ClaytonNorthey92: hrmm I think you're right...will update

joshuasing (commented):
Will `cfg.Seeds` be exposed via an environment variable?
If so, I think you need to add an item to the slice in `cmd/tbcd/`

service/tbc/tbc.go
```diff
@@ -306,26 +311,36 @@ func (s *Server) seed(pctx context.Context, peersWanted int) ([]tbcd.Peer, error
...
 
 	errorsSeen := 0
-	var addrs []net.IP
-	for k := range s.seeds {
-		ips, err := resolver.LookupIP(ctx, "ip", s.seeds[k])
+	var moreSeeds []tbcd.Peer
+	for _, v := range s.seeds {
+		host, port, err := net.SplitHostPort(v)
+		if err != nil {
+			log.Errorf("SplitHostPort: %v", err)
```
joshuasing, line 318: I think we could make this error message a bit easier to debug:
```suggestion
			log.Errorf("Failed to parse host/port: %v", err)
```

ClaytonNorthey92:
> Will `cfg.Seeds` be exposed via an environment variable? If so, I think you need to add an item to the slice in `cmd/tbcd/`

ah yes I had forgottent we have the command (I was thinking of only `op-geth`) I'll update 👍

service/tbc/tbc.go
```diff
@@ -471,9 +486,19 @@ func (s *Server) localPeerManager(ctx context.Context) error {
...
 	peersWanted := 1
 	peerC := make(chan string, peersWanted)
-	address := net.JoinHostPort("127.0.0.1", s.port)
+
+	host, port, err := net.SplitHostPort(s.seeds[0])
+	if err != nil {
+		return err
+	}
+
+	address := net.JoinHostPort(host, port)
```
joshuasing, line 501: I think this could be simplified to use `s.seeds[0]`, unless there's a reason to split and join?
This could catch an invalid seed, however I think this code would eventually error on an invalid host/port anyway. 🤔

ClaytonNorthey92: yeah we _should_ be able to just use `s.seeds[0]` here, I'll update

cmd/tbcd/tbcd.go
```diff
@@ -93,6 +93,12 @@ var (
 			Help:         "address and port tbcd pprof listens on (open <address>/debug/pprof to see available profiles)",
 			Print:        config.PrintAll,
 		},
+		"TBC_SEEDS": config.Config{
+			Value:        &cfg.Seeds,
+			DefaultValue: []string{},
+			Help:         "list of tbc seeds, format of <host>:<port>",
```
joshuasing, line 99: I think we could make this help message more detailed by saying these are used for Bitcoin P2P, and specifying that localnet requires a single host:port instead of a list of seeds.
```suggestion
			Help:         "list of seed domains for Bitcoin P2P, in the format '<host>:<port>' (for localnet, must be a single host:port)",
```

config/config.go
```diff
@@ -88,6 +89,8 @@ func Parse(c CfgMap) error {
 				}
 
 				reflect.ValueOf(v.Value).Elem().SetBool(val)
+			case reflect.Slice:
+				*(v.Value.(*[]string)) = strings.Split(envValue, ",")
```
joshuasing, line 93: This doesn't look right, are we able to use `reflect.ValueOf` here? 🤔

ClaytonNorthey92: hey @joshuasing could you check it now? I was able to use `reflect.AppendSlice` for this, wondering what you think?

joshuasing: @ClaytonNorthey92 This appears to crash:
```
panic: reflect: call of unknown method on ptr Value

goroutine 1 [running]:
reflect.flag.mustBe(...)
        reflect/value.go:233
reflect.AppendSlice({0x1053e0be0?, 0x1057b4c38?, 0x1053a5260?}, {0x1053e2220?, 0x14000039d50?, 0x104e3f360?})
        reflect/value.go:2954 +0x20c
github.com/hemilabs/heminetwork/config.Parse(0x14000039f38?)
        github.com/hemilabs/heminetwork/config/config.go:94 +0x698
main._main()
        github.com/hemilabs/heminetwork/cmd/tbcd/tbcd.go:127 +0x28
main.main()
        github.com/hemilabs/heminetwork/cmd/tbcd/tbcd.go:194 +0x128
```

config/config.go
```diff
@@ -88,6 +89,9 @@ func Parse(c CfgMap) error {
 				}
 
 				reflect.ValueOf(v.Value).Elem().SetBool(val)
+			case reflect.Slice:
+				// we assume that the slice here can be appended to (ex. an empty slice)
+				reflect.AppendSlice(reflect.ValueOf(v.Value), reflect.ValueOf(strings.Split(envValue, ",")))
```
joshuasing, line 94: Maybe something like this would work? https://go.dev/play/p/DhbDNc4H-PL
```suggestion
				value := reflect.ValueOf(v.Value).Elem()
				value.Set(reflect.AppendSlice(value, reflect.ValueOf(strings.Split(envValue, ","))))
```

marcopeereboom (approved):

config/config.go
marcopeereboom: strings.Split works fine here too.

config/config.go
```diff
@@ -88,6 +89,10 @@ func Parse(c CfgMap) error {
 				}
 
 				reflect.ValueOf(v.Value).Elem().SetBool(val)
+			case reflect.Slice:
+				// we assume that the slice here can be appended to (ex. an empty slice)
```
joshuasing, line 93: I think this comment can be removed, as it is no longer accurate.
```suggestion
```

service/tbc/tbc.go
```diff
@@ -67,19 +63,22 @@ const (
 var (
 	zeroHash = new(chainhash.Hash) // used to check if a hash is invalid
 
+	localnetSeeds = []string{
+		"bitcoind:18444",
```
joshuasing, line 67: What do you think about defaulting this to `127.0.0.1:18444` and overriding it in the Docker environment when `bitcoind` will exist?

ClaytonNorthey92: that's a good question @joshuasing ; I was on the fence about this 🤔 do you like `127.0.0.1` better?

joshuasing: Agreement to default to `127.0.0.1` was made on a separate platform, marking as resolved.

joshuasing (approved):

PR #160, opened 2024-06-27
popm/wasm: improve js.Value creation, add JSMarshaler interface

joshuasing:
**Summary**
Improve `js.Value` creation in the WebAssembly PoP Miner, add `JSMarshaler` interface.

**Changes**
- Refactor logic of `jsValueOf` and `jsValueSafe`
- Improve error logging when a value cannot be converted
- Add `JSMarshaler` interface which can be used to implement custom logic for converting a type into a `js.Value`

### DIFF
--- a/web/popminer/api.go
+++ b/web/popminer/api.go
@@ -29,6 +29,8 @@ const (
 // ErrorCode is used to differentiate between error types.
 type ErrorCode uint32
 
+var _ JSMarshaler = (*ErrorCode)(nil)
+
 const (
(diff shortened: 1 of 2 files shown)


ClaytonNorthey92 (approved):
hey @joshuasing my feedback here is optional; I would just suggest adding some unit tests especially for `jsValueSafe`.  There seem to be a lot of conditions there, it seems like a (very) good candidate for table driven tests of what to expect.  and I feel that we should protect ourselves here.

PR #162, opened 2024-06-28
popm/wasm: add new @hemilabs/pop-miner NPM package

joshuasing:
**Summary**
Add `@hemilabs/pop-miner` package, a TypeScript API for the WebAssembly PoP Miner.

**Changes**
  - Add initial `@hemilabs/pop-miner` package (`web/packages/pop-miner`) and supporting files
  - Add GitHub Actions workflow to build and lint NPM packages
  - Update release workflow to publish NPM packages
 
**This is the initial version of this package, more functionality will be added in future pull requests.**
 **This contains more than +9K line changes, however the majority of this is from the PNPM lock files and can be ignored.**


marcopeereboom (dismissed):

marcopeereboom (approved):

ArturDolzan (approved):
We discussed a couple of things on slack, like how the front-end would provide the scriptHash to call the bitcoin balance or utxos functions, or when the user input a bitcoin private key only, how the front-end would know all the other info that is provided by generating a new key, or how the front-end will render the transactions list table, where the user will see the latests blocks and its status. 

These are topics for future PR's, so LGTM.

.github/workflows/node.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: "Setup pnpm ${{ env.PNPM_VERSION }}"
+        uses: pnpm/action-setup@v4
+        with:
+          version: "${{ env.PNPM_VERSION }}"
+
+      - name: "Setup Node ${{ env.NODE_VERSION }}"
+        uses: actions/setup-node@v4
+        with:
+          node-version: "${{ env.NODE_VERSION }}"
```
gndelia, line 50: Not a blocker, but a minor suggestion here: You could commit a  `.nvmrc` file with the node version, so all contributors of this package know which version is needed. `actions/setup-node@v4` has support for `.nvmrc` files (and if using `nvm` locally you also to automatically switch to the correct node version when using this package locally 🎉 )

web/packages/pop-miner/src/types.ts
```diff
@@ -0,0 +1,404 @@
...
+   * to the console.
+   *
+   * Options are 'trace', 'debug', 'info', 'warn', 'error' and 'critical'.
+   *
+   * @remarks
+   * The logging library used by the PoP miner allows more granular control over
+   * logging levels for individual components. For more information, please
+   * see https://github.com/juju/loggo#func-configureloggers
+   */
+  logLevel?: string | undefined;
```
gndelia, line 187: If 

> Options are 'trace', 'debug', 'info', 'warn', 'error' and 'critical'.

What about then explicitly setting those options?

```suggestion
  logLevel?: 'trace' | 'debug' | 'info' | 'warn' | 'error' | 'critical'
```

The `| undefined` could be removed considering `logLevel` is already an optional property... unless we want to allow `logLevel: undefined` (which is still correct, but makes little sense for the caller IMHO)

gndelia (approved):
LGTM - I just left a couple of minor suggestions / improvements 😄

web/packages/pop-miner/src/types.ts
joshuasing: I had considered doing this, however these aren't the only options.
As noted in the documentation above, the underlying logging library used in the PoP miner actually supports a more complex format that allows changing the logging levels for individual components: https://github.com/juju/loggo#func-configureloggers

Would it be worth trying to document this better in the comment, or perhaps improve the type instead of it just being 'string'?

I will remove the `| undefined`, thanks!

.github/workflows/node.yml
joshuasing: Ah, will do, thanks!

web/packages/pop-miner/src/types.ts
gndelia: Their specification to call that function shows it as `string`, so then `string` should be enough then, you're correct.

.github/workflows/node.yml
```diff
@@ -0,0 +1,72 @@
...
+          go-version: "${{ env.GO_VERSION }}"
+          cache: true
+          check-latest: true
+
+      - name: "Setup pnpm ${{ env.PNPM_VERSION }}"
+        uses: pnpm/action-setup@v4
+        with:
+          version: "${{ env.PNPM_VERSION }}"
+
+      - name: "Setup Node ${{ env.NODE_VERSION }}"
```
gndelia, line 45: `${{ env.NODE_VERSION }}` doesn't exist anymore!

```suggestion
      - name: "Setup Node"
```


(Same in `.github/workflows/release.yml`)

gndelia: Again, not a blocker as actions still work, and an empty space is logged

image


So feel free to merge either way if you prefer it that way 😄

joshuasing: Good catch! Thanks!

gndelia (approved):

ClaytonNorthey92 (approved):

Issue #170, opened 2024-07-09
PoP Miner Stats Daemon for Absinthe Reporting


max-sanchez:
### Confirmation

- [X] I have checked for similar issues.

### Problem

We need a way to inform Absinthe of PoP miners.

### Suggested solution

Create a simple daemon we can run which scans our testnet for PoP payouts and reports them to Absinthe.

Suggested design:
Daemon that can run 24/7 and is configured with a starting L2 height and a delay (in L2 blocks) which it scans up to.
Upon initial startup, daemon scans from the starting L2 height up to the tip minus the delay. As additional L2 blocks come in, the daemon progresses scanning new L2 blocks, always up to tip - delay.
As the Daemon progresses, it keeps track of a state machine which maps each EVM address to the L2 block height(s) in which it received a PoP payout.
For each block it scans, it checks if a PoP Payout transaction exists. If it does, it extracts the payout EVM addresses from the transaction's data used to call the GovernanceToken contract, and uses that information to update the state machine it tracks.

TODO: Decide how to report this collected data to Absinthe - they've said they're "flexible" but I'm getting more details on exactly how this would look - probably a POST request to their endpoint, but need to decide whether we're reporting state changes, and how that gets processed on Absinthe's side for our points setup.

gabmontes:
@max-sanchez I have the deamon core functionality ready. Please provide the following parameters:

- Start L2 block
- Delay

In addition, please provide more information on how to report the data to Absinthe or contact me with whoever I can talk to in order to obtain that definition.

gabmontes:
Information sent by Absinthe yesterday:

GraphQL playground: https://graphiql.absinthe.network/
Auth: authorization header ({"Authorization": "Bearer <API KEY>"})
API key: https://send.bitwarden.com/#O3VTDvmjUkW3aLHJAX1gCA/Fc-3nDqWW7WkDO5n_NUwAA

gabmontes:
The GraphQL schema of the API is:

```json
{
    "data": {
        "__schema": {
            "queryType": {
                "fields": [
                    {
                        "name": "api_points"
                    },
                    {
                        "name": "api_points_by_pk"
                    }
                ]
            }
        }
    }
}
```

jcvernaleo:
@gabmontes this one is done, isn't it?

gabmontes:
@jcvernaleo the cron job was created, tested and ready for deployment. That never happened, as far as I know.

Is this still needed? cc / @max-sanchez

[...]

PR #172, opened 2024-07-09
PoP stats service

gabmontes:
**Summary**

This PR adds a service that will collect all the PoP payout transactions, summarize and report to Absinthe.

See [the README file](https://github.com/hemilabs/heminetwork/pull/172/files#diff-694383e8d41fa714efb23ef65c0da601a1b950216da7bb37b2f9680a3251d2c1) for details.

Related to #170.

**Changes**

Add everything to the `/popstats` folder. No additional changes to the repo.


popstats/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2024 Hemi Labs, Inc.
+# Use of this source code is governed by the MIT License,
+# which can be found in the LICENSE file.
+
+FROM node:22-alpine
```
joshuasing, line 5: I think we should pin this image to a SHA256 hash for security reasons.

popstats/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
```
joshuasing, line 6: All other images in this repository have the following labels, what are your thoughts of adding them to this image?
```suggestion

# Build metadata
ARG VERSION
ARG VCS_REF
ARG BUILD_DATE
LABEL org.opencontainers.image.created=$BUILD_DATE \
      org.opencontainers.image.authors="Hemi Labs" \
      org.opencontainers.image.url="https://github.com/hemilabs/heminetwork" \
      org.opencontainers.image.source="https://github.com/hemilabs/heminetwork" \
      org.opencontainers.image.version=$VERSION \
      org.opencontainers.image.revision=$VCS_REF \
      org.opencontainers.image.vendor="Hemi Labs" \
      org.opencontainers.image.licenses="MIT" \
      org.opencontainers.image.title="PoP stats" \
      org.label-schema.build-date=$BUILD_DATE \
      org.label-schema.name="PoP stats" \
      org.label-schema.url="https://github.com/hemilabs/heminetwork" \
      org.label-schema.vcs-url="https://github.com/hemilabs/heminetwork" \
      org.label-schema.vcs-ref=$VCS_REF \
      org.label-schema.vendor="Hemi Labs" \
      org.label-schema.version=$VERSION \
      org.label-schema.schema-version="1.0"

```

popstats/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+Set the following environment variables in a `.env` file:
+
+- ABSINTHE_API_URL and ABSINTHE_API_KEY: URL of the Absinthe GraphQL API and authentication JWT
+- REDIS_URL: Full URI of the Redis database
+
+Optionally, start a Redis instance:
+
+```sh
+npm run redis:start
```
joshuasing, line 31: PNPM is being used, however this command appears to use NPM.
```suggestion
pnpm redis:start
```

popstats/README.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+Finally, build the Docker image and start a container:
+
+```sh
+npm run docker:build
+npm run docker:run
```
joshuasing, line 38: PNPM is being used, however this command appears to use NPM.
```suggestion
pnpm docker:build
pnpm docker:run
```

popstats/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "hemi-pop-stats",
+  "version": "1.0.0",
+  "description": "Report PoP payouts to Absinthe",
+  "license": "MIT",
+  "author": {
+    "name": "Gabriel Montes"
```
joshuasing, line 7: Other packages in this repository use the author "Hemi Labs". I think we should do the same here for consistency.
```suggestion
    "name": "Hemi Labs"
```

joshuasing (changes requested):
Looks good. I have left a few comments.

There is also a PNPM workspace already setup under the `web/` directory, with ESLint and Prettier already setup. What do you think of moving this to `web/packages/pop-stats/`? I think this would help keep everything tidier and more consistent.

popstats/README.md
gabmontes: I removed all the scripts. Seem cleaner now.

popstats/package.json
gabmontes: I don't see a problem with it and prefer proper attribution. Unless it is a policy, let's keep it as is.

popstats/README.md
gabmontes: Same.

gabmontes:
@joshuasing the `/web` folder is focused on the web PoP miner, UI and WASM code. This is a different process I think it should belong to its own folder.

What we should do is to update its configuration to match the standards in regards to linting and formatting we use in all the other JavaScript repos at Bloq and Hemi.

popstats/README.md
```diff
@@ -0,0 +1,47 @@
+# PoP stats
+
+This PR adds a service that will identify all the PoP payout transactions, summarize, calculate the incentive program points and report those to Absinthe.
```
gndelia, line 3: It's a minor thing, but, weirdly, the README starts with "This PR" - once merged it will no longer make sense... Were you thinking of a changelog or something?

Having said that, some minor corrections:

```suggestion
This PR adds a service that identifies all the PoP payout transactions, summarizes, calculates the incentive program points, and reports those to Absinthe.
```

popstats/README.md
```diff
@@ -0,0 +1,47 @@
...
+In order to walk through the history more efficiently, the service processes the blocks in chunks of 7200, which is roughly 1 day worth of Hemi Network blocks.
+Then it reports the results before taking care of the next chunk.
+The intended side-effect is to report points once per day, even during the initial run where many days need to be reported.
+
+In addition, within each chunk, the network is queried for those transfer events in sub-chunks of 1200 blocks (4 hours) to minimize the impact on the RPC nodes. Additional rate limits were set with the same goal.
+
+The state of the process is stored in a single key in a Redis database so the service knows where to start at the beginning of each run.
+That state is the number of the last blocks processed and reported to Absinthe.
+
+The data sent to Absinthe includes the address receiving the PoP payouts, the amount of points granted and some metadata just intended for tracking/auditing purposes.
```
gndelia, line 19: Minor spelling checks 

```suggestion
This service is prepared to be run once a day as a cron job.
On each run, it gets all the HEMI token mint operations (transfers coming from the null address), filters those originated by the depositor address (0x8888...8888), summarizes the events by recipient address, does some data formatting and sends it to Absinthe.

In order to walk through the history more efficiently, the service processes the blocks in chunks of 7200, which is roughly 1 day's worth of Hemi Network blocks.
Then it reports the results before taking care of the next chunk.
The intended side-effect is to report points once per day, even during the initial run where many days need to be reported.

In addition, within each chunk, the network is queried for those transfer events in sub-chunks of 1200 blocks (4 hours) to minimize the impact on the RPC nodes. Additional rate limits were set with the same goal.

The state of the process is stored in a single key in a Redis database so the service knows where to start at the beginning of each run.
That state is the number of the last blocks processed and reported to Absinthe.

The data sent to Absinthe includes the address receiving the PoP payouts, the amount of points granted and some metadata just intended for tracking/auditing purposes.
```

popstats/server.js
```diff
@@ -0,0 +1,268 @@
...
+    fromBlock,
+    toBlock: maxToBlock,
+  }).then((moreTransfers) => ({
+    fromBlock: maxToBlock + 1n,
+    toBlock,
+    transfers: transfers.concat(moreTransfers),
+  }));
+}
+
+const thereAreBlocksToProcess = ({ fromBlock, toBlock }) =>
```
gndelia, line 183: I think this reads better as a question, rather than an affirmative question 🤔  

```suggestion
const areThereBlocksToProcess = ({ fromBlock, toBlock }) =>
```

Thoughts?

gndelia: ok, now I see how it would read better:

"do whilst there are blocks to process"

nvm then!

gndelia (approved):

.github/CODEOWNERS
```diff
@@ -35,3 +35,6 @@
 
 # Release script
 /scripts/release.sh @hemilabs/dev @joshuasing
+
+# PoP payouts reporting service
+/popstats/ @gabmontes
```
joshuasing, line 40: As this repository is maintained by core-eng, I think requesting a review from core-eng would be a good idea for this directory as well:
```suggestion
/popstats/ @hemilabs/dev @gabmontes
```

popstats/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+WORKDIR /home/node/app
+
+COPY package.json pnpm-lock.yaml .
+RUN corepack enable
+RUN pnpm install --prod
+COPY . .
+
+USER node
+
+CMD [ "npm", "start"]
```
joshuasing, line 38: It seems `pnpm` is used to install dependencies, however `npm` is used to run. What do you think of using `pnpm` for both to improve consistency?
```suggestion
RUN pnpm install --prod
COPY . .

USER node

CMD [ "pnpm", "start"]
```

popstats/README.md
joshuasing, line 1: I think we could wrap the contents of this file at 120 characters to match the other markdown files. I think this makes it easier to maintain on smaller screens, as well as in editors like vim or emacs.

popstats/package.json
joshuasing, line 1: Will this package ever be published to NPM? If not, could we add `"private": true` to prevent accidental publishing?

popstats/.nvmrc
```diff
@@ -0,0 +1 @@
+22
```
joshuasing, line 1: I think we should add a newline to the end of this file.

joshuasing (changes requested):
Looks good, a few things I missed on my first pass. Will any GitHub Actions be added for this service?

The `node.yml` file already builds the packages in `web/`, a separate job could be added there that builds this module.

I assume the Dockerfile will also need to be built for releases, would you like me to make a pull request to add building this to the `release.yml` action once this has been merged?

.github/CODEOWNERS
gabmontes: Well, it looks like that defeats the purpose of the CODEOWNERS file. If `@hemilabs/dev` is owner of everything and you all are part of that team, everyone is owner of everything so no one has true ownership.

If `@hemilabs/dev` is added to `/popstats/` as suggested, anyone in that team can approve and merge changes without me having any say. Therefore, you should instead revisit how the rest of the `CODEOWNERS` file is defined.

gabmontes:
@joshuasing GitHub Actions to automatically build the Docker image could be added later. I will define that with @dhidalgX and send a PR or ask you for help. Thanks for pointing that out!

.github/CODEOWNERS
joshuasing: You have a good point. We use code owners are more of a way to request reviews, and it isn't currently required through repo rulesets.

I thought this was additive, not one or the other, and it seemed to be this way for other entries (requesting a review from hemilabs/dev and someone else).
I think having both means that we can request a review from core-eng as well as you, meaning you will still need to review changes but everything gets a secondary review from core.

What are your thoughts?

gabmontes: Even in that case the usage pattern is weird. Let's say we have a team and a user as code owners, then a PR is created and both the team and the user are requested reviews. Now, if either of those (users within the team) or the individual user approves the PR, it goes through.

But as code owner reviews are not required by the repo config, then the CODEOWNERS config only helps with requesting reviews, none of which are mandatory.

Even when I prefer true ownership, I'm ok with the current setup if that works better for you guys.

cc/ @jcvernaleo @max-sanchez

joshuasing (approved):

PR #173, opened 2024-07-10
popm: improve UTXO selection when creating Bitcoin transaction

joshuasing:
**Summary**
Randomly select an available UTXO after filtering the available UTXOs that have a value equal to or greater than the required value for the transaction.

Previously, the selection was not random, and the first UTXO with a value equal to or greater than the required value would be used.

**Changes**
- Rewrite `pickUTXOs` function to return a randomly selected UTXO after filtering the available UTXOs that have a value greater than or equal to the required value needed for the transaction.
- Rename `pickUTXOs` to `pickUTXO` and return `*bfgapi.BitcoinUTXO` instead of `[]*bfgapi.BitcoinUTXO`, as only one UTXO is ever returned
- Add more debug logs for debugging UTXO-related issues

### DIFF
--- a/service/popm/popm.go
+++ b/service/popm/popm.go
@@ -8,10 +8,12 @@ import (
 	"bytes"
 	"cmp"
 	"context"
+	"crypto/rand"
 	"crypto/sha256"
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"math/big"
 	"slices"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


service/popm/popm.go
```diff
@@ -242,23 +244,42 @@ func (m *Miner) bitcoinUTXOs(ctx context.Context, scriptHash []byte) ([]*bfgapi.
...
+		num, len(utxos), amount)
+
+	if num == 0 {
+		// There are no available UTXOs with a value >= amount.
+		return nil, errors.New("insufficient funds to PoP mine, " +
+			"please send additional funds to continue mining")
+	}
+
+	utxo := ux[0]
+	if num > 1 {
```
ClaytonNorthey92, line 270: nit: I don't think we need this check, and I think we can just get a utxo by `ux[r % num]`

ClaytonNorthey92 (approved):
minor nitpick; nothing blocking

marcopeereboom (approved):

max-sanchez (approved):

PR #174, opened 2024-07-10
popm/wasm: fix ethereum address when generating and parsing keys

joshuasing:
**Summary**
Use uncompressed serialised secp256k1 public key to create the Ethereum address.
Previously, the compressed serialised secp256k1 public key was used, which resulted in a different value being provided as the address.

**Changes**
- Use uncompressed serialised secp256k1 public key to create Ethereum addresses in the WebAssembly PoP Miner.


jcvernaleo (approved):
OK

ClaytonNorthey92 (approved):

PR #175, opened 2024-07-10
popm/wasm: add events and clean up globals

joshuasing:
**Summary**
Add events to the PoP Miner and WebAssembly PoP Miner.

**Changes**
- Add events to `popm` and `Miner.RegisterEventHandler(EventHandler)`
- Add events to the WebAssembly PoP Miner and `@hemilabs/pop-miner` package
- Rename `activeMiner()` to `runningMiner()`
- Add a `Service` type and use `svc` global variable (replaces `pm` and `pmMtx`). This allows us to have a global struct to store data in, without the need to use a lock to access the data. The miner is now stored in a `miner` field, guarded by the `minerMtx` RWMutex. This service also contains the event listeners and the listeners map RWMutex.
- Add initial 4 events: `minerStart`, `minerStop`, `mineKeystone`, and `transactionBroadcast`.

### DIFF
--- a/web/packages/pop-miner/src/types.ts
+++ b/web/packages/pop-miner/src/types.ts
@@ -310,7 +310,10 @@ export interface EventListener {
  * @param eventType The event type to listen for. If '' then all events.
  * @param listener The event listener that will be called when the event is dispatched.
  */
-export declare function addEventListener(eventType: EventType | '', listener: EventListener): Promise<void>;
+export declare function addEventListener(
+  eventType: EventType | '',
+  listener: EventListener,
+): Promise<void>;
 
 /**
  * Unregisters an event listener.
@@ -318,7 +321,10 @@ export declare function addEventListener(eventType: EventType | '', listener: Ev
  * @param eventType The event type to stop listening for.
  * @param listener The event listener to unregister.
  */
-export declare function removeEventListener(eventType: EventType | '', listener: EventListener): Promise<void>;
+export declare function removeEventListener(
+  eventType: EventType | '',
+  listener: EventListener,
+): Promise<void>;
 
 /**
  * @see startPoPMiner
--- a/web/packages/pop-miner/src/types.ts
+++ b/web/packages/pop-miner/src/types.ts
@@ -233,6 +233,94 @@ export declare function bitcoinAddressToScriptHash(
 ): Promise<BitcoinAddressToScriptHashResult>;
 
 /**
+ * Represents a type of event.
+ */
+export type EventType =
+  | 'minerStart'
+  | 'minerStop'
+  | 'mineKeystone'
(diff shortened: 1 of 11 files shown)


ArturDolzan (approved):

web/popminer/api.go
```diff
@@ -206,3 +214,65 @@ type BitcoinUTXO struct {
...
+// These events will be forwarded to JavaScript, however we also dispatch events
+// that are specific to the WebAssembly PoP Miner.
+var popmEvents = map[popm.EventType]EventType{
+	popm.EventTypeMineKeystone:         EventTypeMineKeystone,
+	popm.EventTypeTransactionBroadcast: EventTypeTransactionBroadcast,
+}
+
+// eventTypes is a map used to parse string event types.
+var eventTypes = map[string]EventType{
+	"":                                     "", // Listen for all events.
```
marcopeereboom, line 247: maybe use * ?

marcopeereboom (approved):

web/popminer/api.go
joshuasing: I think that would be better, will update.

web/packages/pop-miner/src/types.ts
```diff
@@ -232,6 +232,100 @@ export declare function bitcoinAddressToScriptHash(
...
+  /**
+   * The error that caused the PoP miner to exit, or null.
+   */
+  error?: Error;
+};
+
+/**
+ * Dispatched when the PoP miner begins mining a keystone.
+ */
+export type EventMineKeystone = Event & {
```
ClaytonNorthey92, line 273: Should we call this `EventMineKeystoneStart` based on the description?

web/packages/pop-miner/src/types.ts
```diff
@@ -232,6 +232,100 @@ export declare function bitcoinAddressToScriptHash(
...
+   * The keystone to be mined.
+   */
+  keystone: L2Keystone;
+};
+
+/**
+ * Dispatched when the PoP miner broadcasts a PoP transaction to the Bitcoin
+ * network.
+ */
+export type EventTransactionBroadcast = Event & {
```
ClaytonNorthey92, line 286: Should we call this `EventTransactionBroadcasted`?

ClaytonNorthey92 (approved):

web/packages/pop-miner/src/types.ts
joshuasing: I think `EventMineKeystone` makes more sense, as it is dispatched with `mineKeystone` is called.
If later we add an event for when a keystone has finished being mined, I think that event would be called `EventMinedKeystone`

gndelia (approved):

PR #185, opened 2024-07-16
remove lastErr variables from rpc_test.go

### LINKED ISSUE #182: tbc: fix unused 'lastErr' in RPC tests
### Confirmation

### Describe the bug

Running `make staticcheck` results in `lastErr` variables being reported as unused:
```
service/tbc/rpc_test.go:72:4: this value of lastErr is never used (SA4006)
service/tbc/rpc_test.go:79:4: this value of lastErr is never used (SA4006)
service/tbc/rpc_test.go:89:4: this value of lastErr is never used (SA4006)
service/tbc/rpc_test.go:89:14: Errorf doesn't have side effects and its return value is ignored (SA4017)
(more omitted)
```

### Expected behaviour

Correct retry logic should be implemented in the TBC RPC tests.

### Environment

_No response_

### Additional Information

_No response_

ClaytonNorthey92:
**Summary**
lastErr was added to handle for something that likely shouldn't have been; it's an un-needed way to handle for errors.  replace with simply failing the test if there are errors.

**Changes**
remove lastErr variables from rpc_test.go

fixes #182

### DIFF
--- a/service/tbc/rpc_test.go
+++ b/service/tbc/rpc_test.go
@@ -542,8 +542,9 @@ func TestBalanceByAddress(t *testing.T) {
 						// there is a chance we just haven't finished indexing
 						// the blocks and txs, retry until timeout
 						continue
+					} else {
+						break
 					}
-					break
 				} else {
 					t.Fatalf("received unexpected command: %s", v.Header.Command)
 				}
--- a/service/tbc/rpc_test.go
+++ b/service/tbc/rpc_test.go
@@ -446,7 +446,7 @@ func TestBalanceByAddress(t *testing.T) {
 
 	for _, tti := range testTable {
 		t.Run(tti.name, func(t *testing.T) {
-			ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+			ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
 			defer cancel()
 
 			initialBlocks := 0
--- a/service/tbc/rpc_test.go
+++ b/service/tbc/rpc_test.go
@@ -62,14 +62,14 @@ func TestBlockHeadersByHeightRaw(t *testing.T) {
 		t.Fatal(ctx.Err())
 	}
 
-	if err = tbcapi.Write(ctx, tws.conn, "someid", tbcapi.BlockHeadersByHeightRawRequest{
+	if err := tbcapi.Write(ctx, tws.conn, "someid", tbcapi.BlockHeadersByHeightRawRequest{
 		Height: 55,
 	}); err != nil {
 		t.Fatal(err)
 	}
 
 	var v protocol.Message
-	if err = wsjson.Read(ctx, c, &v); err != nil {
+	if err := wsjson.Read(ctx, c, &v); err != nil {
 		t.Fatal(err)
 	}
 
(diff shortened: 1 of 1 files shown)


service/tbc/rpc_test.go
```diff
@@ -56,42 +55,31 @@ func TestBlockHeadersByHeightRaw(t *testing.T) {
...
-			if err := json.Unmarshal(v.Payload, &response); err != nil {
-				t.Fatal(err)
-			}
-			break
-		} else {
-			lastErr = fmt.Errorf("received unexpected command: %s", v.Header.Command)
-		}
+	var v protocol.Message
+	err = wsjson.Read(ctx, c, &v)
+	if err != nil {
```
joshuasing, line 73: I think this err assignment could be nested (and in other functions below):
```suggestion
	if err = wsjson.Read(ctx, c, &v); err != nil {
```

service/tbc/rpc_test.go
```diff
@@ -56,42 +55,31 @@ func TestBlockHeadersByHeightRaw(t *testing.T) {
...
 
-	if lastErr != nil {
-		t.Fatal(lastErr)
+	if v.Header.Command == tbcapi.CmdBlockHeadersByHeightRawResponse {
+		if err := json.Unmarshal(v.Payload, &response); err != nil {
+			t.Fatal(err)
+		}
+	} else {
+		t.Fatalf("received unexpected command: %s", v.Header.Command)
 	}
```
joshuasing, line 83: I think this logic can be inverted to fail early. (and in other functions below)
```suggestion
	if v.Header.Command != tbcapi.CmdBlockHeadersByHeightRawResponse {
		t.Fatalf("received unexpected command: %s", v.Header.Command)
	}

	if err := json.Unmarshal(v.Payload, &response); err != nil {
		t.Fatal(err)
	}
```

joshuasing (changes requested):

service/tbc/rpc_test.go
```diff
@@ -558,57 +492,49 @@ func TestBalanceByAddress(t *testing.T) {
...
+					t.Logf("unexpected diff: %s", diff)
 
-			if lastErr != nil {
-				t.Fatal(lastErr)
+					// there is a chance we just haven't finished indexing
+					// the blocks and txs, retry until timeout
+					continue
+				} else {
+					break
+				}
```
joshuasing, line 537: I think this `else` statement can be removed to simplify this logic.
```suggestion
				}

				break
```

joshuasing (approved):

service/tbc/rpc_test.go
```diff
@@ -1692,11 +1510,11 @@ func indexAll(ctx context.Context, t *testing.T, tbcServer *Server) {
 
 	hash := bh.BlockHash()
 
-	if err := tbcServer.TxIndexer(ctx, &hash); err != nil {
+	if err = tbcServer.TxIndexer(ctx, &hash); err != nil {
 		t.Fatal(err)
 	}
 
-	if err := tbcServer.UtxoIndexer(ctx, &hash); err != nil {
+	if err = tbcServer.UtxoIndexer(ctx, &hash); err != nil {
```
marcopeereboom, line 1517: When using this idiom you really should keep `:=`.

marcopeereboom (changes requested):

service/tbc/rpc_test.go
joshuasing: I think in most cases here, it is fine to set the value of `err` instead of shadowing it.

ClaytonNorthey92: ah yes will fix

max-sanchez (approved):

PR #186, opened 2024-07-17
popm: exclude tx outputs that would be dust

### LINKED ISSUE #183: Ensure that we don't generate dust
### Confirmation

### Describe the bug

As of now, when we create a txout for pop mining, there is a chance that the "change" that we create is below the "dust threshold" in bitcoind.

When this occurs we get an error of "dust".

The [change amount ](https://github.com/hemilabs/heminetwork/blob/1f168f225ba36bf7eed30f45eebd5227447c839b/service/popm/popm.go#L304) cannot be below the dust threshold in bitcoind, which I fairly certain is defined [here](https://github.com/bitcoin/bitcoin/blob/262260ce1e919613ba60194a5861b0b0a51dfe01/src/policy/policy.cpp#L26) 

### Expected behaviour

Avoid generating dust by considering these txouts too low.

### Environment

Version: 
Operating System: 
Architecture: 

### Additional Information

_No response_

joshuasing:
**Summary**
Avoid "dust" errors when broadcasting PoP transactions to Bitcoin by checking if the transaction output would be considered dust with a relay tx fee of 10,000.

If the transaction output would be considered dust, then don't include the output and instead leave the remaining to be included as a fee.

Fixes #183 

**Changes**
 - Don't add a change transaction output when creating a transaction if the transaction output is dust.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -35,6 +35,7 @@ require (
 	github.com/Azure/go-ansiterm v0.0.0-20230124172434-306776ec8161 // indirect
 	github.com/Microsoft/go-winio v0.6.1 // indirect
 	github.com/Microsoft/hcsshim v0.11.4 // indirect
+	github.com/aead/siphash v1.0.1 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f // indirect
 	github.com/cenkalti/backoff/v4 v4.2.1 // indirect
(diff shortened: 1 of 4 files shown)


ClaytonNorthey92:
hey @joshuasing (and maybe @marcopeereboom or @max-sanchez could answer this) why did we choose 10,000 for the `minRelayTxFee`?  [The bitcoind default looks to be 1000.](https://github.com/bitcoin/bitcoin/blob/3679fa167f7d198d9287f05f210f03e83b76f3b8/src/policy/policy.h#L57).  Are these values different? Or are we handling for something outside of the default?

ClaytonNorthey92 (approved):

marcopeereboom (approved):

PR #188, opened 2024-07-26
Add a header and block cache.

marcopeereboom:
**Summary**
Unwinding a chain is quite expensive when measured in block and header lookups. Often repeating the same process. This creates a cache for both. Note that the header cache is only primed during reads; this is deliberate because the cost of generating the cache entries is too high in the write path.

It tries to keep the block cache at 1GB and the header cache at 100MB.

**Changes**

### DIFF
--- a/cmd/hemictl/hemictl.go
+++ b/cmd/hemictl/hemictl.go
@@ -307,7 +307,7 @@ func tbcdb() error {
 
 		levelDBHome := "~/.tbcd" // XXX
 		network := "testnet3"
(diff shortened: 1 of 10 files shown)


joshuasing (approved):

max-sanchez (approved):

joshuasing (approved):
Looks good to me. I have a semi-working diff to move the other functions with TODO comments to the peer manager, however I think that should be a separate PR.

joshuasing (approved):
Looks good! Peer manager TODO comments will be resolved in a separate pull request.

Issue #193, opened 2024-08-02
bfgapi: add pagination to `pop-txs-for-l2-block`


joshuasing:
### Confirmation

- [X] I have checked for similar issues.

### Problem

The `pop-txs-for-l2-block` response becomes large when there is a lot of active PoP Miners broadcasting transactions.

### Suggested solution

I think we should add pagination to the `pop-txs-for-l2-block` request/response, so that BSS can request the all of the transactions in several requests. This should make this a lot more table, while reducing the overall size of each response, preventing us from exceeding the read limit.

PR #194, opened 2024-08-02
bss: increase read limit for bfg conn to 2 MiB

joshuasing:
**Summary**
Increase the read limit in BSS for the BFG connection to 2 MiB (previously 512 KiB).

When there is a lot of active PoP Miners broadcasting transactions, the `pop-txs-for-l2-block` response from BFG becomes too large for BSS to read. This increases the read limit to temporarily avoid this problem.

Related to: https://github.com/hemilabs/heminetwork/issues/193

**Changes**
 - Add `protocol.ConnOptions` which can be used to provide options when creating a `protocol.Conn`.
 - Increase the BSS -> BFG connection read limit to 2 MiB (previously 512 KiB).

### DIFF
--- a/api/auth/secp256k1_test.go
+++ b/api/auth/secp256k1_test.go
@@ -151,7 +151,9 @@ func TestProtocolHandshake(t *testing.T) {
 	}
 
 	clientURI := testServer.URL
-	conn, err := protocol.NewConn(clientURI, clientAuth)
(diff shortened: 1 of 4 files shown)


ClaytonNorthey92 (approved):

jcvernaleo (approved):
OK

PR #199, opened 2024-08-09
bfg: add more metrics

joshuasing:
**Summary**
Add more metrics to the BFG service:
1. `bfg_pop_broadcasts_total` (counter): Total number of PoP transaction broadcasts
2. `bfg_rpc_calls_total{listener,command}` (counter): Total number of successful RPC commands (note: these may not succeed, but the connection will still be open after the call is processed)
3. `bfg_rpc_calls_duration_seconds{listener,command}` (histogram): RPC call durations in seconds
4. `bfg_rpc_connections` (gauge): Number of active RPC WebSocket connections

**Changes**
- Add `metrics` struct that contains all Prometheus collectors
- Add new collectors, and add metric collection to functions
- Change `as` to `AS` in Docker files (per warning in Docker).

All metrics have been tested using localnet (+ a patch to expose prometheus handler)

### DIFF
--- a/docker/bfgd/Dockerfile
+++ b/docker/bfgd/Dockerfile
@@ -3,7 +3,7 @@
 # which can be found in the LICENSE file.
 
 # Build stage
-FROM golang:1.22.2-alpine@sha256:cdc86d9f363e8786845bea2040312b4efa321b828acdeb26f393faa864d887b0 as builder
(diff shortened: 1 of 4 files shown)


jcvernaleo (approved):
Looks like a useful set of changes.

OK

marcopeereboom (approved):

dhidalgX (approved):
LGTM

PR #201, opened 2024-08-13
Add a rawdb package for blocks

### LINKED ISSUE #155: Use flat files for blocks instead of leveldb
### Confirmation

### Problem

It's dawned on me that directly storing blocks in leveldb may slow down IBD a bit.

### Suggested solution

It's worth doing a quick experiment where we store blocks in files direct and use leveldb to index them. The size of the files can be maxed out at something manageable for backups and copies (determine something in the range of 100M to <1G). Leveldb then would simply record the hash of the block as its key and the index into the block as its value. Filenames have a numerical template of a uint32 (e.g. 00000000.blk) thus block xyz store in file 13 would be stored in leveldb as `[xyz] = {filename: uint32(13), index: uint32(123456)}`.

Before moving ahead with such a change performance must be measured to see if it is worth the trouble.

A nice side-effect of this design is that we can copy blocks around and recreate the index relatively easy.

### Additional Information

_No response_

marcopeereboom:
**Summary**
Add rawdb and rig it up inside leveldb

**Changes**
fixes #155


rawdb/rawdb.go
```diff
@@ -0,0 +1,225 @@
...
+
+	index *leveldb.DB
+}
+
+func New(home string, maxSize int64) (*RawDB, error) {
+	log.Tracef("New")
+	defer log.Tracef("New exit")
+
+	if maxSize < 4096 {
+		return nil, fmt.Errorf("invalid max size: %v", maxSize)
```
ClaytonNorthey92, line 50: nit: could we update this error message to something like:
_invalid max size, must be at least 4096, received %v_

database/level/level.go
```diff
@@ -176,6 +212,12 @@ func New(ctx context.Context, home string, version int) (*Database, error) {
 		return nil, fmt.Errorf("leveldb %v: %w", TransactionsDB, err)
 	}
 
+	// Blocks database is special
+	err = l.openRawDB(BlocksDB, 256*1024*1024)
```
joshuasing, line 215: Could `256 * 1024 * 1024` be extracted into a constant, so that it is obvious what it is?

rawdb/rawdb.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
+
+	tries := 0
+	for {
+		// This should not happen but we must ensure we aren't spinning.
+		if tries > 1 {
+			return fmt.Errorf("could not determinet last filename")
```
ClaytonNorthey92, line 124: nit: spelling mistake "determinet" > "determine"

ClaytonNorthey92 (approved):

rawdb/rawdb.go
joshuasing: You beat me, haha. Resolved in: [`e39f694` (#201)](https://github.com/hemilabs/heminetwork/pull/201/commits/e39f694d5045e5c9a143b4d387daad31c1007619)

rawdb/rawdb.go
```diff
@@ -0,0 +1,220 @@
...
+		} else if fi.Size()+int64(len(value)) > r.maxSize {
+			last++
+			lastData := make([]byte, 8)
+			binary.BigEndian.PutUint32(lastData, last)
+			err = r.index.Put(lastFilenameKey, lastData, nil)
+			if err != nil {
+				return err
+			}
+			tries++
+			continue
```
joshuasing, line 156: I think this `continue` can be removed, as there is nothing below this if/else-if/else.

rawdb/rawdb.go
```diff
@@ -0,0 +1,220 @@
...
+	f, err := os.OpenFile(filename, os.O_RDONLY, 0o600)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		err := f.Close()
+		if err != nil {
+			log.Errorf("close: %v", err)
+		}
+	}()
```
joshuasing, line 209: nit: I think adding a blank line after this line would make this slightly easier to read.
```suggestion
	}()

```

rawdb/rawdb_test.go
```diff
@@ -0,0 +1,101 @@
...
+	remove := true
+	defer func() {
+		if remove {
+			err := os.RemoveAll(home)
+			if err != nil {
+				panic(err)
+			}
+		} else {
+			t.Logf("did not remove home: %v", home)
+		}
```
joshuasing, line 27:
```suggestion
		if remove {
			err := os.RemoveAll(home)
			if err != nil {
				panic(err)
			}
			return
		}
		t.Logf("did not remove home: %v", home)
```

rawdb/rawdb.go
marcopeereboom: yes and no. It's there to explicitly mark the code block as terminal. If someone adds code after the last else clause it won't break.

joshuasing (approved):

marcopeereboom:
I decided to let the leveldb metadata table govern the version information inclusive for rawdb. I am going ahead and merge this and I'll add the blocksize write to metadata in a follow up PR.

PR #202, opened 2024-08-14
all: add user agents and improve build info log

joshuasing:
**Summary**
Add `User-Agent` headers and a `UserAgent` to the secp256k1 authenticator.
This allows our HTTP connections to be identifiable, and provides some basic information about the client.

**Changes**
 - Add user agent creation to the `version` package.
 - Add `User-Agent` header to the popm->bfg connection.
 - Add `UserAgent` to the secp256k1 authenticator.
 - Properly set the `version.PreRelease` variable at link-time for releases.
 - Add ability to set Go linker flags through the Makefile (GO_LDFLAGS).
 - Add `BuildInfo()` function to `version` package that includes basic information about the environment that is useful for debugging.


RamdanSe55:
should have added more storage for people to fund their addresses so the testnet can roun perfectly

crlt-max:
LFG HEMI

version/useragent_test.go
```diff
@@ -0,0 +1,65 @@
...
+		},
+		{
+			product:  "popmd",
+			version:  "0.1.0",
+			comments: []string{"Hemi Labs", "linux/amd64", "+https://github.com/hemilabs/heminetwork"},
+			want:     "popmd/0.1.0 (Hemi Labs; linux/amd64; +https://github.com/hemilabs/heminetwork)",
+		},
+	}
+	for _, tt := range tests {
+		t.Run("", func(t *testing.T) {
```
ClaytonNorthey92, line 58: could we add a name here? like add a `name` field to the test config then
```go
t.Run(tt.name, ...)
```

ClaytonNorthey92 (changes requested):
hey @joshuasing I have a very small request change to a test; it is _almost_ a nitpick but I think it would help determine which test fails if one does.

additionally, and this is completely optional but would give me peace of mind 😌 , I would love a test that sends a specific `UserAgent` to bfg, and ensured that bfg received the useragent correctly (i.e. there is no issue reading that value from where it should be)

version/useragent_test.go
joshuasing: Yep, I can do that. Go does use #%02d for these by default, which lets you tell which test is which but this would probably be better.

ClaytonNorthey92 (approved):

cmd/bfgd/bfgd.go
```diff
@@ -152,6 +152,11 @@ func _main() error {
 	return nil
 }
 
+func init() {
+	version.Component = "bfgd"
+	welcome = "Hemi Bitcoin Finality Governor " + version.BuildInfo()
```
marcopeereboom, line 157: Why isn't this done in global var () ?

This is true for the other ones below this as well.

marcopeereboom (changes requested):

marcopeereboom (approved):

Issue #195, opened 2024-08-02
Paginate PoP Payouts Response


ClaytonNorthey92:
### Confirmation

- [X] I have checked for similar issues.

### Problem

When op-node requests pop payouts, we need to paginate because this response is large and will not scale.

If I am not mistaken, this will require pagination in both bss and bfg getting pop payouts responses.

### Suggested solution

paginate bss and bfg pop payouts/pop tx responses when op node requests them

joshuasing:
@ClaytonNorthey92 Is this a duplicate of, or similar to #193?

ClaytonNorthey92:
hey @joshuasing it's a duplicate essentially, the only difference is that I added the pagination through BSS as well (so the full call path op-node -> bss -> bfg can be paginated)

Syuraih7:
i already clear all task but helmi pilot cant verify my action

PR #203, opened 2024-08-15
bump bss read limit to 10 mb

ClaytonNorthey92:
**Summary**
We have a stopgap in bss to increase the `ReadLimit` to handle the large number of pop payouts prior to implementing pagination.  We need to increase that for good measure because we're getting even more than expected.

This should keep us running until #195 gets deployed 

 
**Changes**
Increase `ReadLimit` to 10mb


joshuasing (approved):

marcopeereboom (approved):

marcopeereboom (changes requested):
I approved it initially but after thinking about it some it seems like these numbers are ridiculously large. We need root cause; this feels like a workaround.

PR #225, opened 2024-08-22
request limit updates to bfg

ClaytonNorthey92:
**Summary**
prior, request limiting was global (not per connection), so users' requests would block each other.  we would like to make this per-user

**Changes**
* made request limit per-connection on bfg

* added mutex per connection to only allow 1 request to be processed at a time for that connection

### DIFF
--- a/service/bfg/bfg.go
+++ b/service/bfg/bfg.go
@@ -107,10 +107,6 @@ type Server struct {
 
 	cfg *Config
 
-	// requests
-	requestLimiter chan bool // Maximum in progress websocket commands
-	// requestTimeout time.Duration
-
 	btcHeight uint64
 
 	server       *http.ServeMux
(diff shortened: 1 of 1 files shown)


service/bfg/bfg.go
```diff
@@ -267,12 +259,16 @@ func (s *Server) handleRequest(parentCtx context.Context, bws *bfgWs, wsid strin
...
 	select {
-	case <-s.requestLimiter:
+	case <-bws.requestLimiter:
+	case <-parentCtx.Done():
+		log.Debugf("handleRequest context done: %v", parentCtx.Err())
+		return
 	default:
-		log.Infof("Request limiter hit %v: %v", bws.addr, cmd)
-		<-s.requestLimiter
+		log.Infof("Request limiter hit %v (%v): %v", bws.addr, bws.sessionId, cmd)
```
ClaytonNorthey92, line 267: if the user has their requests limited, they're probably spamming us.  close their connection.

also, I believe closing this websocket will cause an error and thus will remove it from the sessions map

service/bfg/bfg.go
```diff
@@ -267,12 +259,16 @@ func (s *Server) handleRequest(parentCtx context.Context, bws *bfgWs, wsid strin
 	defer cancel()
 
 	select {
-	case <-s.requestLimiter:
+	case <-bws.requestLimiter:
```
ClaytonNorthey92, line 262: talking with @joshuasing ; it may be a good idea to _also_ have a global request limit too...need to think more about that 🤔

service/bfg/bfg.go
```diff
@@ -284,6 +280,10 @@ func (s *Server) handleRequest(parentCtx context.Context, bws *bfgWs, wsid strin
 
 	log.Tracef("Handling request %v: %v", bws.addr, cmd)
 
+	// ensure that only one handler is called per-time-per-connection
+	bws.mtx.Lock()
```
ClaytonNorthey92, line 284: hmm now that I think about it, this would cause longer running requests to block shorter ones, need to think a bit more 🤔

PR #227, opened 2024-08-23
add queuing system for bitcoin broadcasts

ClaytonNorthey92:
**Summary**
Add a system with a static number of workers to queue/consume bitcoin broadcasts to electrumx

**Changes**
add a db table btc_transaction_broadcast_request that will store broadcast requests.  when someone wants to broadcast a tx, insert into this table (ignoring duplicates).  have a few workers that read from this table and attempt to broadcast transactions to electrumx.

when a worker "checks out" a broadcast request, it updates the last_broadcast_attempt_at column

when a worker confirms a broadcast, it updates the broadcast_at column

wait a random number of  minutes between 1-5 before processing same tx again if not confirmed

don't process txs that are more than 30 minutes hours old

caching l2 keystones response

more efficient finality query

### DIFF
--- a/database/bfgd/postgres/postgres.go
+++ b/database/bfgd/postgres/postgres.go
@@ -1069,16 +1069,16 @@ func (p *pgdb) BtcTransactionBroadcastRequestGetNext(ctx context.Context, onlyNe
 	log.Tracef("BtcTransactionBroadcastRequestGetNext")
 	defer log.Tracef("BtcTransactionBroadcastRequestGetNext exit")
 
-	onlyNewClause := " next_broadcast_attempt IS NOT NULL AND next_broadcast_attempt < NOW() "
+	onlyNewClause := " next_broadcast_attempt_at IS NOT NULL AND next_broadcast_attempt_at <= NOW() "
 	if onlyNew {
-		onlyNewClause = " next_broadcast_attempt IS NULL "
+		onlyNewClause = " next_broadcast_attempt_at IS NULL "
 	}
 
 	querySql := fmt.Sprintf(`
 		UPDATE btc_transaction_broadcast_request 
 		SET last_broadcast_attempt_at = NOW(), 
 		
-		next_broadcast_attempt = NOW() + INTERVAL '1 minute' + RANDOM() * INTERVAL '60 seconds' 
+		next_broadcast_attempt_at = NOW() + INTERVAL '1 minute' + RANDOM() * INTERVAL '60 seconds' 
 		
 		WHERE tx_id = (
 			SELECT tx_id FROM btc_transaction_broadcast_request
--- a/database/bfgd/postgres/postgres.go
+++ b/database/bfgd/postgres/postgres.go
@@ -1083,7 +1083,7 @@ func (p *pgdb) BtcTransactionBroadcastRequestGetNext(ctx context.Context, onlyNe
 		UPDATE btc_transaction_broadcast_request 
 		SET last_broadcast_attempt_at = NOW(), 
 		
-		next_broadcast_attempt_at = NOW() + INTERVAL '1 minute' + RANDOM() * INTERVAL '60 seconds' 
+		next_broadcast_attempt_at = NOW() + INTERVAL '1 minute' + RANDOM() * INTERVAL '240 seconds' 
 		
 		WHERE tx_id = (
 			SELECT tx_id FROM btc_transaction_broadcast_request
--- a/database/bfgd/postgres/postgres.go
+++ b/database/bfgd/postgres/postgres.go
@@ -1042,3 +1042,86 @@ func (p *pgdb) refreshBTCBlocksCanonical(ctx context.Context) error {
 
 	return nil
 }
+
+func (p *pgdb) BtcTransactionBroadcastRequestInsert(ctx context.Context, serializedTx []byte, txId string) error {
+	log.Tracef("BtcTransactionBroadcastRequestInsert")
+	defer log.Tracef("BtcTransactionBroadcastRequestInsert exit")
+
+	const insertSql = `
+		INSERT INTO btc_transaction_broadcast_request 
+		(tx_id, serialized_tx)
+		VALUES ($1, $2)
+	`
+	_, err := p.db.ExecContext(ctx, insertSql, txId, serializedTx)
+	if err != nil {
+		var pgErr *pq.Error
+		if errors.As(err, &pgErr) && pgErr.Code.Class().Name() == "integrity_constraint_violation" {
+			return database.DuplicateError(fmt.Sprintf("duplicate entry: %s", pgErr))
+		}
+		return fmt.Errorf("failed to insert btc_transaction_broadcast_request: %w", err)
+	}
+
+	return nil
+}
+
+// BtcTransactionBroadcastRequestGetNext returns all broadcast requests that
+//  1. was last attempted over 30 seconds ago
+//  2. AND have never been broadcasted
+func (p *pgdb) BtcTransactionBroadcastRequestGetNext(ctx context.Context) ([]byte, error) {
+	log.Tracef("BtcTransactionBroadcastRequestGetNext")
+	defer log.Tracef("BtcTransactionBroadcastRequestGetNext exit")
+
+	const querySql = `
+		UPDATE btc_transaction_broadcast_request 
+		SET last_broadcast_attempt_at = NOW()
+		WHERE tx_id = (
+			SELECT tx_id FROM btc_transaction_broadcast_request
+			WHERE 
+			(
+				last_broadcast_attempt_at IS NULL
+				OR
+				last_broadcast_attempt_at < NOW() - INTERVAL '10 minutes'
+			)
+			AND broadcast_at IS NULL
+			ORDER BY created_at ASC
+			LIMIT 1
+		)
+		RETURNING serialized_tx
+	`
+	rows, err := p.db.QueryContext(ctx, querySql)
+	if err != nil {
+		return nil, fmt.Errorf("could not get next btc_transaction_broadcast_request: %v", err)
+	}
+
+	defer rows.Close()
+
+	for rows.Next() {
+		var serializedTx []byte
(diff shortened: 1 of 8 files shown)


service/bfg/bfg.go
```diff
@@ -317,6 +317,92 @@ func (s *Server) handleBitcoinBalance(ctx context.Context, bbr *bfgapi.BitcoinBa
...
+		hash := mb.TxHash()
+
+		if err := s.db.PopBasisInsertPopMFields(ctx, &bfgd.PopBasis{
+			BtcTxId:             hash[:],
+			BtcRawTx:            database.ByteArray(serializedTx),
+			PopMinerPublicKey:   publicKeyUncompressed,
+			L2KeystoneAbrevHash: tl2.L2Keystone.Hash(),
+		}); err != nil {
+			log.Errorf("error occurred inserting pop basis: %s", err)
+			continue
```
ClaytonNorthey92, line 383: don't continue if duplicate, confirm and drop

database/bfgd/scripts/0009.sql
```diff
@@ -0,0 +1,17 @@
+-- Copyright (c) 2024 Hemi Labs, Inc.
+-- Use of this source code is governed by the MIT License,
+-- which can be found in the LICENSE file.
+
+BEGIN;
+
+UPDATE version SET version = 9;
+
+CREATE TABLE btc_transaction_broadcast_request (
```
ClaytonNorthey92, line 9: add a partial index for querying

service/bfg/bfg.go
```diff
@@ -317,6 +317,92 @@ func (s *Server) handleBitcoinBalance(ctx context.Context, bbr *bfgapi.BitcoinBa
...
+		}
+
+		_, err = s.btcClient.Broadcast(ctx, serializedTx)
+		if err != nil {
+			log.Errorf("broadcast tx: %s", err)
```
ClaytonNorthey92, line 388: if we fail here, let's set the database record for an "immediate retry".  based on the rules this would take N minutes to retry.

database/bfgd/postgres/postgres.go
```diff
@@ -1041,3 +1041,86 @@ func (p *pgdb) refreshBTCBlocksCanonical(ctx context.Context) error {
...
+	const querySql = `
+		UPDATE btc_transaction_broadcast_request 
+		SET last_broadcast_attempt_at = NOW()
+		WHERE tx_id = (
+			SELECT tx_id FROM btc_transaction_broadcast_request
+			WHERE 
+			(
+				last_broadcast_attempt_at IS NULL
+				OR
+				last_broadcast_attempt_at < NOW() - INTERVAL '10 minutes'
```
ClaytonNorthey92, line 1082: perhaps this should be shorter; I want a way to retry broadcasts. perhaps 1 minute is better

e2e/e2e_ext_test.go
```diff
@@ -1432,6 +1437,7 @@ func TestBitcoinUTXOs(t *testing.T) {
 // 2. call BitcoinBroadcast RPC on BFG
 // 3. ensure that a pop_basis was inserted with the expected values
 func TestBitcoinBroadcast(t *testing.T) {
+	t.Skip()
```
ClaytonNorthey92, line 1440: whoops; don't skip this

ClaytonNorthey92:
tested this out, a few things need fixing:
* the index doesn't do anything with the query
* need a better way to determine retries, be smarter with errors

ClaytonNorthey92:
talking to @max-sanchez 

going to make these changes:
* 2 queues
  * high priority for sending txs for the first time
  * low priority for retries
* on success delete from queue
* after 30 minutes, if no success, delete from queue
* stagger retries
* add "last error seen" to db for debug
* some errors are higher priority than others, inspect.  assume all errors go into low priority except
  * "fees to low" 
  * invalid tx itself
  * mempool conflict
  * bad-txns-spends-conflicting-tx
* save all errors, log them out before delete, so they show up in papertrail

marcopeereboom (approved):

service/bfg/bfg.go
```diff
@@ -317,6 +319,102 @@ func (s *Server) handleBitcoinBalance(ctx context.Context, bbr *bfgapi.BitcoinBa
...
+		if tl2 == nil {
+			log.Errorf("could not find pop tx")
+			cancel()
+			continue
+		}
+
+		publicKeyUncompressed, err := pop.ParsePublicKeyFromSignatureScript(mb.TxIn[0].SignatureScript)
+		if err != nil {
+			log.Errorf("could not parse public key from signature script: %v", err)
+		}
```
joshuasing, line 378: Is this missing a `continue` and `cancel()`?

service/bfg/bfg.go
```diff
@@ -317,6 +319,102 @@ func (s *Server) handleBitcoinBalance(ctx context.Context, bbr *bfgapi.BitcoinBa
...
+
+		err = s.db.BtcTransactionBroadcastRequestConfirmBroadcast(ctx, mb.TxID())
+		if err != nil {
+			log.Errorf("could not confirm broadcast: %v", err)
+			cancel()
+			continue
+		}
+
+		log.Infof("successfully broadcast tx %s, for l2 keystone %s", mb.TxID(), hex.EncodeToString(tl2.L2Keystone.Hash()))
+		cancel()
```
joshuasing, line 414: I think it would be a good idea to extract this into a separate function, allowing us to `defer cancel()` and use `return`. This will prevent accidental context leaks.

service/bfg/bfg.go
```diff
@@ -317,6 +319,102 @@ func (s *Server) handleBitcoinBalance(ctx context.Context, bbr *bfgapi.BitcoinBa
...
+			if err != nil {
+				log.Errorf("could not delete %v", err)
+			}
+			cancel()
+			continue
+		}
+
+		s.metrics.popBroadcasts.Inc()
+
+		log.Infof("hash is %s", hex.EncodeToString(hash[:]))
```
joshuasing, line 404: This is going to be very noisy, could we please make it a debug log?

service/bfg/bfg.go
```diff
@@ -1198,16 +1275,35 @@ func (s *Server) handleBtcFinalityByKeystonesRequest(ctx context.Context, bfkr *
...
-	defer log.Tracef("handleL2KeystonesRequest exit")
+func (s *Server) getL2KeystonesCache() []hemi.L2Keystone {
+	log.Tracef("getL2KeystonesCache")
+	defer log.Tracef("getL2KeystonesCache exit")
+
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	if s.l2keystonesCache == nil {
+		return []hemi.L2Keystone{}
```
joshuasing, line 1286: Is there a reason to initialise this slice? If not, I think we should just return `nil`:
```suggestion
		return nil
```

service/bfg/bfg.go
```diff
@@ -1198,16 +1275,35 @@ func (s *Server) handleBtcFinalityByKeystonesRequest(ctx context.Context, bfkr *
...
+	}
+
+	results := []hemi.L2Keystone{}
+	for _, v := range s.l2keystonesCache {
+		results = append(results, v)
+	}
```
joshuasing, line 1292: I think pre-allocating the `results` slice and using the built-in `copy` function may improve performance here. (suggestion below is untested)
```suggestion
	results := make([]hemi.L2Keystone, len(s.l2keystonesCache))
	copy(results, s.l2keystonesCache)
```

service/bfg/bfg.go
```diff
@@ -1198,16 +1275,35 @@ func (s *Server) handleBtcFinalityByKeystonesRequest(ctx context.Context, bfkr *
...
+func (s *Server) refreshL2KeystoneCache(ctx context.Context) {
+	log.Tracef("refreshL2KeystoneCache")
+	defer log.Tracef("refreshL2KeystoneCache exit")
+
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
 
-	results, err := s.db.L2KeystonesMostRecentN(ctx, uint32(l2kr.NumL2Keystones))
+	results, err := s.db.L2KeystonesMostRecentN(ctx, 100)
 	if err != nil {
```
joshuasing, line 1305: This error check no longer returns, is that intentional?

service/bfg/bfg.go
```diff
@@ -1307,22 +1419,20 @@ func (s *Server) handleNewL2Keystones(ctx context.Context, nlkr *bfgapi.NewL2Key
...
-			response.Error = protocol.Errorf("invalid l2 keystone")
-			return response, nil
+
+	go func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		ks := hemiL2KeystonesToDb(nlkr.L2Keystones)
+		err := s.db.L2KeystonesInsert(ctx, ks)
+		if err != nil {
```
joshuasing, line 1430: This error check does not have a return, is that intentional?

joshuasing (changes requested):

joshuasing (approved):
Unblocking merge, as this pull request is urgent.

service/bfg/bfg.go
ClaytonNorthey92: yes! thank you

service/bfg/bfg.go
ClaytonNorthey92: ok let me see what I can do 🕵️‍♂️

service/bfg/bfg.go
ClaytonNorthey92: ah yes thank you

service/bfg/bfg.go
ClaytonNorthey92: ah no, some fallout from previous iteration, will change

service/bfg/bfg.go
ClaytonNorthey92: will update

service/bfg/bfg.go
ClaytonNorthey92: nice catch, just a mistake

PR #228, opened 2024-08-23
Dynamic gas

kingdancing:
**Summary**

Using a Dynamic gas.
**Changes**

### DIFF
--- a/service/popm/popm.go
+++ b/service/popm/popm.go
@@ -11,8 +11,10 @@ import (
 	"crypto/rand"
 	"crypto/sha256"
 	"encoding/hex"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"math/big"
 	"net/http"
 	"slices"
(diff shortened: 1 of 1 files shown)


service/popm/popm.go
```diff
@@ -355,7 +393,12 @@ func (m *Miner) mineKeystone(ctx context.Context, ks *hemi.L2Keystone) error {
 
 	// Estimate BTC fees.
 	txLen := 285 // XXX: for now all transactions are the same size
-	feePerKB := 1024 * m.cfg.StaticFee
+	url := "https://mempool.space/testnet/api/v1/fees/recommended"
```
ClaytonNorthey92, line 396: I don't think we want to rely on `mempool.space` for fee estimation, we can use the existing electrumx connection and perhaps the [fee estimation](https://electrumx.readthedocs.io/en/latest/protocol-methods.html#blockchain-estimatefee) method.

ClaytonNorthey92 (changes requested):
thank you for the contribution, I left one comment to start; we _probably_ want to estimate fees via electrumx, as that is our entrypoint to the bitcoin network.  if you feel strongly about this change, could you update it to use electrumx then tag us and ask us to review again?

we will likely have more review comments referring to your strategy at that point, but for starters let's use electrumx

jcvernaleo (changes requested):
You should not include a binary file in the repo.

service/popm/popm.go
```diff
@@ -85,11 +87,47 @@ type Config struct {
...
+// GetFastestFee 发送 GET 请求并获取 fastestFee
+func GetFastestFee(url string) (uint, error) {
+	// 发送 GET 请求
+	resp, err := http.Get(url)
+	if err != nil {
+		return 0, fmt.Errorf("failed to send GET request: %v", err)
+	}
+	defer resp.Body.Close()
+
+	// 读取响应体
```
jcvernaleo, line 108: All comments should be in English to match the rest of the repo.

jcvernaleo (changes requested):

jcvernaleo:
Based on the binary file inclusion, I'm closing this PR.

If you want to re submit without a binary file, we'd be happy to review that again.

Issue #230, opened 2024-08-26
Finalities by keystone query is inefficient


ClaytonNorthey92:
### Confirmation

- [X] I have checked for similar issues.

### Describe the bug

the query in `L2BTCFinalityByL2KeystoneAbrevHash` is inefficient and I think it is causing delays in the endpoints that use it.

```sql
                                                                              QUERY PLAN                                                                              
----------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop Left Join  (cost=7387.73..434999296.77 rows=368 width=242) (actual time=671.681..671.694 rows=0 loops=1)
   InitPlan 2 (returns $1)
     ->  Aggregate  (cost=1898.41..1898.42 rows=1 width=8) (never executed)
           ->  Seq Scan on btc_blocks_can btc_blocks_can_2  (cost=0.00..1749.13 rows=59713 width=8) (never executed)
   ->  Index Scan Backward using l2_keystone_l2_block_number_idx on l2_keystones  (cost=0.29..2588.48 rows=1 width=185) (actual time=671.680..671.680 rows=0 loops=1)
         Filter: (l2_keystone_abrev_hash = ANY ('{}'::bytea[]))
         Rows Removed by Filter: 22443
   ->  Hash Right Join  (cost=5489.02..7544.47 rows=1396 width=74) (never executed)
         Hash Cond: (btc_blocks_can.hash = pop_basis.btc_block_hash)
         ->  Seq Scan on btc_blocks_can  (cost=0.00..1749.13 rows=59713 width=41) (never executed)
         ->  Hash  (cost=5471.57..5471.57 rows=1396 width=66) (never executed)
               ->  Bitmap Heap Scan on pop_basis  (cost=23.25..5471.57 rows=1396 width=66) (never executed)
                     Recheck Cond: (l2_keystones.l2_keystone_abrev_hash = l2_keystone_abrev_hash)
                     ->  Bitmap Index Scan on l2_keystone_abrev_hash_idx  (cost=0.00..22.90 rows=1396 width=0) (never executed)
                           Index Cond: (l2_keystone_abrev_hash = l2_keystones.l2_keystone_abrev_hash)
   SubPlan 1
     ->  Aggregate  (cost=1182030.56..1182030.57 rows=1 width=8) (never executed)
           ->  Hash Join  (cost=3479.05..1178190.69 rows=1535951 width=8) (never executed)
[...]

PR #231, opened 2024-08-26
added indexes and removed aggregate functions

ClaytonNorthey92:
added indexes to btc_blocks_can materialized view, no longer using aggregate functions

**Summary**
performance improvements for l2 keystones by finality

**Changes**
added indexes to btc_blocks_can materialized view, no longer using aggregate functions

without pasting the entire query plan, this reduces the query to the following (a high cost, but fast):
```sql
(cost=19.64..85167.78 rows=394 width=242) (actual time=81.977..81.990 rows=1 loops=1)
```

prior, it was so long that I could not run `analyze`
```sql
 Nested Loop Left Join  (cost=51725.76..472025135.82 rows=394 width=242)
```

### DIFF
--- a/database/bfgd/postgres/postgres.go
+++ b/database/bfgd/postgres/postgres.go
@@ -20,14 +20,14 @@ import (
 )
 
 const (
-	bfgdVersion = 8
+	bfgdVersion = 9
 
 	logLevel = "INFO"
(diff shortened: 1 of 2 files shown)


ClaytonNorthey92:
fixes #230

PR #232, opened 2024-08-27
Add mechanism to recover from forks at runtime

marcopeereboom:
**Summary**
During runtime we quiesce p2p when indexing to prevent index corruption. When tbc unquiesces several p2p commands may have happened, including a forking event, which wasn't tracked. When tbc tries to recover it may end up with a chain geometry it currently cannot process.

**Changes**
- Remove index flag and overload that into the quiesce flag
- Add a function that can determine the common ancestor between two blocks.
- Add a function that determines the current chain state and unwinds indexes to a common ancestor and then winds them back to tip.

### DIFF
--- a/database/tbcd/level/level.go
+++ b/database/tbcd/level/level.go
@@ -441,29 +441,6 @@ func (l *ldb) BlockHeadersInsert(ctx context.Context, bhs *wire.MsgHeaders) (tbc
 			errors.New("block headers insert: invalid")
 	}
 
-	// Iterate over the block headers and skip block headers we already
-	// have in the database. Rely on caching to make this not suck terribly.
-	bhDB := l.pool[level.BlockHeadersDB]
-	var x int
-	for _, rbh := range bhs.Headers {
-		bhash := rbh.BlockHash()
-		has, err := bhDB.Has(bhash[:], nil)
-		if err != nil {
-			return tbcd.ITInvalid, nil, nil, 0,
-				fmt.Errorf("block headers insert has: %w", err)
-		}
-		if has {
-			x++
-		} else {
-			break
-		}
-	}
-	bhs.Headers = bhs.Headers[x:]
-	if len(bhs.Headers) == 0 {
-		return tbcd.ITInvalid, nil, nil, 0,
-			database.DuplicateError("block headers insert duplicate")
-	}
-
 	// Ensure we can connect these blockheaders prior to starting database
 	// transaction. This also obtains the starting cumulative difficulty
 	// and  height.
@@ -481,6 +458,28 @@ func (l *ldb) BlockHeadersInsert(ctx context.Context, bhs *wire.MsgHeaders) (tbc
 			fmt.Errorf("block headers open transaction: %w", err)
 	}
 	defer bhsDiscard()
+
+	// Iterate over the block headers and skip block headers we already
+	// have in the database. Rely on caching to make this not suck terribly.
+	var x int
+	for _, rbh := range bhs.Headers {
+		bhash := rbh.BlockHash()
+		has, err := bhsTx.Has(bhash[:], nil)
+		if err != nil {
+			return tbcd.ITInvalid, nil, nil, 0,
+				fmt.Errorf("block headers insert has: %w", err)
+		}
+		if has {
+			x++
+		} else {
+			break
+		}
+	}
+	bhs.Headers = bhs.Headers[x:]
+	if len(bhs.Headers) == 0 {
+		return tbcd.ITInvalid, nil, nil, 0,
+			database.DuplicateError("block headers insert duplicate")
+	}
 
 	// blocks missing
 	bmTx, bmCommit, bmDiscard, err := l.startTransaction(level.BlocksMissingDB)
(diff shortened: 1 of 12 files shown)


database/tbcd/level/level.go
```diff
@@ -411,70 +436,75 @@ func (l *ldb) BlockHeadersInsert(ctx context.Context, bhs [][80]byte) (tbcd.Inse
...
+		has, err := bhsTx.Has(bhash[:], nil)
+		if err != nil {
+			return tbcd.ITInvalid, nil, nil, 0,
+				fmt.Errorf("block headers insert has: %w", err)
+		}
+		if has {
+			x++
+		} else {
+			break
+		}
```
joshuasing, line 476: I think this could be simplified:
```suggestion
		if !has {
			break
		}
		x++
```

ClaytonNorthey92 (approved):

service/tbc/tbc.go
```diff
@@ -894,11 +1012,13 @@ func (s *Server) handleAddr(_ context.Context, p *peer, msg *wire.MsgAddr) {
 	}
 
 	if err := s.pm.PeersInsert(peers); err != nil {
-		log.Errorf("Insert peers: %v", err)
+		return fmt.Errorf("Insert peers: %w", err)
```
joshuasing, line 1015: Error strings shouldn't start with capitals, per https://go.dev/wiki/CodeReviewComments#error-strings
```suggestion
		return fmt.Errorf("insert peers: %w", err)
```

service/tbc/tbc.go
```diff
@@ -913,33 +1033,39 @@ func (s *Server) handleAddrV2(_ context.Context, p *peer, msg *wire.MsgAddrV2) {
...
```
joshuasing, line 1036: Error strings shouldn't start with capitals, per https://go.dev/wiki/CodeReviewComments#error-strings
```suggestion
		return fmt.Errorf("insert peers: %w", err)
```

service/tbc/tbc.go
```diff
@@ -1030,14 +1203,10 @@ func (s *Server) syncBlocks(ctx context.Context) {
...
-		s.quiesced = true // XXX if it's set and we exit with an error, what should we do??
 		go func() {
-			// we really want to push the indexing reentrancy into this call
-			log.Infof("quiescing p2p and indexing to: %v @ %v",
-				bhb, bhb.Height)
 			if err = s.SyncIndexersToHash(ctx, bhb.BlockHash()); err != nil {
-				log.Errorf("sync blocks: %v", err)
-				return
+				// XXX this a panic?
+				panic(fmt.Errorf("sync blocks: %v", err))
```
joshuasing, line 1209: This is a panic, so it doesn't matter much, but staticcheck may complain:
```suggestion
				panic(fmt.Errorf("sync blocks: %w", err))
```

service/tbc/tbc.go
```diff
@@ -1089,43 +1258,36 @@ func (s *Server) handleHeaders(ctx context.Context, p *peer, msg *wire.MsgHeader
...
 			//	log.Errorf("get headers %v: %v", p, err)
 			//	return
 			//}
-			return
+			return nil
 		}
 		// Real error, abort header fetch
-		log.Errorf("block headers insert: %v", err)
-		return
+		return fmt.Errorf("block headers insert: %v", err)
```
joshuasing, line 1290: I think the `%w` verb can be used here to wrap the error (and in all of the `fmt.Errorf`s below):
```suggestion
		return fmt.Errorf("block headers insert: %w", err)
```

service/tbc/tbc.go
```diff
@@ -1166,46 +1325,42 @@ func (s *Server) handleHeaders(ctx context.Context, p *peer, msg *wire.MsgHeader
...
+		return fmt.Errorf("handle block unable to get raw block %v: %w", bhs, err)
 	}
 
 	if s.cfg.BlockSanity {
 		err := blockchain.CheckBlockSanity(block, s.chainParams.PowLimit,
 			s.timeSource)
 		if err != nil {
-			log.Errorf("Unable to validate block hash %v: %v", bhs, err)
-			return
+			return fmt.Errorf("handle block unable to validate block hash %v: %v",
```
joshuasing, line 1362: I think the `%w` verb can be used here to wrap the error:
```suggestion
			return fmt.Errorf("handle block unable to validate block hash %v: %w",
```

joshuasing (changes requested):

service/tbc/crawler.go
```diff
@@ -470,6 +469,13 @@ func (s *Server) UtxoIndexerUnwind(ctx context.Context, startBH, endBH *tbcd.Blo
 	defer log.Tracef("UtxoIndexerUnwind exit")
 
 	// XXX dedup with TxIndexedWind; it's basically the same code but with the direction, start anf endhas flipped
+	s.mtx.Lock()
+	if !s.indexing {
```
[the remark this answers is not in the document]
marcopeereboom: I am not ready for this yet. Once this settles down we can have this debate.

ttl/ttl.go
```diff
@@ -9,22 +9,36 @@ import (
 	"errors"
 	"sync"
 	"time"
+
+	"github.com/juju/loggo"
+)
+
+var (
+	logLevel = "INFO"
+	log      = loggo.GetLogger("ttl")
```
joshuasing, line 18: This variable is unused, which makes the `init` function and `logLevel` variable redundant.

joshuasing (approved):

ttl/ttl.go
marcopeereboom: linter can bite me. I keep adding and removing this same block when debugging. So let's leave it in place but remain silent by default.

PR #235, opened 2024-08-27
trim old, new index

ClaytonNorthey92:
**Summary**
trim queue items older than an hour

add index for retries, using sorted `created_at`

**Changes**
trim queue items older than an hour

add index for retries, using sorted `created_at`

### DIFF
--- a/database/bfgd/postgres/postgres.go
+++ b/database/bfgd/postgres/postgres.go
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	bfgdVersion = 9
+	bfgdVersion = 10
 
 	logLevel = "INFO"
 	verbose  = false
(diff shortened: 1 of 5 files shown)


joshuasing (approved):

database/bfgd/postgres/postgres.go
```diff
@@ -1150,3 +1150,19 @@ func (p *pgdb) BtcTransactionBroadcastRequestSetLastError(ctx context.Context, t
 
 	return nil
 }
+
+func (p *pgdb) BtcTransactionBroadcastRequestTrim(ctx context.Context) error {
```
ClaytonNorthey92, line 1154: I am going to add a test around this before merge

max-sanchez (approved):

PR #239, opened 2024-08-29
more fragile test fixes

ClaytonNorthey92:
instead of ignoring the wrong commands, wait until the correct one until the test times out

**Summary**

**Changes**


jcvernaleo (approved):
I wish the wait times were less all over the place but guess not much to be done about that.

OK

ClaytonNorthey92:
> I wish the wait times were less all over the place but guess not much to be done about that.
> 
> OK

@jcvernaleo yeah the "waits" could be smarter, I'll admit

PR #242, opened 2024-08-30
remove insertion of pop_basis during broadcast

ClaytonNorthey92:
we (potentially) insert pop_basis twice, once when broadcasting, another when parsing btc blocks.  we don't need the former at this point, remove it so we take stress off of the db.

**Summary**
we (potentially) insert pop_basis twice, once when broadcasting, another when parsing btc blocks.  we don't need the former at this point

**Changes**
remove the first insert so we take stress off of the db


joshuasing (approved):

jcvernaleo (approved):
OK

PR #245, opened 2024-09-05
need to select FOR UPDATE to lock row

ClaytonNorthey92:
**Summary**
prior, when selecting a row to update, we were not locking the row selected so multiple processes could get this row at the same time.

**Changes**
select row `FOR UPDATE` to lock row to ensure only one process gets it


joshuasing (approved):

jcvernaleo (approved):
OK

PR #246, opened 2024-09-05
bfgd: add handling for remote IP headers

joshuasing:
**Summary**
Add handling for remote IP headers added by proxies for requests to BFG, such as `X-Forwarded-For`.

The `X-Forwarded-For` header is untrustworthy when the request comes from a source that is not a trusted proxy. This header is inherently dangerous, as it can be changed and used to spoof the client IP address.

**Improper parsing and handling of these headers can have serious security implications, as such this implementation follows all known best practices, and should be reviewed thoroughly.**

There can be multiple `X-Forwarded-For` headers present in the request, and the IP addresses in the headers must be treated as a single list of IP addresses, starting with the first IP of the first header and ending with the last IP of the last header.

The first "trustworthy" client IP address from these headers will be selected by searching the IP addresses in reverse order, skipping all addresses that are trusted proxies, and then using the first untrusted IP address.

Mozilla has some best practices for parsing this header here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

---

To use this functionality in BFG, two new environment variables must be set (not setting either will disable this functionality):
 - `BFG_TRUSTED_PROXIES`: Must be set to a comma-separated list of IP addresses or CIDR ranges for the proxies trusted by BFG. If the request comes from an address that is not trusted, the source IP will be used instead of attempting to parse any remote IP headers. For example, to trust the proxy `10.0.1.5` and in the CIDR `10.0.2.0/24`, set `BFG_TRUSTED_PROXIES="10.0.1.5,10.0.2.0/24"`
 - `BFG_REMOTE_IP_HEADERS`: Must be set to a comma-separated list of HTTP header names in which to search for remote IP addresses. For example, to use the `X-Forwarded-For`, set `BFG_REMOTE_IP_HEADERS="X-Forwarded-For"`.

**Changes**
 - Add remote IP header handling in BFG.


marcopeereboom (approved):
All these settings are proof positive that we should burn kubernetes to the ground.

marcopeereboom (approved):
needing config for k8s is terrible

jcvernaleo (approved):
OK

service/bfg/bfg.go
```diff
@@ -1006,8 +1012,9 @@ func (s *Server) deleteSession(id string) {
 }
 
 func (s *Server) handleWebsocketPrivate(w http.ResponseWriter, r *http.Request) {
-	log.Tracef("handleWebsocketPrivate: %v", r.RemoteAddr)
-	defer log.Tracef("handleWebsocketPrivate exit: %v", r.RemoteAddr)
+	remoteAddr := s.remoteIP(r)
```
ClaytonNorthey92, line 1015: are we handling for a case where `remoteAddr == ""` in any uses of `s.remoteIP(...)`?

service/bfg/bfg.go
```diff
@@ -1766,3 +1786,122 @@ func (s *Server) Run(pctx context.Context) error {
 
 	return err
 }
+
+// remoteIP returns the remote client IP address for the http request.
+func (s *Server) remoteIP(req *http.Request) string {
+	remoteAddr, _, err := net.SplitHostPort(req.RemoteAddr)
+	if err != nil {
+		return ""
```
ClaytonNorthey92, line 1794: should we at least log this error out?

service/bfg/bfg.go
joshuasing: If `remoteIP` returns `""` then `req.RemoteAddr` is invalid (either not a hostport, or not a valid IP address).
I don't think that can ever happen, and if it does then something is very wrong.

service/bfg/bfg.go
joshuasing: We could add a log here, but this should never happen.

service/bfg/bfg.go
ClaytonNorthey92: > and if it does then something is very wrong.

should we error in this case?

ClaytonNorthey92 (approved):

ClaytonNorthey92 (approved):

PR #249, opened 2024-09-05
bfg: check if msg will be logged before logging spew.Sdump

joshuasing:
**Summary**
Add check for whether log level is enabled before logging `spew.Sdump` outputs.
Calling `spew.Sdump` is rather expensive, and unnecessary when the output is going to be ignored.

**Changes**
- Add check for whether log level is enabled before logging `spew.Sdump` outputs in BFG and protocol.


jcvernaleo (approved):
OK

marcopeereboom (changes requested):
This is an issue throughout the code. We should fix this forever by forking spew and cutting it off at the source. Alternatively, btcsuite uses a closure that does this but that becomes unreadable real quick. Let's stay away from this check and fix it for real.

PR #251, opened 2024-09-10
popm/wasm: make automaticFees default to enabled

joshuasing:
**Summary**
The documentation for `@hemilabs/pop-miner` states that `automaticFees` defaults to `true`:
```typescript
  /**
   * Whether to do automatic fee estimation using the mempool.space API.
   * Defaults to true. If disabled, then only the staticFee will be used.
   *
   * When the value is true, the 'economy' recommended fee type will be used.
   */
  automaticFees?: boolean | RecommendedFeeType;
```

This makes the code behave as documented.

**Changes**
- Enable automatic fees in the WebAssembly PoP Miner when `automaticFees` is `undefined`.


ArturDolzan (approved):
LGTM

ClaytonNorthey92 (approved):

PR #256, opened 2024-09-12
popm/wasm: add automatic fee multiplier, fix auto fee args

joshuasing:
**Summary**
Add `automaticFeeMultiplier` parameter to `MinerStartArgs` in the WebAssembly PoP Miner.

This allows the used fee to be increased or decreased by a percentage. For example, if the recommended fee returned by the `mempool.space` API is `500`, and the multiplier is set to `1.1`, the output fee will be `550` sats/vB.

**Changes**
- Add `automaticFeeMultiplier?: number` to `MinerStartArgs` (defaults to `1.1`, +10%).
- Make `automaticFees: true` use the `fastest` recommended fee by default.
- Add range check for `automaticFeeRefreshSeconds` (if set, must be greater than zero).

### DIFF
--- a/web/packages/pop-miner/src/browser/index.ts
+++ b/web/packages/pop-miner/src/browser/index.ts
@@ -59,6 +59,9 @@ export const startPoPMiner: typeof types.startPoPMiner = (args) => {
     network: args.network,
     privateKey: args.privateKey,
     logLevel: args.logLevel ?? '',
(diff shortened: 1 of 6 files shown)


jcvernaleo (approved):
OK

ArturDolzan (approved):
LGTM

PR #257, opened 2024-09-12
add tx id to broadcast error log so we can trace it

ClaytonNorthey92:
**Summary**
when debugging by looking through logs, it's useful to be able to associate a log with a tx id, we currently don't do this on broadcast errors

**Changes**
add tx id to broadcast error


jcvernaleo (approved):
OK

joshuasing (approved):

PR #258, opened 2024-09-16
electrumx: move conn metrics into clientConn

joshuasing:
**Summary**
Move ElectrumX connection metric handling (`connections_opened`, `connections_open`, `connections_closed`) into the individual connection handling code.

This aims to fix an inconsistency where if multiple attempts are made to close a connection, `connections_open` can be decremented multiple times for the one connection, causing the number to potentially be lower than the real number (and possibly negative).

**Changes**
- Move handling for ElectrumX metrics `connections_opened`, `connections_open` and `connections_closed` to the `clientConn` struct in `conn.go`.
- Set `(*clientConn).conn` to `nil` when `(*clientConn).Close()` is called and the connection has been closed, preventing attempts to close the connection multiple times (`net.ErrClosed` will be returned).
- Return `net.ErrClosed` in `(*clientConn).call` if `(*clientConn).conn` is `nil` due to the connection being closed.

### DIFF
--- a/hemi/electrumx/conn.go
+++ b/hemi/electrumx/conn.go
@@ -48,6 +48,11 @@ func newClientConn(conn net.Conn, metrics *metrics, onClose func(c *clientConn))
 		metrics: metrics,
 	}
 
+	if metrics != nil {
+		metrics.connsOpened.Inc()
+		metrics.connsOpen.Inc()
+	}
+
 	go c.pinger()
 	return c
 }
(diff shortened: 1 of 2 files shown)


hemi/electrumx/conn.go
```diff
@@ -48,6 +48,11 @@ func newClientConn(conn net.Conn, metrics *metrics, onClose func(c *clientConn))
 		metrics: metrics,
 	}
 
+	if metrics != nil {
+		metrics.connsOpened.Inc()
+		metrics.connsOpen.Inc()
```
marcopeereboom, line 53: Ah yes, OO induced races. *sigh*

marcopeereboom (approved):

ClaytonNorthey92 (approved):

PR #264, opened 2024-09-24
Add mempool handling to tbc

marcopeereboom:
**Summary**

**Changes**
This adds a mempool struct that adds/removes tx's from the mempool. Also add some pretty stats to print during block download.

### DIFF
--- a/service/tbc/crawler.go
+++ b/service/tbc/crawler.go
@@ -69,8 +69,11 @@ func (s *Server) TxIndexHash(ctx context.Context) (*HashHeight, error) {
 }
 
 func (s *Server) findCommonParent(ctx context.Context, bhX, bhY *tbcd.BlockHeader) (*tbcd.BlockHeader, error) {
-	// This function assumes that the highest block height connects to the
-	// lowest block height.
+	// This function has one odd corner case. If bhX and bhY are both on a
+	// "long" chain without multiple blockheaders it will terminate on the
+	// first height that has a single blockheader. This is to be expected!
+	// This function "should" be called between forking blocks and then
+	// it'll find the first common parent.
 
 	// 0. If bhX and bhY are the same return bhX.
 	if bhX.Hash.IsEqual(bhY.Hash) {
--- a/service/tbc/crawler.go
+++ b/service/tbc/crawler.go
@@ -23,12 +23,28 @@ import (
 	"github.com/hemilabs/heminetwork/database/tbcd"
 )
 
+func s2h(s string) chainhash.Hash {
+	h, err := chainhash.NewHashFromStr(s)
+	if err != nil {
+		panic(err)
+	}
+	return *h
+}
+
 var (
 	UtxoIndexHashKey = []byte("utxoindexhash") // last indexed utxo hash
 	TxIndexHashKey   = []byte("txindexhash")   // last indexed tx hash
 
 	ErrNotLinear       = errors.New("not linear") // not a valid chain
 	ErrAlreadyIndexing = errors.New("already indexing")
+
+	testnet3Checkpoints = map[chainhash.Hash]uint64{
+		s2h("0000000000003c46fc60e56b9c2ae202b1efec83fcc7899d21de16757dea40a4"): 3000000,
+		s2h("000000000000001669469c0354b3f341a36b10ab099d1962f7ec4fae528b1f1d"): 2900000,
+		s2h("000000000000010dd0863ec3d7a0bae17c1957ae1de9cbcdae8e77aad33e3b8c"): 2000000,
+		s2h("0000000000478e259a3eda2fafbeeb0106626f946347955e99278fe6cc848414"): 1000000,
+		s2h("000000000001a7c0aaa2630fbb2c0e476aafffc60f82177375b2aaa22209f606"): 500000,
+	}
 )
 
 type HashHeight struct {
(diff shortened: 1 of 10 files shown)


joshuasing (approved):

service/tbc/tbc.go
```diff
@@ -906,6 +934,20 @@ func (s *Server) peerConnect(ctx context.Context, peerC chan string, p *peer) {
...
 
+	// Get p2p information.
+	err = p.write(defaultCmdTimeout, wire.NewMsgGetAddr())
+	if err != nil && !errors.Is(err, net.ErrClosed) {
+		log.Errorf("peer get addr: %v", err)
+		return
+	}
+
+	// Start building the mempool.
+	err = p.write(defaultCmdTimeout, wire.NewMsgMemPool())
```
ClaytonNorthey92, line 945: should we only do this if `cfg.MempoolEnabled = true`?

service/tbc/tbc.go
```diff
@@ -1469,37 +1515,46 @@ func (s *Server) handleInv(ctx context.Context, p *peer, msg *wire.MsgInv, raw [
...
+			log.Debugf("inventory filtered witness block: %v", v.Hash)
 		default:
-			log.Infof("inventory unknown: %v", spew.Sdump(v.Hash))
+			log.Errorf("inventory unknown: %v", spew.Sdump(v.Hash))
+		}
+	}
+
+	if s.cfg.MempoolEnabled && txsFound {
+		if err := s.mempool.invTxsInsert(ctx, msg); err != nil {
+			go s.downloadMissingTx(ctx, p)
```
ClaytonNorthey92, line 1545: is the _only error we'll ever get from invTxsInsert_ a missing tx? it may be the case now, but this may be a gotcha if we end up returning different errors from `invTxsInsert`.  Can we do a check that the err is indeed "tx missing"?

service/tbc/mempool.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/davecgh/go-spew/spew"
+)
+
+type mempool struct {
+	mtx sync.RWMutex
+
+	txs  map[chainhash.Hash][]byte // when nil, tx has not been downloaded
+	size int                       // total memory used by mempool
```
ClaytonNorthey92, line 22: should size be a `uint`?

service/tbc/mempool.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (m *mempool) txsInsert(ctx context.Context, msg *wire.MsgTx, raw []byte) error {
+	log.Tracef("txsInsert")
+	defer log.Tracef("txsInsert exit")
+
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+
+	if tx := m.txs[msg.TxHash()]; tx == nil {
```
ClaytonNorthey92, line 56: this will return `true` for both txs that are in the map and have a `nil` value as well as ones that are not in the map, is this what we want?

exemplified https://go.dev/play/p/RUqBqPkdr5R

ClaytonNorthey92: should we check `ok` instead?

https://go.dev/play/p/qqWF-vlgaSK

ClaytonNorthey92 (changes requested):

service/tbc/mempool.go
ClaytonNorthey92: the reason that I ask is because `invTxsInsert` sets keys to `nil`, so I assume there is a different between `nil` value and key not existing in the map.  I want to be sure this is safe with that assumption.

service/tbc/crawler.go
```diff
@@ -67,6 +68,56 @@ func (s *Server) TxIndexHash(ctx context.Context) (*HashHeight, error) {
 	return s.mdHashHeight(ctx, TxIndexHashKey)
 }
 
+func (s *Server) findCommonParent(ctx context.Context, bhX, bhY *tbcd.BlockHeader) (*tbcd.BlockHeader, error) {
```
ClaytonNorthey92, line 71: if this isn't used, let's leave it out of this pr

service/tbc/mempool.go
```diff
@@ -0,0 +1,126 @@
...
+		m.txs[msg.TxHash()] = raw
+		m.size += len(raw)
```
ClaytonNorthey92, line 58: though it's unlikely, should we protect against overflow of this value? 🤔

service/tbc/crawler.go
marcopeereboom: It is now.

service/tbc/tbc.go
marcopeereboom: That is correct and this code is ossified.

service/tbc/mempool.go
marcopeereboom: We'll certainly have OOM'd before that ever happens :-)

service/tbc/mempool.go
marcopeereboom: Sharp observation but this is deliberate in order to deal with the the p2p protocol. You have to issue a inv, that returns *just* a hash. We use that condition to key off and issue a getdata. If tx == nil it means we have to still download it. This is awkward but correct.

ClaytonNorthey92 (approved):

PR #271, opened 2024-10-01
Add a 'b' key to record whoch blocks have been indexed

marcopeereboom:
**Summary**

**Changes**

### DIFF
--- a/database/tbcd/level/level.go
+++ b/database/tbcd/level/level.go
@@ -980,6 +980,10 @@ func (l *ldb) BlockTxUpdate(ctx context.Context, direction int, txs map[tbcd.TxK
 	}
 	defer txsDiscard()
 
+	block := make([]byte, 33)
+	block[0] = 'b'
+	var blk []byte
+
 	txsBatch := new(leveldb.Batch)
 	for k, v := range txs {
 		// cache is being emptied so we can slice it here.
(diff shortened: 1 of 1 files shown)


ClaytonNorthey92 (approved):

database/tbcd/level/level.go
```diff
@@ -989,17 +995,33 @@ func (l *ldb) BlockTxUpdate(ctx context.Context, direction int, txs map[tbcd.TxK
 			key = k[0:65]
 			value = nil
 
+			// insert block hash to determine if it was indexed later
+			if _, ok := bm[string(k[33:65])]; !ok {
+				copy(block[1:], k[33:65])
```
joshuasing, line 1000: I think we could avoid reslicing `key` twice by adding a variable:
```suggestion
			hash := k[33:65]
			if _, ok := bm[string(hash)]; !ok {
				copy(block[1:], hash)
```

joshuasing (changes requested):

database/tbcd/level/level.go
marcopeereboom: More often than not it will collide and then we make a variable that has single use. Also, that is cached in the tight loop.

cmd/hemictl/hemictl.go
```diff
@@ -261,18 +261,17 @@ func tbcdb() error {
 
 	case "blocksmissing":
 		count := args["count"]
-		c, err := strconv.ParseUint(count, 10, 64)
+		c, err := strconv.ParseInt(count, 10, 64)
```
joshuasing, line 264: This is potentially cast to a 32-bit integer (depending on the platform) below, instead of parsing as a 64-bit int and casting to an int - could we parse as a 32-bit integer? I'm not sure we will need to use count > `4,294,967,296`.
```suggestion
		c, err := strconv.ParseInt(count, 10, 32)
```

joshuasing (changes requested):

joshuasing (approved):

PR #274, opened 2024-10-03
Replace electrumx with electrs

ClaytonNorthey92:
**Summary**
We want to replace electrumx with electrs, here is how we can do it in heminetwork.

 
**Changes**
* update rpc interface to implement electrs rpc; the main thing here is that parameters are now passed in based on their position, rather than json object key/value pairings
* replace "electrumx" with "electrs" in source code
* update localnet to use electrs

### DIFF
--- a/hemi/electrs/electrs.go
+++ b/hemi/electrs/electrs.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by the MIT License,
 // which can be found in the LICENSE file.
 
-package electrumx
+package electrs
 
 import (
 	"context"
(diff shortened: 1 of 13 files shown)


hemi/electrs/electrs.go
```diff
@@ -23,12 +23,12 @@ import (
...
 
-var log = loggo.GetLogger("electrumx")
+var log = loggo.GetLogger("electrs")
 
 // Prometheus subsystem name.
-const promSubsystem = "electrumx"
+const promSubsystem = "electrs"
 
-// https://electrumx.readthedocs.io/en/latest/protocol-basics.html
+// https://electrs.readthedocs.io/en/latest/protocol-basics.html
```
ClaytonNorthey92, line 31: update

ClaytonNorthey92:
@joshuasing I think I want to add a workflow to build and push an electrs tag from the repo that we can use in our networks, what do you think? I don't think that electrs has an official (working) docker repo/pipeline.

joshuasing (approved):

PR #276, opened 2024-10-04
stability and fragile test fixes

ClaytonNorthey92:
**Summary**
if we get multiple keystones before we mine one, there is a chance that keystones get queued up twice; we copy the known keystones on each reception, then we try to mine, but we only lock during the copying step, so keystones may be copied twice in two different go routines then mined

don't mine known keystones in goroutines, it isn't helping

use less than comparison in tests, if an rpc gets called more than the expected number of times, this is ok

add consistent (longer) timeout to tests

**Changes**
see above


ClaytonNorthey92:
I ran this 5 times (before rebasing) and it passed each time:
image

PR #288, opened 2024-10-22
added page and limit to finalities by keystones request

ClaytonNorthey92:
**Summary**
added page and limit to finalities by keystones request, defaulting limit to 100 if 0 and setting an upper limit of 100 per page

**Changes**
added page and limit to finalities by keystones request, defaulting limit to 100 if 0 and setting an upper limit of 100 per page


joshuasing (approved):

jcvernaleo (approved):
OK

PR #291, opened 2024-10-23
version: bump version to 0.5.1

joshuasing:
**Summary**
Update the version variables in the `version` package to version v0.5.1.
Ideally we should keep these variables set to the next planned version, so that development binaries always have the next planned version.

GoReleaser will always change these in release binaries, so this only affects binaries not built with GoReleaser.

**Changes**
- Change `version.Minor` to `5`.
- Change `version.Patch` to `1`.


ClaytonNorthey92 (approved):

marcopeereboom (approved):

jcvernaleo (approved):
OK

PR #293, opened 2024-10-24
make localnet tests more robust

ClaytonNorthey92:
**Summary**
 make localnet tests more robust

prior, localnet would wait a certain period of time, then check if we reached certain values of different resoruces in our local chain (ex. HEMI balance, pop txs mined).  this was problematic because there was no guarantee that we would reach those within the given amount of time for various reasons.

we could simply increase the time we wait, but then we would have to wait that amount of time per test.

**Changes**
now, set a max timeout of 10 minutes for the test + 2 minute warm up.  poll the network every 10 seconds for values reaching a reasonable threshold.  if we reach this threshold within the timeout, success! otherwise failure.

### DIFF
--- a/e2e/docker-compose.yml
+++ b/e2e/docker-compose.yml
@@ -15,6 +15,7 @@ services:
       - "-rpcallowip=0.0.0.0/0"
       - "-rpcbind=0.0.0.0:18443"
       - "-txindex=1"
+      - "-rpcworkqueue=400"
     ports:
       - "18443:18443"
       - "18444:18444"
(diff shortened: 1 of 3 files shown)


e2e/docker-compose.yml
```diff
@@ -15,6 +15,7 @@ services:
       - "-rpcallowip=0.0.0.0/0"
       - "-rpcbind=0.0.0.0:18443"
       - "-txindex=1"
+      - "-rpcworkqueue=400"
```
ClaytonNorthey92, line 18: this is likely not needed, but it doesn't hurt; I was seeing a few warnings of it being too small to handle requests so I bumped it up

joshuasing (approved):

jcvernaleo (approved):
OK

PR #294, opened 2024-10-25
Corrent container name in e2e tests

jcvernaleo:
The l1 runs geth, not op-geth.


ClaytonNorthey92 (approved):
this looks good to me, pending the localnet test here https://github.com/hemilabs/heminetwork/actions/runs/11520447247

the localnet test is a bit fragile at the moment due to some timing, so even if that fails I'll inspect the output and if it's a timing issue let's merge this


#293 should fix the fragility

joshuasing (approved):
This needs to be fixed in infra too :)

PR #297, opened 2024-10-30
Try to prevent header download stalls.

marcopeereboom:
**Summary**

**Changes**

### DIFF
--- a/service/tbc/crawler.go
+++ b/service/tbc/crawler.go
@@ -73,18 +73,18 @@ var (
 	}
 )
 
-type NotLinear string
+type NotLinearError string
 
-func (en NotLinear) Error() string {
-	return string(en)
+func (e NotLinearError) Error() string {
+	return string(e)
 }
 
-func (nl NotLinear) Is(target error) bool {
-	_, ok := target.(NotLinear)
+func (e NotLinearError) Is(target error) bool {
+	_, ok := target.(NotLinearError)
 	return ok
 }
 
-var ErrNotLinear = NotLinear("not linear")
+var ErrNotLinear = NotLinearError("not linear")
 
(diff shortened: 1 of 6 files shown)


service/tbc/crawler.go
```diff
@@ -72,6 +72,36 @@ var (
...
+type NotLinear string
+
+func (en NotLinear) Error() string {
+	return string(en)
+}
+
+func (nl NotLinear) Is(target error) bool {
+	_, ok := target.(NotLinear)
+	return ok
+}
```
joshuasing, line 84: I think this type should be called `NotLinearError`, as it implements `error`.
Additionally, the receiver name is different in each function - could it be changed to one name for both?

```suggestion
type NotLinearError string

func (e NotLinearError) Error() string {
	return string(e)
}

func (e NotLinearError) Is(target error) bool {
	_, ok := target.(NotLinear)
	return ok
}
```

service/tbc/tbc.go
```diff
@@ -1110,11 +854,47 @@ func (s *Server) syncBlocks(ctx context.Context) {
...
+							p, err)
+						return
+					}
+				}
+				s.pm.All(ctx, hp)
+
+			} else {
+				log.Debugf("handle all")
+				s.pm.All(ctx, s.headersPeer)
+			}
```
joshuasing, line 897: nit: I prefer avoiding `else`; so this is completely optional and there may be a reason you have avoided this which is not immediately obvious:
```suggestion
				return
			}

			log.Debugf("handle all")
			s.pm.All(ctx, s.headersPeer)
```

marcopeereboom: This is one is deliberate to make it obvious since the if clause is lengthy. I am keeping it.

joshuasing (approved):

PR #298, opened 2024-10-31
Fix major part of stalls and an index loop bug.

marcopeereboom:
**Summary**

**Changes**

### DIFF
--- a/service/tbc/peer.go
+++ b/service/tbc/peer.go
@@ -139,14 +139,24 @@ func (p *peer) handshake(ctx context.Context, conn net.Conn) error {
 
 	// 3. ask for v2 addresses, this has to be done before verack despite
 	// what the spec says.
-	if v.ProtocolVersion >= 70016 {
-		err = writeTimeout(defaultHandshakeTimeout, conn, wire.NewMsgSendAddrV2(), p.protocolVersion, p.network)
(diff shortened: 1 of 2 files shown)


service/tbc/peer.go
```diff
@@ -139,14 +139,24 @@ func (p *peer) handshake(ctx context.Context, conn net.Conn) error {
...
 		}
 	}
 
-	// 4. send verack
+	// 4. ask for headers.
+	if uint32(v.ProtocolVersion) >= wire.SendHeadersVersion {
+		err = writeTimeout(defaultHandshakeTimeout, conn, wire.NewMsgSendHeaders(),
+			p.protocolVersion, p.network)
+		if err != nil {
+			return fmt.Errorf("could not send addrv2: %w", err)
```
joshuasing, line 155: nit: the error message says "addrv2"
```suggestion
			return fmt.Errorf("could not send headers: %w", err)
```

joshuasing (approved):

service/tbc/peer.go
marcopeereboom: Welcome to bitcoin, it is correct :-)

ClaytonNorthey92 (approved):

PR #301, opened 2024-11-06
Make isCanonical call not fail on incorrect height.

marcopeereboom:
When the caller, for whatever reason, calls isCanonical with a block header that sits higher on the chain than best block header it cannot be canonical. This case did return false but also an error. Remove the error since the block simply is not canonical.

**Summary**

**Changes**


jcvernaleo (approved):
OK

ClaytonNorthey92 (approved):

PR #303, opened 2024-11-07
Add InsertBlock RPC

marcopeereboom:
When TBC is used inside other services we may need to insert blocks. The use case is an op-geth that has seen a block in the past yet tbc does not know it because it isn't canonical. Now that block may be needed for an index since it *was* canonical at some point in the past. It's a bit of a hack but there is no harm in having blocks sitting in the database that have been reorged out.

### DIFF
--- a/service/tbc/rpc.go
+++ b/service/tbc/rpc.go
@@ -16,6 +16,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/coder/websocket"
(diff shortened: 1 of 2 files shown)


joshuasing (approved):

service/tbc/rpc.go
```diff
@@ -545,7 +560,43 @@ func (s *Server) handleTxBroadcastRawRequest(ctx context.Context, req *tbcapi.Tx
...
 
-	return &tbcapi.TxBroadcastResponse{TxID: txid}, nil
+	return &tbcapi.TxBroadcastRawResponse{TxID: txid}, nil
+}
+
+func (s *Server) handleBlockInsertRequest(ctx context.Context, req *tbcapi.BlockInsertRequest) (any, error) {
+	log.Tracef("handleBlockInsertRequest")
+	defer log.Tracef("handleBlockInsertRequest exit")
+
+	_, err := s.db.BlockInsert(ctx, btcutil.NewBlock(req.Block))
```
joshuasing, line 570: As `req.Block` is a pointer, it is possible that the JSON payload could be `{"block": null}`.
Adding a nil-check above this would prevent possible panics; however I am not sure how `btcutil.NewBlock` will handle `nil`.

joshuasing: Thanks!

joshuasing (approved):

PR #320, opened 2024-11-19
Add External Header Mode to TBC + Misc Regular Node Updates

max-sanchez:
**Summary**
This PR adds an "External Header Mode" for TBC to run in, which causes P2P to be disabled and for TBC to maintain lightweight BTC consensus knowledge based purely on headers that are added/removed by upstream code that manages the External Header Mode TBC instance. Also added a few methods for normal-mode TBC for fetching specific data desired for hVM.

**Changes**
External Header Mode:
- Added the ability to spin up a TBC instance in ExternalHeaderMode which does not connect to P2P and receives header insertions/deletions from upstream code
- Ability to associate ExternalHeaderMode TBC instance state with an upstream state ID for upstream reconciliation
- Added the ability to remove headers from the chain's view of consensus and set a specific header as canonical afterwards, so that ExternalHeaderMode TBC can be rolled back to a previous consensus state
- Added the ability for TBC in ExternalHeaderMode to be initialized with a specific "effective" genesis block, which is assumed permanently canonical so that lightweight conesnsus can be tracked from a specific non-genesis height
- Updated methods that will not work with ExternalHeaderMode to return errors when called on ExternalHeaderMode TBC

Other Updates:
- Added ScriptHashAvailableToSpend method to check for the existence of a single outpoint in the UTXO table of a normal-mode TBC node
- Added FullBlockAvailable method to check whether a normal-mode TBC node has knowledge of a particular full block

### DIFF
--- a/database/tbcd/database.go
+++ b/database/tbcd/database.go
@@ -18,6 +18,7 @@ import (
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
+	"github.com/syndtr/goleveldb/leveldb"
 
 	"github.com/hemilabs/heminetwork/database"
 )
@@ -52,25 +53,23 @@ const (
 	RTChainFork    RemoveType = 3 // Removal walked canonical chain backwards far enough that another chain is now canonical
 )
 
-var (
-	rtStrings = map[RemoveType]string{
-		RTInvalid:      "invalid",
-		RTChainDescend: "canonical chain descend",
-		RTForkDescend:  "fork chain descend",
-		RTChainFork:    "canonical descend changed canonical",
-	}
-
-	DefaultUpstreamStateId = [32]byte{
-		0xFF, 0x00, 0xFF, 0x00, 0xFF, 0x00,
-		0x44, 0x45, 0x46, 0x41, 0x55, 0x4C, 0x54, 0x55, 0x50, 0x53, // DEFAULTUPS
-		0x54, 0x52, 0x45, 0x41, 0x4D, 0x53, 0x54, 0x41, 0x54, 0x45, // TREAMSTATE
-		0x00, 0xFF, 0x00, 0xFF, 0x00, 0xFF,
-	}
-)
+var rtStrings = map[RemoveType]string{
+	RTInvalid:      "invalid",
+	RTChainDescend: "canonical chain descend",
+	RTForkDescend:  "fork chain descend",
+	RTChainFork:    "canonical descend changed canonical",
+}
 
 func (rt RemoveType) String() string {
 	return rtStrings[rt]
 }
+
+type Transaction struct {
+	Batch       *leveldb.Batch
+	Transaction *leveldb.Transaction
+}
+
+type PostHook func(ctx context.Context, transactions map[string]Transaction) error
 
 type Database interface {
 	database.Database
@@ -85,14 +84,10 @@ type Database interface {
 	BlockHeaderByHash(ctx context.Context, hash *chainhash.Hash) (*BlockHeader, error)
 	BlockHeaderGenesisInsert(ctx context.Context, wbh *wire.BlockHeader, height uint64, diff *big.Int) error
 
-	// Upstream state id
-	UpstreamStateId(ctx context.Context) (*[32]byte, error)
-	SetUpstreamStateId(ctx context.Context, upstreamStateId *[32]byte) error
-
 	// Block headers
 	BlockHeadersByHeight(ctx context.Context, height uint64) ([]BlockHeader, error)
-	BlockHeadersInsert(ctx context.Context, bhs *wire.MsgHeaders, upstreamStateId *[32]byte) (InsertType, *BlockHeader, *BlockHeader, int, error)
-	BlockHeadersRemove(ctx context.Context, bhs *wire.MsgHeaders, tipAfterRemoval *wire.BlockHeader, upstreamStateId *[32]byte) (RemoveType, *BlockHeader, error)
+	BlockHeadersInsert(ctx context.Context, bhs *wire.MsgHeaders, postHook PostHook) (InsertType, *BlockHeader, *BlockHeader, int, error)
+	BlockHeadersRemove(ctx context.Context, bhs *wire.MsgHeaders, tipAfterRemoval *wire.BlockHeader, postHook PostHook) (RemoveType, *BlockHeader, error)
 
 	// Block
 	BlocksMissing(ctx context.Context, count int) ([]BlockIdentifier, error)
--- a/database/tbcd/database.go
+++ b/database/tbcd/database.go
@@ -42,6 +42,36 @@ func (it InsertType) String() string {
 	return itStrings[it]
 }
 
+// Canonical chain geometry changes resulting from header removal
+type RemoveType int
+
+const (
+	RTInvalid      RemoveType = 0 // Invalid removal for generic reason (ex: no headers to remove)
+	RTChainDescend RemoveType = 1 // Removal walked the canonical chain backwards, but existing chain is still canonical
+	RTForkDescend  RemoveType = 2 // Removal walked a non-canonical chain backwards, no change to canonical chain remaining canonical
+	RTChainFork    RemoveType = 3 // Removal walked canonical chain backwards far enough that another chain is now canonical
+)
+
+var (
+	rtStrings = map[RemoveType]string{
+		RTInvalid:      "invalid",
+		RTChainDescend: "canonical chain descend",
+		RTForkDescend:  "fork chain descend",
+		RTChainFork:    "canonical descend changed canonical",
+	}
+
+	DefaultUpstreamStateId = [32]byte{
+		0xFF, 0x00, 0xFF, 0x00, 0xFF, 0x00,
+		0x44, 0x45, 0x46, 0x41, 0x55, 0x4C, 0x54, 0x55, 0x50, 0x53, // DEFAULTUPS
+		0x54, 0x52, 0x45, 0x41, 0x4D, 0x53, 0x54, 0x41, 0x54, 0x45, // TREAMSTATE
+		0x00, 0xFF, 0x00, 0xFF, 0x00, 0xFF,
+	}
+)
+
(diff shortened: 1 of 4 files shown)


database/tbcd/database.go
```diff
@@ -42,6 +42,36 @@ func (it InsertType) String() string {
...
+	}
+
+	DefaultUpstreamStateId = [32]byte{
+		0xFF, 0x00, 0xFF, 0x00, 0xFF, 0x00,
+		0x44, 0x45, 0x46, 0x41, 0x55, 0x4C, 0x54, 0x55, 0x50, 0x53, // DEFAULTUPS
+		0x54, 0x52, 0x45, 0x41, 0x4D, 0x53, 0x54, 0x41, 0x54, 0x45, // TREAMSTATE
+		0x00, 0xFF, 0x00, 0xFF, 0x00, 0xFF,
+	}
+)
+
```
marcopeereboom, line 70: This probably doesn't belong here.

database/tbcd/database.go
```diff
@@ -42,6 +42,36 @@ func (it InsertType) String() string {
...
+// Canonical chain geometry changes resulting from header removal
+type RemoveType int
+
+const (
+	RTInvalid      RemoveType = 0 // Invalid removal for generic reason (ex: no headers to remove)
+	RTChainDescend RemoveType = 1 // Removal walked the canonical chain backwards, but existing chain is still canonical
+	RTForkDescend  RemoveType = 2 // Removal walked a non-canonical chain backwards, no change to canonical chain remaining canonical
+	RTChainFork    RemoveType = 3 // Removal walked canonical chain backwards far enough that another chain is now canonical
+)
+
```
marcopeereboom, line 54: Why do we need to remove headers? So long the headers are inserted consecutively (and if they don't bad things happen!!!) the heaviest will always win. I am not necessarily opposed but this seems like a lot of extra code that may complicate things quite a bit.

database/tbcd/database.go
```diff
@@ -53,11 +83,16 @@ type Database interface {
 	// Block header
 	BlockHeaderBest(ctx context.Context) (*BlockHeader, error) // return canonical
 	BlockHeaderByHash(ctx context.Context, hash *chainhash.Hash) (*BlockHeader, error)
-	BlockHeaderGenesisInsert(ctx context.Context, wbh *wire.BlockHeader) error
+	BlockHeaderGenesisInsert(ctx context.Context, wbh *wire.BlockHeader, height uint64, diff *big.Int) error
+
+	// Upstream state id
+	UpstreamStateId(ctx context.Context) (*[32]byte, error)
+	SetUpstreamStateId(ctx context.Context, upstreamStateId *[32]byte) error
```
marcopeereboom, line 91: From the looks of it this should use: `MetadataGet` and `MetadataPut` instead of inserting this in the interface.

database/tbcd/level/level.go
```diff
@@ -377,14 +443,33 @@ func (l *ldb) BlockHeaderGenesisInsert(ctx context.Context, wbh *wire.BlockHeade
...
+	// instead (used for external header mode). In the event that an effective genesis block
+	// is supplied but the cumulative difficulty is not set, the difficulty of that
+	// effective genesis block (set above) will be retained, which has no effect on local
+	// consensus determination but will not permit direct comparison of cumulative difficulty
+	// against the full chain, only relative comparisons between cumulative difficulties of
+	// blocks on top of the same effective genesis block are guaranteed valid.
+	if diff != nil && diff.Cmp(big.NewInt(0)) > 0 {
+		cdiff = diff
+	}
+	ebh := encodeBlockHeader(height, h2b(wbh), cdiff)
```
marcopeereboom, line 468: This should be removed. If we force sending in genesis then crash if diff is nil.

database/tbcd/level/level.go
```diff
@@ -377,14 +443,33 @@ func (l *ldb) BlockHeaderGenesisInsert(ctx context.Context, wbh *wire.BlockHeade
...
 	bhBatch.Put(bhash[:], ebh[:])
 
 	bhBatch.Put([]byte(bhsCanonicalTipKey), ebh[:])
+	bhBatch.Put([]byte(upstreamStateIdKey), tbcd.DefaultUpstreamStateId[:])
```
marcopeereboom, line 473: Are we sure this does not belong in metadata?

database/tbcd/level/level.go
```diff
@@ -419,13 +504,367 @@ func (l *ldb) BlockHeaderGenesisInsert(ctx context.Context, wbh *wire.BlockHeade
...
+// indicates incorrect upstream behavior.
+//
+// If any of the above requirements are not true, this function will return
+// an error. If this function returns an error, no changes have been made to
+// the underlying database state as all validity checks are done before db
+// modifications are applied.
+//
+// If an upstreamCursor is provided, it is updated atomically in the database
+// along with the state transition of removing the block headers.
+func (l *ldb) BlockHeadersRemove(ctx context.Context, bhs *wire.MsgHeaders, tipAfterRemoval *wire.BlockHeader, upstreamStateId *[32]byte) (tbcd.RemoveType, *tbcd.BlockHeader, error) {
```
marcopeereboom, line 609: This is very complicated and seems to be doing a bit too much. Not sure how to fix it yet but jotting down initial findings.

The big question is: do we have to remove blockheaders? and why?

database/tbcd/level/level.go
```diff
@@ -419,13 +504,367 @@ func (l *ldb) BlockHeaderGenesisInsert(ctx context.Context, wbh *wire.BlockHeade
...
+
 // BlockHeadersInsert decodes and inserts the passed blockheaders into the
 // database. Additionally it updates the hight/hash and missing blocks table as
 // well.  On return it informs the caller about potential forking situations
 // and always returns the canonical and last inserted blockheader, which may be
 // the same.
 // This call uses the database to prevent reentrancy.
-func (l *ldb) BlockHeadersInsert(ctx context.Context, bhs *wire.MsgHeaders) (tbcd.InsertType, *tbcd.BlockHeader, *tbcd.BlockHeader, int, error) {
+func (l *ldb) BlockHeadersInsert(ctx context.Context, bhs *wire.MsgHeaders, upstreamStateId *[32]byte) (tbcd.InsertType, *tbcd.BlockHeader, *tbcd.BlockHeader, int, error) {
 	log.Tracef("BlockHeadersInsert")
```
marcopeereboom, line 868: Same here, do we need upstream state id in here?

service/tbc/tbc.go
```diff
@@ -87,17 +88,26 @@ type Config struct {
...
 	PprofListenAddress      string
 	Seeds                   []string
+
+	// Fields used for running TBC in External Header Mode, where P2P is disabled
+	// and TBC is used to determine consensus based on headers fed from external
+	// code that manages the TBC node.
+	ExternalHeaderMode      bool              // Whether Header-Only Mode is enabled
+	EffectiveGenesisBlock   *wire.BlockHeader // The header to use as the first block in TBC's consensus view
+	GenesisHeightOffset     uint64            // The height of the effective genesis block
+	GenesisDifficultyOffset big.Int           // The cumulative difficulty of the effective genesis block
```
marcopeereboom, line 98: This is theoretical, but should we create a fake blockchain params here? so that we don't have to if/then/else all over the place.

service/tbc/tbc.go
```diff
@@ -158,14 +168,22 @@ func NewServer(cfg *Config) (*Server, error) {
 	if cfg == nil {
```
marcopeereboom, line 168: Should we create a headers only peer manager?

service/tbc/tbc.go
```diff
@@ -1677,6 +1878,58 @@ func (s *Server) FeesAtHeight(ctx context.Context, height, count int64) (uint64,
...
+		}
+	}
+
+	if block != nil {
+		// Were able to get full block from database
+		return true, nil
+	} else {
+		return false, errors.New("fetching block did not return error but block is nil")
+	}
+}
```
marcopeereboom, line 1903: This should return an actionable error instead of mixing flags and bools creating 4 actionable cases.

service/tbc/tbc.go
```diff
@@ -1825,7 +2086,10 @@ func (s *Server) Run(pctx context.Context) error {
 			return fmt.Errorf("block header best: %w", err)
 		}
 
-		if err = s.insertGenesis(ctx); err != nil {
+		// This Run function is only called in regular (not external header) mode, so this is true genesis block @ 0
+		// and we pass in a nil difficulty so it calculates the starting difficulty as the genesis block's own local
+		// difficulty.
+		if err = s.insertGenesis(ctx, 0, nil); err != nil {
 			return fmt.Errorf("insert genesis: %w", err)
```
marcopeereboom, line 2093: Pass new(big.Int) instead of nil. Crashing is great.

service/tbc/tbc.go
```diff
@@ -2031,3 +2295,79 @@ func (s *Server) Run(pctx context.Context) error {
...
+		if len(gb) > 1 {
+			// Impossible to have more than one block at the genesis height
+			return fmt.Errorf("invalid state, have %d effective genesis blocks", len(gb))
+		}
+		gh := genesis.BlockHash()
+		if !bytes.Equal(gb[0].Hash[:], gh[:]) {
+			return fmt.Errorf("effective genesis block hash mismatch, db has %x but genesis should be %x", gb[0].Hash, gh)
+		}
+	}
+	log.Infof("TBC set up in External Header Mode, effectiveGenesis=%x, tip=%x", genesis.BlockHash(), bhb.Hash)
```
marcopeereboom, line 2354: Don't love this message.

service/tbc/tbc.go
```diff
@@ -2031,3 +2295,79 @@ func (s *Server) Run(pctx context.Context) error {
...
+		return errors.New("ExternalHeaderTearDown called but external header mode is not enabled in config")
+	}
+
+	err := s.DBClose()
+	if err != nil {
+		log.Errorf("db close: %v", err)
+		return err
+	}
+	return nil
+}
```
marcopeereboom, line 2373: This should somehow be automatically handled. Needs some thought.

marcopeereboom (approved):

joshuasing:
Superseded by https://github.com/hemilabs/heminetwork/pull/324

PR #323, opened 2024-11-21
Add batched medata put/get and wrap them in transactions

marcopeereboom:
**Summary**

**Changes**

### DIFF
--- a/database/tbcd/level/level.go
+++ b/database/tbcd/level/level.go
@@ -30,7 +30,11 @@ import (
 )
 
 // Locking order:
+//	The theory here is to order these in somewhat "usage" order and to keep
+//	them from interacting too much. Metadata goes first because it is
+//	almost never used but may interact wildly.
 //
+//	Metadata	XXX debate this
 //	BlockHeaders
 //	BlocksMissing
 //	HeightHash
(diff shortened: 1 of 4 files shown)


database/tbcd/level/level_test.go
```diff
@@ -0,0 +1,114 @@
...
+	defer func() {
+		err := db.Close()
+		if err != nil {
+			t.Fatal(err)
+		}
+	}()
+
+	x := 255
+	rows := make([]tbcd.Row, x+1)
+	for i := 0; i <= x; i++ {
```
joshuasing, line 38:
```suggestion
	for i := range x {
```

database/tbcd/level/level.go
```diff
@@ -187,25 +191,121 @@ func (l *ldb) startTransaction(db string) (*leveldb.Transaction, commitFunc, dis
...
+					string(keys[k]), err))
+			}
+			return nil, fmt.Errorf("%s: %w", string(keys[k]), err)
+		}
+		if errors.Is(err, leveldb.ErrNotFound) {
+			// overload err
+			err = database.NotFoundError(err.Error())
+		}
+		rows[k] = tbcd.Row{Key: keys[k], Value: value, Error: err}
+	}
```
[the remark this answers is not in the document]
marcopeereboom: There is a debate to be had over copying or not copying the value. In this case this ends up with one less copy of all pointers and hence i picked that option.

database/tbcd/level/level_test.go
marcopeereboom: While I typically agree in this case we need to show the old ugly way because x+1; this is also a test and we want to be explicit and fail if assumptions are wrong.

joshuasing (approved):

jcvernaleo (approved):
OK

PR #325, opened 2024-11-21
tbc: fix err logging in sync indexers

CodeDragonVN:
**Summary**

Fix incorrect error handling in crawler synchronization methods.

**Changes**

- Changed `if err == nil` to `if err != nil` in `SyncIndexersToHash` to correctly return errors.
- Made the same change in `syncIndexersToBest`.


jcvernaleo (approved):
OK

joshuasing (approved):
Thank you!

marcopeereboom (changes requested):
This is incorrect. The bug is real but the intent was to ,move the log into the if.

marcopeereboom (approved):

PR #330, opened 2024-11-26
correct batcher and proposer in localnet

ClaytonNorthey92:
**Summary**
I was incorrect with my understanding of op-proposer and op-batcher, there is a single instance of each that connects to each op-geth and op-node

**Changes**
fix op-proposer and op-batcher config


joshuasing (approved):

jcvernaleo (approved):
OK

PR #355, opened 2025-01-06
hemi/electrs: improve response handling

SashaMalysehko:
**Summary**
- Add context cancellation checks in call method
- Enhance handling of incoming requests with ID=0
- Improve debug logging for ignored requests

**Changes**
Added a context check in the call method before executing operations to ensure correct cancellation of operations.
Reworked the readResponse function, adding an infinite loop to correctly process incoming requests with ID=0.
Improved logging of ignored requests to make debugging easier.
These changes make the code more reliable and resilient to errors, and improve debugging when problems with request processing occur.

### DIFF
--- a/hemi/electrs/conn.go
+++ b/hemi/electrs/conn.go
@@ -62,6 +62,13 @@ func (c *clientConn) call(ctx context.Context, method string, params, result any
 	log.Tracef("call")
 	defer log.Tracef("call exit")
 
+	// Check context before acquiring lock
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
+	}
+
 	c.mx.Lock()
 	defer c.mx.Unlock()
 
@@ -86,6 +93,14 @@ func (c *clientConn) call(ctx context.Context, method string, params, result any
 	if err != nil {
 		return fmt.Errorf("create request: %w", err)
 	}
+
+	// Check context again before writing
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
+	}
+
 	if err = writeRequest(ctx, c.conn, req); err != nil {
 		return fmt.Errorf("write request: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


hemi/electrs/conn.go
```diff
@@ -62,6 +62,13 @@ func (c *clientConn) call(ctx context.Context, method string, params, result any
 	log.Tracef("call")
 	defer log.Tracef("call exit")
 
+	// Check context before acquiring lock
```
marcopeereboom, line 65: This is the first thing that happens in the underlying code and thus not needed.

hemi/electrs/conn.go
```diff
@@ -86,6 +93,14 @@ func (c *clientConn) call(ctx context.Context, method string, params, result any
...
 		return fmt.Errorf("create request: %w", err)
 	}
+
+	// Check context again before writing
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
+	}
+
```
marcopeereboom, line 103: This is the first thing that happens in the underlying code and thus not needed.

hemi/electrs/conn.go
```diff
@@ -131,37 +146,38 @@ func readResponse(ctx context.Context, r io.Reader, reqID uint64) (*JSONRPCRespo
...
-	b, err := reader.ReadBytes('\n')
-	if err != nil {
-		return nil, err
-	}
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
```
marcopeereboom, line 154: If the context was closed we would not have made it here.

hemi/electrs/conn.go
```diff
@@ -131,37 +146,38 @@ func readResponse(ctx context.Context, r io.Reader, reqID uint64) (*JSONRPCRespo
...
+		if err = json.Unmarshal(b, &res); err != nil {
+			return nil, fmt.Errorf("unmarshal response: %w", err)
+		}
+		if res.Error != "" {
+			return nil, RPCError(res.Error)
+		}
+
+		if res.ID != reqID {
+			if res.ID == 0 {
+				// Log the ignored request for debugging purposes
```
marcopeereboom, line 172: I don't even know what this means. Is this an unsolicited command?

marcopeereboom (changes requested):
My inclination is to close this PR but I'll let @joshuasing have a go at it.

PR #357, opened 2025-01-06
add wait groups to tbc fork tests

ClaytonNorthey92:
**Summary**
During tests, the test tbc server would be shut down (sometimes) after the `t.Tempdir()` was deleted.  When leveldb attempts to do something with files in that directory for the server's sake, it may show an error:
```
2025-01-06T16:47:13.1276786Z 2025-01-06 16:47:01 INFO tbc tbc.go:2310 RPC server shutdown cleanly
2025-01-06T16:47:13.1277485Z 2025-01-06 16:47:01 ERROR level level.go:85 close metadata: open /tmp/TestIndexNoFork742784104/001/localnet/metadata: no such file or directory
```

**Changes**
add wait groups to tbc fork tests to prevent `t.Tempdir()` from being deleted prior to tbc servers being torn down cleanly

### DIFF
--- a/service/tbc/tbcfork_test.go
+++ b/service/tbc/tbcfork_test.go
@@ -853,9 +853,11 @@ func errorIsOneOf(err error, errs []error) bool {
 }
 
 func TestFork(t *testing.T) {
+	wg := sync.WaitGroup{}
 	ctx, cancel := context.WithCancel(context.Background())
 	defer func() {
 		cancel()
+		wg.Wait()
 	}()
 
 	n, err := newFakeNode(t, "18444") // TODO: should use random free port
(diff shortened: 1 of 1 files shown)


service/tbc/tbcfork_test.go
```diff
@@ -853,9 +853,11 @@ func errorIsOneOf(err error, errs []error) bool {
 }
 
 func TestFork(t *testing.T) {
+	wg := sync.WaitGroup{}
 	ctx, cancel := context.WithCancel(context.Background())
 	defer func() {
 		cancel()
+		wg.Wait()
```
marcopeereboom, line 860: Yikes I am not sure I love a wait in a deferred closure. I think we need to revisit entry and exit of this entire function. Pretty much by definition the wait should be at the very bottom of the test followed by the delete of the temp dir. Are you sure this is the root cause or is this a symptom that this works around?

marcopeereboom (changes requested):

PR #360, opened 2025-01-08
tbcd: move block header cache from items to size

marcopeereboom:
**Summary**
We really should use size as the metric to determine cache size. This flips that for blockheaders.

**Changes**
This fixes a major bug in headers only mode where headers were not purged from the cache. This may have caused problems we had not diagnosed.

### DIFF
--- a/database/tbcd/database.go
+++ b/database/tbcd/database.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2024 Hemi Labs, Inc.
+// Copyright (c) 2024-2025 Hemi Labs, Inc.
 // Use of this source code is governed by the MIT License,
 // which can be found in the LICENSE file.
 
@@ -103,6 +103,7 @@ type Database interface {
 	BlockInsert(ctx context.Context, b *btcutil.Block) (int64, error)
 	// BlocksInsert(ctx context.Context, bs []*btcutil.Block) (int64, error)
 	BlockByHash(ctx context.Context, hash *chainhash.Hash) (*btcutil.Block, error)
+	BlockCacheStats() CacheStats
 
 	// Transactions
 	BlockUtxoUpdate(ctx context.Context, direction int, utxos map[Outpoint]CacheOutput) error
@@ -412,4 +413,13 @@ func TxIdBlockHashFromTxKey(txKey TxKey) (*chainhash.Hash, *chainhash.Hash, erro
 		return nil, nil, fmt.Errorf("invalid block hash: %w", err)
 	}
 	return txId, blockHash, nil
+}
+
+// Cache
+type CacheStats struct {
+	Hits   int
+	Misses int
+	Purges int
+	Size   int
+	Items  int
 }
--- a/database/tbcd/database.go
+++ b/database/tbcd/database.go
@@ -91,6 +91,7 @@ type Database interface {
 	BlockHeaderBest(ctx context.Context) (*BlockHeader, error) // return canonical
 	BlockHeaderByHash(ctx context.Context, hash *chainhash.Hash) (*BlockHeader, error)
 	BlockHeaderGenesisInsert(ctx context.Context, wbh *wire.BlockHeader, height uint64, diff *big.Int) error
+	BlockHeaderCacheStats() CacheStats
 
 	// Block headers
 	BlockHeadersByHeight(ctx context.Context, height uint64) ([]BlockHeader, error)
--- a/database/tbcd/level/blockcache.go
+++ b/database/tbcd/level/blockcache.go
@@ -1,0 +1,78 @@
+// Copyright (c) 2025 Hemi Labs, Inc.
+// Use of this source code is governed by the MIT License,
+// which can be found in the LICENSE file.
+
+package level
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+)
+
+const blockSize = 1677721 // ~1.6MB rough size of a mainnet block as of Jan 2025
+
+type lowIQLRU struct {
+	mtx sync.RWMutex
+
+	size int // this is the approximate max size
+
+	m         map[chainhash.Hash][]byte
+	totalSize int
+}
+
+func (l *lowIQLRU) Put(v *btcutil.Block) {
+	l.mtx.Lock()
+	defer l.mtx.Unlock()
+
+	hash := v.Hash()
+	if _, ok := l.m[*hash]; ok {
+		return
+	}
+
+	block, err := v.Bytes()
+	if err != nil {
+		panic(err)
+		// XXX don't cache but panic for now for diagnostic
+	}
+
+	// XXX add eviction here
+	if l.totalSize+len(block) >= l.size {
+		panic("evict")
+	}
+
+	l.m[*hash] = block
+	l.totalSize += len(block)
+}
+
+func (l *lowIQLRU) Get(k *chainhash.Hash) (*btcutil.Block, bool) {
+	l.mtx.RLock()
+	defer l.mtx.RUnlock()
+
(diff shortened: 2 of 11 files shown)


database/tbcd/level/level.go
```diff
@@ -108,16 +110,26 @@ func headerHash(header []byte) *chainhash.Hash {
 }
 
 type Config struct {
-	Home             string // home directory
-	BlockCache       int    // number of blocks to cache
-	BlockheaderCache int    // number of blocks headers to cache
+	Home                 string // home directory
+	BlockCache           int    // number of blocks to cache
+	BlockheaderCacheSize string // size of block header cache
+	blockheaderCacheSize int    // paser size of block header cache,
```
joshuasing, line 116: nit: block header is two words, as such it could make sense to capitalise the H:
```suggestion
	BlockHeaderCacheSize string // size of block header cache
	blockHeaderCacheSize int    // paser size of block header cache,
```

database/tbcd/level/level.go
```diff
@@ -146,10 +158,14 @@ func New(ctx context.Context, cfg *Config) (*ldb, error) {
...
 		log.Infof("block cache: DISABLED")
 	}
-	if cfg.BlockheaderCache > 0 {
-		l.headerCache = lowIQMapNew(cfg.BlockheaderCache)
-
-		log.Infof("blockheader cache: %v", cfg.BlockheaderCache)
+	if cfg.blockheaderCacheSize > 0 {
+		l.headerCache, err = lowIQMapNewSize(cfg.blockheaderCacheSize)
+		if err != nil {
+			panic(spew.Sdump(cfg))
```
joshuasing, line 164: Was this added for debugging purposes? If so, can it be removed?
```suggestion
```

database/tbcd/level/headercache.go
```diff
@@ -47,9 +50,31 @@ func (l *lowIQMap) Get(k *chainhash.Hash) (*tbcd.BlockHeader, bool) {
...
+func (l *lowIQMap) Purge(k *chainhash.Hash) {
+	l.mtx.RLock()
+	defer l.mtx.RUnlock()
+
+	delete(l.m, *k)
+}
+
+func lowIQMapNewCount(count int) (*lowIQMap, error) {
+	if count <= 0 {
+		return nil, fmt.Errorf("invalid count %v", count)
```
joshuasing, line 62:
```suggestion
		return nil, fmt.Errorf("invalid count: %v", count)
```

database/tbcd/level/headercache.go
```diff
@@ -47,9 +50,31 @@ func (l *lowIQMap) Get(k *chainhash.Hash) (*tbcd.BlockHeader, bool) {
...
+	}, nil
+}
+
+// lowIQMapNewSize does a bit of math to estimate the number of cache items.
+// Since it is an estimate it will overflow if Difficulty becomes bigger than
+// 64 bits. This is not an issue since 100MB caches all of mainnet in Jan 2025
+// (~819200 items).
+func lowIQMapNewSize(size int) (*lowIQMap, error) {
+	if size <= 0 {
+		return nil, fmt.Errorf("invalid size %v", size)
```
joshuasing, line 76:
```suggestion
		return nil, fmt.Errorf("invalid size: %v", size)
```

database/tbcd/level/headercache.go
```diff
@@ -47,9 +50,31 @@ func (l *lowIQMap) Get(k *chainhash.Hash) (*tbcd.BlockHeader, bool) {
 	return bh, ok
 }
 
-func lowIQMapNew(count int) *lowIQMap {
+func (l *lowIQMap) Purge(k *chainhash.Hash) {
+	l.mtx.RLock()
+	defer l.mtx.RUnlock()
```
joshuasing, line 55: I think this should call `Lock`/`Unlock`, as it is writing to the map (delete), not reading.
```suggestion
	l.mtx.Lock()
	defer l.mtx.Unlock()
```

joshuasing (changes requested):

database/tbcd/level/level.go
marcopeereboom: Nein. I have resisted that and will continue to resist that :-)

database/tbcd/level/level.go
marcopeereboom: Yeah, just wanted early eyes on this before moving forward.

database/tbcd/level/headercache.go
marcopeereboom: derp, yep.

database/tbcd/level/headercache.go
```diff
@@ -47,9 +50,40 @@ func (l *lowIQMap) Get(k *chainhash.Hash) (*tbcd.BlockHeader, bool) {
 	return bh, ok
 }
 
-func lowIQMapNew(count int) *lowIQMap {
+func (l *lowIQMap) Purge(k *chainhash.Hash) {
```
ClaytonNorthey92, line 53: if this function is not used, can we delete it?

database/tbcd/level/headercache.go
```diff
@@ -47,9 +50,40 @@ func (l *lowIQMap) Get(k *chainhash.Hash) (*tbcd.BlockHeader, bool) {
...
+
+	delete(l.m, *k)
+}
+
+func (l *lowIQMap) PurgeBatch(ks []*chainhash.Hash) {
+	l.mtx.Lock()
+	defer l.mtx.Unlock()
+
+	for v := range ks {
+		delete(l.m, *ks[v])
```
ClaytonNorthey92, line 65: Although I don't think that it's possible at this point for this pointer to be `nil`, I am going to ask that we make sure that this pointer is not `nil` before de-referencing it (perhaps log a warning as well).

Also, does the input need to be a slice of pointers? or could we just do `[]chainhash.Hash`?

ClaytonNorthey92 (changes requested):

ClaytonNorthey92 (approved):

database/tbcd/level/headercache.go
marcopeereboom: We need to punish the caller if the pointer is nil. As it is written now it cannot be nil. Also, these operations are already expensive so not blowing the pre-fetcher is also nice :-)

database/tbcd/level/headercache.go
marcopeereboom: removed

database/tbcd/database.go
```diff
@@ -413,3 +415,12 @@ func TxIdBlockHashFromTxKey(txKey TxKey) (*chainhash.Hash, *chainhash.Hash, erro
 	}
 	return txId, blockHash, nil
 }
+
+// Cache
+type CacheStats struct {
+	Hits   int
+	Misses int
+	Purges int
```
joshuasing, line 423: nit: would "Evictions" make more sense here?
```suggestion
	Evictions int
```

database/tbcd/level/blockcache.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (l *lowIQLRU) Stats() tbcd.CacheStats {
+	l.mtx.Lock()
+	defer l.mtx.Unlock()
+	l.c.Items = len(l.m)
+	return l.c
+}
+
+func lowIQLRUNewSize(size int) (*lowIQLRU, error) {
```
joshuasing, line 103: We seem to have a weird mix of `newABC` and `abcNew` - should we choose one and stick to it more consistently?
I think `[nN]ewABC` is a lot more common, and used throughout Go stdlib.

joshuasing (commented):
Would the term "evictions" make more sense for cache stats, then "purges"?
"purge" sounds like all items are removed, vs just one item being evicted 🤔

database/tbcd/level/blockcache.go
marcopeereboom: yep, fixed

PR #369, opened 2025-01-14
change paradigm from checking for missing childrent to keeping track of tip via electrs

### LINKED ISSUE #370: BFG incorrect fork handling
### Confirmation

### Describe the bug

BFG, when there are multiple long forks, doesn't handle these well.  It looks for "missing children" when it could be smarter about keeping track of the tip of the best chain and walking back from that.

### Expected behaviour

BFG should keep track of the best chain's tip, and ensure that, when walked backward, it can fill in missing blocks

### Environment

Version: 
Operating System: 
Architecture: 

### Additional Information

_No response_

ClaytonNorthey92:
**Summary**

prior, to solve for forks we would try to fill in "missing children".  the way we (I) had written it, there is no way to guarantee that we will get to the tip; if there were siblings at that height then we would not check for missing blocks.

a better way to do it it to check for the tip of the btc chain to change; when that changes, update the btc chain until we get to a known block

**Changes**
rather than walking up the btc chain, monitor the tip and walk down until we come across a known block

fixes #370


ClaytonNorthey92:
this is failing a localnet test, I will dive deeper in the AM

ClaytonNorthey92:
ok localnet test passes https://github.com/hemilabs/heminetwork/actions/runs/12761873612/job/35569512517

service/bfg/bfg.go
```diff
@@ -297,32 +297,58 @@ func NewServer(cfg *Config) (*Server, error) {
...
 
-			log.Infof("received %d heights with no children, will re-check", len(heights))
-			for _, height := range heights {
-				log.Infof("reprocessing block at height %d", height)
-				if err := s.processBitcoinBlock(ctx, height); err != nil {
+			s.updateBtcHeightCache(height)
+
+			log.Tracef("checking for missing blocks, walking from tip %d to %d or first known block from tip", height, s.cfg.BTCStartHeight)
+
+			for height > s.cfg.BTCStartHeight {
```
ClaytonNorthey92, line 337: `>=`

PR #372, opened 2025-01-14
twc: added skeleton for twc

AL-CT:
**Summary**
Created a basic skeleton for Tiny Wallet, which should later implement some of the logic extracted from bfg:

- twcapi
- twc
- twcd
- support for pinging twc rpc endpoint using hemictl


marcopeereboom:
We'll revive this branch if needed. Now tracked in #396

PR #374, opened 2025-01-14
Depend on btc client (electrs) for canonical chain

ClaytonNorthey92:
**Summary**
 change chain-tracking paradigm

**Changes**
* no longer calculate canonical chain in bfg
* start from tip, walking backwards
* upon startup walk backwards processing blocks to the start height, note that we only process txs in blocks that we have not seen
* every 5 seconds, walk back from current tip, replacing blocks that are not in the canonical chain until we come accross one that we have seen
* height is now unique in the btc blocks table
* if a block is deleted, ensure we delete pop_basis rows in the db
* move inserts to the end of btc block processing as to not be affected by electrs timeouts (they _should_ also be atomic but they never were, that's for another day)

### DIFF
--- a/database/bfgd/postgres/postgres.go
+++ b/database/bfgd/postgres/postgres.go
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	bfgdVersion = 14
+	bfgdVersion = 15
 
 	logLevel = "INFO"
 	verbose  = false
@@ -58,13 +58,6 @@ func New(ctx context.Context, uri string) (*pgdb, error) {
 	p := &pgdb{
 		Database: pg,
 		db:       pg.DB(),
-	}
-
-	// first, refresh the materialized view so it can be used in case it was
-	// never refreshed before this point
-	err = p.refreshBTCBlocksCanonical(ctx)
-	if err != nil {
-		return nil, err
 	}
 
 	return p, nil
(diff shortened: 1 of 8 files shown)


database/bfgd/postgres/postgres.go
```diff
@@ -242,6 +235,31 @@ func (p *pgdb) L2KeystonesMostRecentN(ctx context.Context, n uint32, page uint32
 	return ks, nil
 }
 
+func (p *pgdb) BtcBlockReplace(ctx context.Context, btcBlock *bfgd.BtcBlock) (int64, error) {
+	log.Tracef("BtcBlockReplace")
+	defer log.Tracef("BtcBlockReplace exit")
+
+	insertSql := `
```
ClaytonNorthey92, line 242: add a comment as to what this is doing (how `height` is now unique)

database/bfgd/scripts/0015.sql
```diff
@@ -0,0 +1,26 @@
...
+DROP TRIGGER btc_blocks_canonical_refresh_btc_blocks ON btc_blocks;
+
+DROP FUNCTION refresh_btc_blocks_can();
+
+-- no longer have BFG responsible for the canonical chain
+DROP MATERIALIZED VIEW btc_blocks_can;
+
+
+-- we now trust electrs/bitcoind to maintain the best chain, ensure these
+-- three columns are unique together
```
ClaytonNorthey92, line 18: update this comment it's not entirely correct

marcopeereboom (approved):
I think this is just a bandaid as well. But let's get this in and march towards a new and fresh bfg.

AL-CT (approved):

PR #378, opened 2025-01-16
tbcd: Fix startup/shutdown messages

marcopeereboom:
**Summary**
Startup and shutdown messages had become inconsistent over time. Clean that up.

### DIFF
--- a/service/tbc/peermanager.go
+++ b/service/tbc/peermanager.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2024 Hemi Labs, Inc.
+// Copyright (c) 2024-2025 Hemi Labs, Inc.
 // Use of this source code is governed by the MIT License,
 // which can be found in the LICENSE file.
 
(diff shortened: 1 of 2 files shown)


service/tbc/peermanager.go
```diff
@@ -436,7 +436,10 @@ func (pm *PeerManager) Run(ctx context.Context) error {
 			go pm.connectSlot(ctx, p)
 
 		case <-ctx.Done():
-			log.Infof("exit")
+			err := ctx.Err()
+			if errors.Is(err, context.Canceled) {
+				return nil
+			}
 			return ctx.Err()
```
joshuasing, line 443: `err` is defined and only used once in the if statement. Maybe it could be used as the return value as well?
```suggestion
			err := ctx.Err()
			if errors.Is(err, context.Canceled) {
				return nil
			}
			return err
```
or
```suggestion
			if errors.Is(ctx.Err(), context.Canceled) {
				return nil
			}
			return ctx.Err()
```

jcvernaleo (approved):
OK

joshuasing (approved):

PR #380, opened 2025-01-16
tbcd, bfgd: add various hemi specific indexes

marcopeereboom:
**Summary**
Hemi needs to keep track of L2 keystones in order to provide callers the bitcoin view of the hemi network.

**Changes**
This PR adds L2 keystone indexing and hides this behind a configuration flag.

### DIFF
--- a/database/tbcd/level/level.go
+++ b/database/tbcd/level/level.go
@@ -1769,7 +1769,7 @@ func (l *ldb) BlockKeystoneUpdate(ctx context.Context, direction int, keystones 
 				// Only store unknown keystones
 				continue
 			}
-			var value [chainhash.HashSize + hemi.L2KeystoneAbrevSize + 7]byte
+			var value [chainhash.HashSize + hemi.L2KeystoneAbrevSize]byte
 			copy(value[0:32], v.BlockHash[:])
 			copy(value[32:], v.AbbreviatedKeystone)
 			kssBatch.Put(k[:], value[:])
--- a/database/tbcd/level/level.go
+++ b/database/tbcd/level/level.go
@@ -1727,8 +1727,18 @@ func (l *ldb) BlockKeystoneUpdate(ctx context.Context, direction int, keystones 
 	for k, v := range keystones {
 		switch direction {
 		case -1:
-			kssBatch.Delete(k[:])
-			panic("not yet")
+			foundKeystone, err := kssTx.Get(k[:], nil)
+			if err != nil {
+				continue
+			}
+			fbh, err := chainhash.NewHash(foundKeystone[0:32])
+			if err != nil {
+				return err // XXX really can't happen
+			}
+			// Only delete keystone if it is in the previously found block.
+			if fbh.IsEqual(&v.BlockHash) {
+				kssBatch.Delete(k[:])
+			}
 		case 1:
 			has, err := kssTx.Has(k[:], nil)
 			if err != nil {
--- a/database/tbcd/level/level.go
+++ b/database/tbcd/level/level.go
@@ -27,6 +27,7 @@ import (
 	"github.com/hemilabs/heminetwork/database/level"
 	"github.com/hemilabs/heminetwork/database/tbcd"
 	"github.com/hemilabs/heminetwork/hemi"
+	"github.com/hemilabs/heminetwork/hemi/pop"
 )
 
 // Locking order:
@@ -281,6 +282,27 @@ func (l *ldb) MetadataBatchGet(ctx context.Context, allOrNone bool, keys [][]byt
 	defer mdDiscard()
 
 	return l.transactionBatchGet(ctx, mdDB, allOrNone, keys)
+}
+
+func (l *ldb) L2KeystoneAbrevByAbrevHash(ctx context.Context, abrevhash []byte) (*hemi.L2KeystoneAbrev, error) {
+	log.Tracef("L2KeystoneAbrevByAbrevHash")
+	defer log.Tracef("L2KeystoneAbrevByAbrevHash exit")
+
+	kssDB := l.pool[level.KeystonesDB]
+	eks, err := kssDB.Get(abrevhash, nil)
+	if err != nil {
+		if errors.Is(err, leveldb.ErrNotFound) {
+			return nil, database.NotFoundError(fmt.Sprintf("l2 keystone not found: %v", abrevhash))
+		}
+		return nil, fmt.Errorf("l2 keystone get: %w", err)
+	}
+	tl2, err := pop.ParseTransactionL2FromOpReturn(eks)
+	if err != nil {
+		log.Errorf("can't parse PKScript from keystone DB: %s", err)
+		return nil, err
+	}
+
+	return tl2.L2Keystone, nil
 }
 
 // BatchAppend appends rows to batch b.
--- a/database/tbcd/level/level.go
+++ b/database/tbcd/level/level.go
@@ -27,7 +27,6 @@ import (
 	"github.com/hemilabs/heminetwork/database/level"
 	"github.com/hemilabs/heminetwork/database/tbcd"
 	"github.com/hemilabs/heminetwork/hemi"
-	"github.com/hemilabs/heminetwork/hemi/pop"
 )
 
 // Locking order:
@@ -284,25 +283,25 @@ func (l *ldb) MetadataBatchGet(ctx context.Context, allOrNone bool, keys [][]byt
 	return l.transactionBatchGet(ctx, mdDB, allOrNone, keys)
 }
 
-func (l *ldb) BlockKeystoneAbrevByL2KeystoneAbrevHash(ctx context.Context, abrevhash []byte) (*hemi.L2KeystoneAbrev, error) {
-	log.Tracef("BlockKeystoneAbrevByL2KeystoneAbrevHash")
-	defer log.Tracef("BlockKeystoneAbrevByL2KeystoneAbrevHash exit")
+func (l *ldb) BlockKeystoneByL2KeystoneAbrevHash(ctx context.Context, abrevhash chainhash.Hash) (*tbcd.Keystone, error) {
+	log.Tracef("BlockKeystoneByL2KeystoneAbrevHash")
+	defer log.Tracef("BlockKeystoneByL2KeystoneAbrevHash exit")
 
 	kssDB := l.pool[level.KeystonesDB]
-	eks, err := kssDB.Get(abrevhash, nil)
+	eks, err := kssDB.Get(abrevhash.CloneBytes(), nil)
 	if err != nil {
 		if errors.Is(err, leveldb.ErrNotFound) {
 			return nil, database.NotFoundError(fmt.Sprintf("l2 keystone not found: %v", abrevhash))
 		}
 		return nil, fmt.Errorf("l2 keystone get: %w", err)
 	}
-	tl2, err := pop.ParseTransactionL2FromOpReturn(eks)
-	if err != nil {
-		log.Errorf("can't parse PKScript from keystone DB: %s", err)
-		return nil, err
+
+	ks := tbcd.Keystone{
+		BlockHash:           chainhash.Hash(eks[:32]),
+		AbbreviatedKeystone: eks[32:],
 	}
 
-	return tl2.L2Keystone, nil
+	return &ks, nil
 }
 
 // BatchAppend appends rows to batch b.
@@ -1770,7 +1769,7 @@ func (l *ldb) BlockKeystoneUpdate(ctx context.Context, direction int, keystones 
 				// Only store unknown keystones
 				continue
 			}
-			var value [chainhash.HashSize + hemi.L2KeystoneAbrevSize]byte
+			var value [chainhash.HashSize + hemi.L2KeystoneAbrevSize + 7]byte
 			copy(value[0:32], v.BlockHash[:])
 			copy(value[32:], v.AbbreviatedKeystone)
 			kssBatch.Put(k[:], value[:])
--- a/database/tbcd/level/level.go
+++ b/database/tbcd/level/level.go
@@ -1707,9 +1707,9 @@ func (l *ldb) BlockTxUpdate(ctx context.Context, direction int, txs map[tbcd.TxK
 	return nil
 }
 
-func (l *ldb) BlockHemiUpdate(ctx context.Context, direction int, keystones map[chainhash.Hash][]byte) error {
-	log.Tracef("BlockHemiUpdate")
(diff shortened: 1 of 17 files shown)


service/tbc/crawler.go
```diff
@@ -1215,20 +1214,387 @@ func (s *Server) TxIndexer(ctx context.Context, endHash *chainhash.Hash) error {
 	return fmt.Errorf("invalid direction: %v", direction)
 }
 
+func processKeystones(blockHash *chainhash.Hash, txs []*btcutil.Tx, kssCache map[chainhash.Hash]tbcd.Keystone) error {
```
ClaytonNorthey92, line 1217: TODO before merge: add a test case for this

database/tbcd/level/level.go
```diff
@@ -282,6 +283,27 @@ func (l *ldb) MetadataBatchGet(ctx context.Context, allOrNone bool, keys [][]byt
 	return l.transactionBatchGet(ctx, mdDB, allOrNone, keys)
 }
 
+func (l *ldb) BlockKeystoneByL2KeystoneAbrevHash(ctx context.Context, abrevhash chainhash.Hash) (*tbcd.Keystone, error) {
+	log.Tracef("BlockKeystoneByL2KeystoneAbrevHash")
+	defer log.Tracef("BlockKeystoneByL2KeystoneAbrevHash exit")
+
+	kssDB := l.pool[level.KeystonesDB]
```
ClaytonNorthey92, line 290: let's add a check here to ensure that this exists

```
kssDB, ok := ...
if !ok { ...
```

database/tbcd/level/level.go
```diff
@@ -1707,6 +1729,66 @@ func (l *ldb) BlockTxUpdate(ctx context.Context, direction int, txs map[tbcd.TxK
 	return nil
 }
 
+func (l *ldb) BlockKeystoneUpdate(ctx context.Context, direction int, keystones map[chainhash.Hash]tbcd.Keystone) error {
+	log.Tracef("BlockKeystoneUpdate")
+	defer log.Tracef("BlockKeystoneUpdate exit")
+
+	if !(direction == 1 || direction == -1) {
```
ClaytonNorthey92, line 1736: can we define these directions as contants:
```
const DirectionForward = 1
const DirectionBackward = -1
```

or something similar? it's not clear what we're doing unless I read through this function (which may grow or become more complex over time)

joshuasing: I agree with this. I believe we previously discussed having a Direction type but decided against it - this would significantly improve readability without adding a type.

database/tbcd/level/level.go
```diff
@@ -1707,6 +1729,66 @@ func (l *ldb) BlockTxUpdate(ctx context.Context, direction int, txs map[tbcd.TxK
 	return nil
 }
 
+func (l *ldb) BlockKeystoneUpdate(ctx context.Context, direction int, keystones map[chainhash.Hash]tbcd.Keystone) error {
```
ClaytonNorthey92, line 1732: are there any tests written specifically for this function? I don't see any 🤔 

we need the following test cases:
* in both directions
  * empty `keystones` map
  * `nil` keystones` map
  * a keystone that does not exist in the db
  * a keystone that does exist in the db
  * multiple keystones that exist in the db
  * multiple keystones that don't exist in the db
  * multiple keystones that do and don't exist in the db
  
for each of those, ensure that the data that we store can also be de-serialized correctly (i.e. a read should give you back what a write stored, we're not trimming bytes etc.)

database/tbcd/level/level.go
```diff
@@ -1707,6 +1729,66 @@ func (l *ldb) BlockTxUpdate(ctx context.Context, direction int, txs map[tbcd.TxK
...
+		case 1:
+			has, err := kssTx.Has(k[:], nil)
+			if err != nil {
+				return fmt.Errorf("keystone update has: %w", err)
+			}
+			if has {
+				// Only store unknown keystones
+				continue
+			}
+			var value [chainhash.HashSize + hemi.L2KeystoneAbrevSize]byte
```
ClaytonNorthey92, line 1772: let's assert that `(len(v.BlockHash) + len(v.AbbreviatedKeystone) == chainhash.HashSize + hemi.L2KeystoneAbrevSize`

if not, we could get data corruption here, which would be bad

service/tbc/crawler.go
ClaytonNorthey92: I think we need to a add a test case here, to ensure that we pull the correct bytes off of a `btcutil.Tx` and store them correctly

service/tbc/crawler.go
```diff
@@ -1215,20 +1220,379 @@ func (s *Server) TxIndexer(ctx context.Context, endHash *chainhash.Hash) error {
...
+
+		for _, txOut := range tx.MsgTx().TxOut {
+			aPoPTx, err := pop.ParseTransactionL2FromOpReturn(txOut.PkScript)
+			if err != nil {
+				// log.Tracef("error parsing tx l2: %s", err)
+				continue
+			}
+			if _, ok := kssCache[*aPoPTx.L2Keystone.Hash()]; ok {
+				// Multiple keystones may exist in block, only store first
+				continue
```
ClaytonNorthey92, line 1238: I don't think we should continue here; there may be a case where another keystone is found

service/tbc/crawler.go
```diff
@@ -1215,20 +1220,379 @@ func (s *Server) TxIndexer(ctx context.Context, endHash *chainhash.Hash) error {
...
+				height, err)
+		}
+		index, err := s.findPathFromHash(ctx, endHash, bhs)
+		if err != nil {
+			return 0, last, fmt.Errorf("could not determine canonical path %v: %w",
+				height, err)
+		}
+		// Verify it connects to parent
+		if !hash.IsEqual(bhs[index].ParentHash()) {
+			return 0, last, fmt.Errorf("%v does not connect to: %v",
```
ClaytonNorthey92, line 1326: hey @marcopeereboom what is the intention here? If a block doesn't connect to the expected parent (I assume a re-org could cause this?) then we just stop indexing?

ClaytonNorthey92 (changes requested):
hey @marcopeereboom I have a few changes to request; most requests are around test coverage, there is one indexing question that I have as well.

database/tbcd/level/level.go
marcopeereboom: If it doesn't exist we crash. That's intended because the caller did it wrong. This pattern is used all over the code.

database/tbcd/level/level.go
marcopeereboom: Yeah we keep having this conversation and while i do not expressively disagree in this case the code would overflow lines everywhere rendering it even more unreadable. And grade school math did tell us that negative is on the left and positive on the right.

I am sympathetic to the argument but code readability wins.

service/tbc/crawler.go
```diff
@@ -1215,20 +1220,379 @@ func (s *Server) TxIndexer(ctx context.Context, endHash *chainhash.Hash) error {
...
+	}
+
+	return blocksProcessed, last, nil
+}
+
+func (s *Server) KeystoneIndexerUnwind(ctx context.Context, startBH, endBH *tbcd.BlockHeader) error {
+	log.Tracef("KeystoneIndexerUnwind")
+	defer log.Tracef("KeystoneIndexerUnwind exit")
+
+	// XXX dedup with KeystoneIndexerWind; it's basically the same code but with the direction, start anf endhas flipped
```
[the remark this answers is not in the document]
marcopeereboom: This is copy pasted 3 times and is crawler boilerplate. This is a huge lift and needs to happen but unfortunately we have bigger fish to fry right now.

service/tbc/crawler.go
marcopeereboom: This is subtle but correct. We only care about the very first UNIQUE keystone.

See slack conversation for the explanation.

marcopeereboom: a

database/tbcd/level/level.go
marcopeereboom: These are all chainhash.Hash that are forced to 32 bytes.

service/tbc/crawler.go
marcopeereboom: it means we do not have all data and thus we cannot continue. Either we have data corruption or a blockheader was missed (which can happen in the insane world of testnet3). We return to give the caller the opportunity to reinsert the blockheader and block. @max-sanchez actually does that in the embedded tbc service in op-geth.

service/tbc/crawler.go
```diff
@@ -1215,20 +1220,379 @@ func (s *Server) TxIndexer(ctx context.Context, endHash *chainhash.Hash) error {
...
+			last = hh
+			// Flush
+			break
+		}
+
+		// Move to previous block
+		height := bh.Height - 1
+		pbh, err := s.db.BlockHeaderByHash(ctx, bh.ParentHash())
+		if err != nil {
+			return 0, last, fmt.Errorf("block headers by height %v: %w",
```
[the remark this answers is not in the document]
marcopeereboom: See response there too

database/tbcd/level/level.go
AL-CT: Made a [push](https://github.com/hemilabs/heminetwork/pull/380/commits/926c7dd779fef51643ad2ff7851f6c71f6c8f8da#diff-4432cd20e2ef8143b40647cc6ca5cd8795b49da98cb7400b6573e8893ac473e5R955-R964) that is used to check for these:
- a keystone that does not exist in the db
- a keystone that does exist in the db
- multiple keystones that exist in the db
- multiple keystones that don't exist in the db
- multiple keystones that do and don't exist in the db

They aren't explicitly for this function, but indirectly do some of the work

database/tbcd/level/level_test.go
```diff
@@ -112,3 +125,265 @@ func TestMD(t *testing.T) {
...
+
+func TestKeystoneUpdate(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+
+	home := t.TempDir()
+	t.Logf("temp: %v", home)
+
+	cfg := level.NewConfig(home, "", "")
+	db, err := level.New(ctx, cfg)
```
ClaytonNorthey92, line 147: let's do this before each test, in the `t.Run(...)` function.  this way we start with a _fresh database_ on every test, so we know the state of the test when it starts.  the way the test is written now, each test case will leave data behind for the next case, so we're not starting with a clean state.

database/tbcd/level/level_test.go
```diff
@@ -112,3 +125,265 @@ func TestMD(t *testing.T) {
...
+			for v := range tti.expectedInDB {
+				_, err := db.BlockKeystoneByL2KeystoneAbrevHash(ctx, v)
+				if err != nil {
+					t.Fatalf("keystone not in db: %v", err)
+				}
+			}
+
+			for k, v := range tti.expectedOutDB {
+				_, err := db.BlockKeystoneByL2KeystoneAbrevHash(ctx, k)
+				if err == nil {
```
ClaytonNorthey92, line 324: let's also add a check here to ensure that the error is a "not found" error, otherwise _any_ error would pass this test

AL-CT: good catch, also fixed👍

database/tbcd/level/level_test.go
```diff
@@ -112,3 +125,268 @@ func TestMD(t *testing.T) {
...
+		}
+
+		abrvKss := aPoPTx.L2Keystone.Serialize()
+
+		kssCache := validKssCache
+		if i > 2 {
+			kssCache = invalidKssCache
+		}
+
+		kssCache[*hemi.L2KeystoneAbbreviate(l2Keystone).Hash()] = tbcd.Keystone{
```
ClaytonNorthey92, line 200: is `kssCache` used anywhere? 🤔 can we remove?

database/tbcd/level/level_test.go
```diff
@@ -112,3 +125,268 @@ func TestMD(t *testing.T) {
...
+					t.Fatalf("keystone not in db: %v", err)
+				}
+			}
+
+			for k, v := range tti.expectedOutDB {
+				_, err := db.BlockKeystoneByL2KeystoneAbrevHash(ctx, k)
+				if err == nil {
+					t.Fatalf("keystone in db: %v", spew.Sdump(v))
+				} else {
+					_, ok := err.(database.NotFoundError)
```
ClaytonNorthey92, line 325: it's probably better to use `errors.Is` here

database/tbcd/level/level_test.go
```diff
@@ -112,3 +125,268 @@ func TestMD(t *testing.T) {
...
+			kssMap:         validKssCache,
+			expectedInDB:   validKssCache,
+			expectedOutDB:  invalidKssCache,
+		},
+
+		{
+			name:          "invalidRemove",
+			direction:     []int{-1},
+			kssMap:        invalidKssCache,
+			expectedOutDB: anyKssCache,
```
ClaytonNorthey92, line 251: hey @AL-CT could we define these not as variables, but as each their own `map[chainhash.Hash]tbcd.Keystone`?

for example:
```go
expectedOutDB: map[chainhash.Hash]tbcd.Keystone{
  ...
}
```

I fear that sharing variables between tests here may cause unexpected behavior

AL-CT: fair, also forced me to make things more readable when fixing it

service/tbc/crawler_test.go
```diff
@@ -123,3 +123,7 @@ func BenchmarkMap1000000(b *testing.B) {
 		allocateMap(1e6)
 	}
 }
+
+func TestProcessKeystones(t *testing.T) {
```
AL-CT, line 127: forgot to delete this

service/tbc/crawler.go
AL-CT: Changed `mustHaveKss` to test the results of processKeystones explicitly in the fork testing.

ClaytonNorthey92 (approved):

service/tbc/tbcfork_test.go
```diff
@@ -229,7 +249,7 @@ func (b *btcNode) newSignedTxFromTx(name string, inTx *btcutil.Tx, amount btcuti
 	if err != nil {
 		return nil, err
 	}
-	b.t.Logf("rdeeem pkScript: %x", pkScript)
+	b.t.Logf("rdeeem pkScript (%v): %x", name, pkScript)
```
joshuasing, line 252: nit: typo
```suggestion
	b.t.Logf("redeem pkScript (%v): %x", name, pkScript)
```

joshuasing (approved):

PR #384, opened 2025-01-24
tbc: add UtxosByScriptHashCount

marcopeereboom:
**Summary**
We need a way to count all utxos for a specified scripthash. Unfortunately this does mean iterating through all utxos since leveldb does not have the upper limit as a call.


joshuasing (approved):

jcvernaleo (approved):
OK

PR #395, opened 2025-02-05
tbc: add keystone indexer to synced calculation

marcopeereboom:
**Summary**
I forgot to add a check for keystone indexer when calculating synced.

**Changes**
Add logic to determine if chain is synced when keystone indexers are enabled.


jcvernaleo (approved):
OK

ClaytonNorthey92 (approved):

PR #399, opened 2025-02-08
Add LevelDB compaction handling (#388)

### LINKED ISSUE #388: tbc: Rethink CompactRange on leveldb transaction commits
### Confirmation

### Problem

We use leveldb transactions willy-nilly in the code to keep the caller honest. This should be fine if it wasn't for an idiosyncrasy in the leveldb code. Per the leveldb transaction documentation:
```
// Transaction is very expensive and can overwhelm compaction, especially if
// transaction size is small. Use with caution.
// The rule of thumb is if you need to merge at least same amount of
// `Options.WriteBuffer` worth of data then use transaction, otherwise don't.
```

When we don't compact we sometimes read back the wrong data, nay the previous data really. I do not know the impact of compaction on IBD and/or runtime and thus this needs to be reevaluated.

### Suggested solution

None of options available to do a quick fix are palatable. The leveldb packages embeds receivers and treats them as objects and obscures vital internals that are needed to determine the need for compaction.

There are several possibilities. To do this externally or internally.

If we go the internal route we need to clone the leveldb repo into our code and add the missing functionality to batch. We need the internalLen of a batch to determine when to flush.

If we go the external route we need to hijack the startTransaction call in tbcd/leveldb and fully wrap the database transaction so that we can hijack the Put call to keep track of internalLen ourselves (8 + len(key) + len(value)). The added complexity is that shitty leveldb implementation hides the WriteBuffer option which is the value needed to determine if we indeed need to flush.

Yet another option is to keep track of internalLen in the caller (our database code) and thus do the magic there. This is super error prone and a bunch of copy/paste that will quickly turn to shit.

### Additional Information

_No response_

VolodymyrBg:
## Description
This PR implements automatic compaction handling for LevelDB transactions to prevent performance degradation over time. The changes monitor database statistics and trigger background compaction when needed.

## Changes
- Added background compaction mechanism after transaction commits
- Implemented database stats monitoring to determine when compaction is needed
- Added proper error handling and logging for compaction operations
- Added CompactionThreshold constant for future tuning

## Testing
The changes have been tested with:
- Normal transaction operations
- High-volume transaction scenarios
- Error handling verification

## Related Issues
Fixes #388

## Notes
- Compaction runs in the background to minimize impact on normal operations
- Added logging to track compaction operations
- The CompactionThreshold can be adjusted based on performance requirements

### DIFF
--- a/database/tbcd/level/level.go
+++ b/database/tbcd/level/level.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2024-2025 Hemi Labs, Inc.
+// Copyright (c) 2024 Hemi Labs, Inc.
 // Use of this source code is governed by the MIT License,
 // which can be found in the LICENSE file.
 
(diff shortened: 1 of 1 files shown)


marcopeereboom:
bzzzt wrong LRU implementation. the whole point of trashing the hashicorp lru (which is too fat) is to have variable sized data.

PR #406, opened 2025-02-15
Clayton/localnode mainnet

ClaytonNorthey92:
**Summary**
update mainnet localnode compose + localnet to the latest op-geth and op-node

added needed flags, files, etc. for hvm phase 0

NOTE: this breaks localnet at op-geth+tbc startup on blank directory

**Changes**
see summary

### DIFF
--- a/localnode/docker-compose_mainnet.yml
+++ b/localnode/docker-compose_mainnet.yml
@@ -151,7 +151,7 @@ services:
 
   # Hemi L2 node
   op-geth-l2:
-    image: "ghcr.io/hemilabs/op-geth@sha256:43860fcb691c85e344bd17a660a19e254924953b2ccebe03f2d87d425eeaecbb" # 337407d
+    image: "ghcr.io/hemilabs/op-geth:a5a12ae@sha256:6e21d631b073a363ff859f4e82bbbcbde339801b5f5e5b9e42a5eb34081e054e"
     depends_on:
       geth-l1:
         condition: "service_started"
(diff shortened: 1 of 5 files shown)


jcvernaleo (commented):
I think you should add hemi-mainnet-rollup.json as well.

localnode/docker-compose_mainnet.yml
```diff
@@ -160,7 +160,7 @@ services:
       op-geth-l2-init-tbc:
         condition: "service_completed_successfully"
     volumes:
-      - ./mainnet-genesis.json:/tmp/genesis.json
+      - ./mainnetgenesis.json:/tmp/genesis.json
```
jcvernaleo, line 163: Why get rid of the hyphen?  mainnet-genesis.json is more like the naming for testnet so I think we need to stay consistant.

jcvernaleo (changes requested):

ClaytonNorthey92:
> I think you should add hemi-mainnet-rollup.json as well.

@jcvernaleo yes! whoops. will add that

localnode/docker-compose_mainnet.yml
ClaytonNorthey92: laziness on my end; I downloaded the file without a hyphen and just changed it in the compose file 🤦 will update

jcvernaleo (approved):
OK

joshuasing (approved):

PR #408, opened 2025-02-17
fragile test fixes

ClaytonNorthey92:
**Summary**
a few fragile test fixes

**Changes**
as of now, op-conductor is not hvm-aware so we do not use it in higher environments.  there is a problematic step in localnet testing that is meant to test some op-conductor functionality.  this slows down the tests, is incorrect, and is inconsistent.  remove until we update op-conductor, at which time we will add an appropriate sequencer failover test

decrease l2 block time to speed up testing

a test that tests two bfgs doesn't wait for the insertion of l2 keystones, add that wait and add t.Parallel() so it doesn't block other parallel tests.


jcvernaleo (approved):
OK

max-sanchez (approved):

joshuasing (approved):

PR #409, opened 2025-02-18
all: fix int64 and uint64 casting to int

AL-CT:
**Summary**
Github detected various CodeQL vulnerabilites related to a few instances in the code where strings were getting parsed to int64 and uint64, which were later getting cast to int without checking the upper bounds.

**Changes**
Changed to parse these with 32 bits explicitly, so at the very least an error will return if the parsed string exceeds the int32 bounds, rather than a potential undetected overflow in 32-bit architectures.

### DIFF
--- a/cmd/hemictl/hemictl.go
+++ b/cmd/hemictl/hemictl.go
@@ -13,6 +13,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"math"
 	"os"
 	"os/user"
 	"path/filepath"
@@ -415,8 +416,11 @@ func tbcdb() error {
 		maxCache := args["maxcache"]
 		var mc uint64
 		if maxCache != "" {
-			if mc, err = strconv.ParseUint(maxCache, 10, 32); err != nil {
+			if mc, err = strconv.ParseUint(maxCache, 10, 64); err != nil {
 				return fmt.Errorf("maxCache: %w", err)
+			}
+			if mc > math.MaxInt32 {
+				return fmt.Errorf("maxcache exceeds max int value")
 			}
 			cfg.MaxCachedTxs = int(mc)
 		}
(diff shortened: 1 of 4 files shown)


marcopeereboom (approved):

cmd/hemictl/hemictl.go
```diff
@@ -418,6 +419,9 @@ func tbcdb() error {
 			if mc, err = strconv.ParseUint(maxCache, 10, 64); err != nil {
 				return fmt.Errorf("maxCache: %w", err)
 			}
+			if mc > math.MaxInt32 {
+				return fmt.Errorf("maxcache exceeds max int value")
+			}
 			cfg.MaxCachedTxs = int(mc)
```
joshuasing, line 425: It appears `strconv#ParseInt` supports providing a `bitSize` of `0`, which will use the platform's `int` bit size (int32/int64) - per https://pkg.go.dev/strconv#ParseInt. This allows us to cast the output to `int` without any issues. This could also be used for all of the other cases where `ParseInt` or `ParseUint` is used and the output is cast to `int` or `uint`.

There also appears to be some cases here where we are using `ParseUint` and casting to an `int`, presumably because we don't want to allow inputs less than zero - In these cases, I think either we should add a bounds check or the config value type should be changed to `uint`.

```suggestion
			if mc, err = strconv.ParseInt(maxCache, 10, 0); err != nil {
				return fmt.Errorf("maxCache: %w", err)
			}
			cfg.MaxCachedTxs = int(mc)
```

joshuasing (changes requested):

joshuasing (approved):

cmd/hemictl/hemictl.go
AL-CT: Ah nice, fixed!

PR #411, opened 2025-02-18
tbc: use proper database for index hashes

marcopeereboom:
**Summary**
For performance reasons we ended up updating index hashes in metadata. This turned out to be a mistake because op-geth punished us by killing tbc in between the index and the index hash update.

**Changes**
Move index hashes into the proper database thus making it part of the database transaction thus making it atomic.


ClaytonNorthey92 (changes requested):
one small change requested, one nitpick

service/tbc/tbc_test.go
```diff
@@ -121,6 +122,64 @@ func countKeystones(b *btcutil.Block) int {
...
+
+	_, err = s.BlockHeadersByHeight(ctx, 9)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Logf("testdb upgrade successful")
+
+}
```
joshuasing, line 181: There looks to be unnecessary blank lines at the start/end of this function:
```suggestion
func TestDbUpgrade(t *testing.T) {
	home := t.TempDir()
	t.Logf("temp: %v", home)

	err := extract("testdata/testdatabase.tar.gz", home)
	if err != nil {
		t.Fatal(err)
	}

	ctx, cancel := context.WithCancel(context.Background())
	defer func() {
		cancel()
	}()

	// Connect tbc service
	cfg := &Config{
		AutoIndex:            false,
		BlockCacheSize:       "10mb",
		BlockheaderCacheSize: "1mb",
		BlockSanity:          false,
		LevelDBHome:          home,
		//LogLevel:                "tbcd=TRACE:tbc=TRACE:level=DEBUG",
		MaxCachedTxs:            1000, // XXX
		MaxCachedKeystones:      1000, // XXX
		Network:                 "testnet3",
		PrometheusListenAddress: "",
		ListenAddress:           "",
		PeersWanted:             0,
		MempoolEnabled:          false,
		Seeds:                   []string{"127.0.0.1:18444"},
	}
	_ = loggo.ConfigureLoggers(cfg.LogLevel)
	s, err := NewServer(cfg)
	if err != nil {
		t.Fatal(err)
	}
	go func() {
		err := s.Run(ctx)
		if err != nil && !errors.Is(err, context.Canceled) {
			panic(err)
		}
	}()

	time.Sleep(1000 * time.Millisecond)

	s.syncBlocks(ctx)

	_, err = s.BlockHeadersByHeight(ctx, 9)

	if err != nil {
		t.Fatal(err)
	}

	t.Logf("testdb upgrade successful")
}
```

service/tbc/tbc.go
```diff
@@ -2641,10 +2658,11 @@ func (s *Server) ExternalHeaderSetup(ctx context.Context, upstreamStateId []byte
 	defer log.Tracef("ExternalHeaderSetup exit")
 
 	if !s.cfg.ExternalHeaderMode {
-		return errors.New("ExternalHeaderSetup called but external header mode is not enabled in config")
+		return errors.New("ExternalHeaderSetup called but external" +
```
joshuasing, line 2661: There's a missing space here, so the error would be "...externalheader mode..."
```suggestion
		return errors.New("ExternalHeaderSetup called but external " +
```

service/tbc/tbc_test.go
AL-CT: Yup, thanks!

joshuasing (changes requested):

service/tbc/util_test.go
```diff
@@ -0,0 +1,158 @@
...
+	// log.Printf("Extracting: %v -> %v\n", filename, dst)
+	var err error
+	archive := filepath.Ext(filename)
+	switch archive {
+	case ".zip":
+		_, err = unzip(filename, dst)
+	case ".gz":
+		err = gunzip(filename, dst)
+	default:
+		err = fmt.Errorf("Unknown archive type: %v", archive)
```
joshuasing, line 155: Error strings shouldn't be capitalised: https://go.dev/wiki/CodeReviewComments#error-strings
```suggestion
		err = fmt.Errorf("unknown archive type: %v", archive)
```

joshuasing (changes requested):

cmd/hemictl/hemictl.go
```diff
@@ -185,41 +198,32 @@ func parseArgs(args []string) (string, map[string]string, error) {
...
-		err := s.DBClose()
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "db close: %v\n", err)
-			os.Exit(1)
+	go func() {
+		if err := s.Run(ctx); err != nil {
+			panic(fmt.Errorf("run server: %w", err))
 		}
 	}()
+	for !s.Running() {
```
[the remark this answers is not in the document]
AL-CT: Run() doesn't actually return until all goroutines inside tbc are finished, so it would be stuck there

service/tbc/tbc.go
AL-CT: Fixed

service/tbc/util_test.go
AL-CT: Fixed

cmd/hemictl/hemictl.go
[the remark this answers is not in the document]
marcopeereboom: This is a test and that's why it is ok to pretend to be a mutex.

cmd/hemictl/hemictl.go
```diff
@@ -1090,6 +1088,26 @@ func printJSON(where io.Writer, indent string, payload any) error {
...
+
+	select {
+	case <-ctx.Done():
+	case s := <-signalChan: // First signal, cancel context.
+		if callback != nil {
+			callback(s) // Do whatever caller wants first.
+			cancel()
+		}
+	}
+	<-signalChan // Second signal, hard exit.
```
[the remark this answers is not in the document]
marcopeereboom: Yes. We use this pattern in all our daemons.

service/tbc/tbc_test.go
```diff
@@ -121,6 +122,61 @@ func countKeystones(b *btcutil.Block) int {
 	return keystonesFound
 }
 
+func TestDbUpgrade(t *testing.T) {
```
ClaytonNorthey92, line 125: if I add this change, which returns early from the "upgrade db for loop", this test still passes. therefore I do not believe it's actually testing the upgrade.  could we be more specific here? what _should_ the state of the db be before and after?
```
diff --git a/database/tbcd/level/level.go b/database/tbcd/level/level.go
index e1cefad..2375b5a 100644
--- a/database/tbcd/level/level.go
+++ b/database/tbcd/level/level.go
@@ -211,6 +211,7 @@ func New(ctx context.Context, cfg *Config) (*ldb, error) {
 
        // Upgrade database
        for {
+               return l, nil
                dbVersion, err := l.Version(ctx)
                if err != nil {
                        if errors.Is(err, leveldb.ErrNotFound) {
```

marcopeereboom: I added a test to verify all end conditions of the upgrade and we added a new tar file that is moved from v1 to v2.

AL-CT (approved):

ClaytonNorthey92 (approved):

PR #415, opened 2025-02-19
bfg: take into account l2 keystones that are more final than 100

ClaytonNorthey92:
**Summary**
there is a bug in the postgres query that only takes into account the first 100 bitcoin blocks when determining effective height, so if a keystone was published older than the first 100 it wouldn't get counted in effective height.

remove this clause; I am not sure why it was introduced but it is likely fallout from when trying to optimize these queries

add a test that should have caught this

**Changes**
see summary


ClaytonNorthey92:
meh one bug in my change...fixing

jcvernaleo (approved):
OK

joshuasing (approved):

Issue #413, opened 2025-02-19
hemictl: implement annotated commands in tbcdb


marcopeereboom:
### Confirmation

- [x] I have checked for similar issues.

### Problem

There are a bunch of missing commands in hemictl tbcdb action. Please add and fix.

### Suggested solution

write code

joshuasing:
@AL-CT Have we done this? 🤔

Issue #283, opened 2024-10-14
Add directory config for hemictl


marcopeereboom:
### Confirmation

- [x] I have checked for similar issues.

### Problem

We need to be able to set `cfg.LevelDBHome = "~/.tbcd"` when launching hemictl. While digging in there also add network (testnet3/mainnet etc)

### Suggested solution

Add code.

marcopeereboom:
Fixed in #411

Issue #221, opened 2024-08-21
hemictl: overall clean up


joshuasing:
### Confirmation

- [X] I have checked for similar issues.

### Suggested solution

The code for `hemictl` is currently messy, it would be nice to clean it up.

Issue #220, opened 2024-08-21
hemictl: parse key/value from args


joshuasing:
### Confirmation

- [X] I have checked for similar issues.

### Suggested solution

In `hemictl`, when making a RPC request like `tbcapi-block-headers-by-height`, we should be able to provide arguments in the format `height=100` instead of needing to provide them as JSON (e.g. `{"height": 100}`).

PR #419, opened 2025-02-24
hemictl: various fixes and overall improvement

AL-CT:
**Summary**
Hemictl is a little messy and lacking some functionalities, as detailed in the following issues:
- #220
- #221 
- #283 
- #413 

**Changes**
- Allow to pass arguments as key/value pairs (not just json) for generic commands
- Introduce the "api" subcommand which serves as a prefix for passing api generic commands
- "Cleanup" the code so it is a little more readable
- Add more comprehensive "help" commands to many of the daemons
- Add flags parsing (e.g., "-h, -help") for the subcommands
- Add a few TBC related actions


cmd/hemictl/hemictl.go
```diff
@@ -1142,12 +1232,182 @@ func HandleSignals(ctx context.Context, cancel context.CancelFunc, callback func
...
+			b, err = Jsonify(args[1:])
+			if err != nil {
+				return nil, err
+			}
+			log.Infof("parsed arguments as %v", b)
+		}
+		err = json.Unmarshal([]byte(b), &clone)
+		if err != nil {
+			return nil, fmt.Errorf("invalid payload: %w", err)
+		}
```
joshuasing, line 1279: Instead of attempting to parse it as JSON, then returning a bool for whether it succeeds - what do you think about: attempt to decode it, use it if it does decode, otherwise attempt Jsonify and decode again?

This saves us decoding it twice if JSON is passed in.

cmd/hemictl/hemictl.go
```diff
@@ -1142,12 +1232,182 @@ func HandleSignals(ctx context.Context, cancel context.CancelFunc, callback func
...
+	}
+	return clone, nil
+}
 
+// hemictlApi is a structure used to satisfy the protocol.API interface.
+type hemictlApi struct {
```
joshuasing, line 1285: Go generally recommend using ALLCAPS for initialisms and acronyms: https://go.dev/wiki/CodeReviewComments#initialisms

```suggestion
type hemictlAPI struct {
```

joshuasing (changes requested):
Looks good, just a few minor comments

cmd/hemictl/hemictl.go
AL-CT: Good point, fixed!

cmd/hemictl/hemictl.go
AL-CT: Noted. Fixed 👍

joshuasing (approved):

marcopeereboom (approved):

PR #429, opened 2025-03-06
tbc: use chainhash.Hash without pointers in the lowIQLRU cache

marcopeereboom:
**Summary**
Add a diagnostic to root cause a bug.

**Changes**


jcvernaleo (approved):
I've been running with this on a fresh data directory since Friday without any issues.

And the code looks reasonable too :)

OK

joshuasing (approved):

PR #431, opened 2025-03-10
bfg: add config option to disable public ws conns

joshuasing:
**Summary**
Add a configuration option `BFG_DISABLE_PUBLIC_CONNS` to disable public WebSocket connections on BFG.
When a PoP Miner connects and this option is enabled, the PoP Miner will be disconnected with an authentication error, which will result in the PoP Miner not retrying the connection.

**Changes**
- Add `BFG_DISABLE_PUBLIC_CONNS` environment variable to BFG.
- Add handling to close public WebSocket connections with a `protocol.ErrPublicKeyAuth` error when `DisablePublicConns` is `true`.

### DIFF
--- a/cmd/bfgd/bfgd.go
+++ b/cmd/bfgd/bfgd.go
@@ -129,6 +129,12 @@ var (
 			Help:         "a btc private key, this is only needed when connecting to another BFG",
 			Print:        config.PrintSecret,
 		},
+		"BFG_DISABLE_PUBLIC_CONNS": config.Config{
+			Value:        &cfg.DisablePublicConns,
+			DefaultValue: false,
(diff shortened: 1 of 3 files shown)


ClaytonNorthey92 (changes requested):
hey @joshuasing , could we add a test case for this?  you can find an example that you could copy and paste from here https://github.com/hemilabs/heminetwork/pull/359/files#diff-f7bc76e889209de062a2c7c80346ee555bf7d7b24f5290830fac4a06cc87dcd5L4011

ClaytonNorthey92 (approved):

AL-CT (approved):

jcvernaleo (approved):
OK

PR #432, opened 2025-03-10
all: re-enable gosec linter

AL-CT:
**Summary**
The gosec linter was disabled due to the sheer amount of noise it produced. This pr fixes the flagged vulnerabilities in order to re-enable the linter.

**Changes**
Fixed most the vulnerabilities flagged by gosec,  with the exception of [G115](https://github.com/securego/gosec/blob/9452efe4ad770f703df825fb32cd693b421fc916/README.md?plain=1#L143) which produces too many false positives, and was thus kept disabled.

### DIFF
--- a/rawdb/rawdb.go
+++ b/rawdb/rawdb.go
@@ -8,6 +8,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"sync"
@@ -165,6 +166,14 @@ func (r *RawDB) Insert(key, value []byte) error {
 		} else {
 			// Encoded coordinates.
 			c := make([]byte, 4+4+4)
+
+			if fi.Size() > math.MaxUint32 {
+				return fmt.Errorf("file info exceeds max uint32 %v", fi.Size())
+			}
+			if len(value) > math.MaxUint32 {
+				return fmt.Errorf("size of value exceeds max uint32 %v", len(value))
+			}
+
 			binary.BigEndian.PutUint32(c[0:4], last)
 			binary.BigEndian.PutUint32(c[4:8], uint32(fi.Size()))
 			binary.BigEndian.PutUint32(c[8:12], uint32(len(value)))
--- a/service/popm/popm.go
+++ b/service/popm/popm.go
@@ -13,6 +13,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"math"
 	"math/big"
 	"net/http"
 	"slices"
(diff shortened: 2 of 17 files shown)


rawdb/rawdb.go
```diff
@@ -165,6 +166,14 @@ func (r *RawDB) Insert(key, value []byte) error {
 		} else {
 			// Encoded coordinates.
 			c := make([]byte, 4+4+4)
+
```
marcopeereboom, line 169: I have issues with this fake test. We either have a max size or not. This fixes no problems.

service/popm/popm.go
```diff
@@ -299,6 +300,10 @@ func pickUTXO(utxos []*bfgapi.BitcoinUTXO, amount int64) (*bfgapi.BitcoinUTXO, e
 }
 
 func createTx(l2Keystone *hemi.L2Keystone, btcHeight uint64, utxo *bfgapi.BitcoinUTXO, payToScript []byte, feeAmount int64, minRelayTxFee int64) (*btcwire.MsgTx, error) {
+	if btcHeight > math.MaxUint32 {
```
marcopeereboom, line 303: this is also dumb

marcopeereboom (changes requested):
I went through this and a bunch of these changes help or do nothing while making the code slower and more unreadable. If we want maxima that's fine but checking for uint32 max isn't it.

.golangci.yaml
```diff
@@ -33,7 +33,7 @@ linters:
     - "gofumpt" # Runs gofumpt.
     - "goheader" # Enforces copyright header.
     # TODO enable gosec, when bored run 'golangci-lint run -Egosec' to fix these
```
joshuasing, line 35: I think we could remove this comment now.

.golangci.yaml
```diff
@@ -88,6 +88,11 @@ linters-settings:
     replace-allow-list:
       - "github.com/coder/websocket" # Currently uses our fork with a bug fix.
 
+  # Checks for potential security issues
+  gosec:
+    excludes:
+    - G115 # Potential integer overflow when converting between integer types
```
joshuasing, line 94: This is valid YAML, but to keep it consistent with the rest of the file I think there should be an extra indent here:
```suggestion
      - G115 # Potential integer overflow when converting between integer types
```

joshuasing (approved):
Couple of comments, otherwise looks good.

.golangci.yaml
AL-CT: Fixed 👍

joshuasing (approved):

joshuasing:
Closing this for now, hopefully we can add these in the future.

PR #433, opened 2025-03-11
tbc: convert hash pointers to hashes as variable arguments

marcopeereboom:
**Summary**
Turns out some callers send in nil and play with the underlying data so we cannot reliably use pointers to hashes. Convert them to stack variables. There is no real impact on performance at the cost of some stack space but this does harden the call stack considerably.

**Changes**

### DIFF
--- a/database/tbcd/level/level.go
+++ b/database/tbcd/level/level.go
@@ -1503,7 +1503,7 @@ func (l *ldb) BlockExistsByHash(ctx context.Context, hash chainhash.Hash) (bool,
 		if errors.Is(err, leveldb.ErrNotFound) {
 			return false, nil
 		}
-		return false, fmt.Errorf("block exists: %w", err)
+		return false, fmt.Errorf("check block exists: %w", err)
 	}
 	return ok, nil
 }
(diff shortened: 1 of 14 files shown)


database/tbcd/level/level.go
```diff
@@ -1486,6 +1486,28 @@ func (l *ldb) BlockByHash(ctx context.Context, hash *chainhash.Hash) (*btcutil.B
...
+		}
+	}
+
+	bDB := l.rawPool[level.BlocksDB]
+	ok, err := bDB.Has(hash[:])
+	if err != nil {
+		if errors.Is(err, leveldb.ErrNotFound) {
+			return false, nil
+		}
+		return false, fmt.Errorf("block exists: %w", err)
```
joshuasing, line 1506: "block exists:" makes it sound like the block existing is the error, which isn't the case here - could we make this clearer:
```suggestion
		return false, fmt.Errorf("check block exists: %w", err)
```

joshuasing (changes requested):

AL-CT (approved):

PR #434, opened 2025-03-11
ttl: fix typos in comments

Pricstas:
- "callback functions" → "callback function" (singular)
- "key value pair" → "key-value pair" (hyphenated)
- "Len return" → "Len returns" (grammar fix)

### DIFF
--- a/ttl/ttl.go
+++ b/ttl/ttl.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2024 Hemi Labs, Inc.
+// Copyright (c) 2024-2025 Hemi Labs, Inc.
 // Use of this source code is governed by the MIT License,
 // which can be found in the LICENSE file.
 


ttl/ttl.go
```diff
@@ -87,7 +87,7 @@ func (tm *TTL) ttl(ctx context.Context, key any) {
 }
 
 // Put inserts the provided key and value into the TTL map. The ttl values
-// designates the duration of the validity of this key value pair. The expired
+// designates the duration of the validity of this key value-pair. The expired
```
joshuasing, line 90: I don't believe "value-pair" is correct, however key value should be written as "key-value":
https://en.wikipedia.org/wiki/Name%E2%80%93value_pair
```suggestion
// designates the duration of the validity of this key-value pair. The expired
```

joshuasing (changes requested):

marcopeereboom (approved):

AL-CT (approved):
LGTM

joshuasing (approved):
Thank you!

PR #435, opened 2025-03-11
tbc: fix race condition in fork tests

AL-CT:
**Summary**
Fixes a [race condition ](https://github.com/hemilabs/heminetwork/actions/runs/13774032146/job/38519089244) that occurs due to multiple blockheader messages being sent to TBC before it has time to read them.

**Changes**
Fix slight error in locking mechanism for rawpeer conn writes, and added a test that should make it easier to identify hidden race conditions such as this one.

### DIFF
--- a/service/tbc/peer/rawpeer/rawpeer.go
+++ b/service/tbc/peer/rawpeer/rawpeer.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2024 Hemi Labs, Inc.
+// Copyright (c) 2024-2025 Hemi Labs, Inc.
 // Use of this source code is governed by the MIT License,
 // which can be found in the LICENSE file.
 
@@ -110,8 +110,8 @@ func (r *RawPeer) Id() int {
 
 func (r *RawPeer) Write(timeout time.Duration, msg wire.Message) error {
 	r.mtx.Lock()
+	defer r.mtx.Unlock()
 	conn := r.conn
-	r.mtx.Unlock()
 	if conn == nil {
 		return fmt.Errorf("write: %w", ErrNoConn)
 	}
--- a/service/tbc/peer/rawpeer/rawpeer_test.go
+++ b/service/tbc/peer/rawpeer/rawpeer_test.go
@@ -53,7 +53,7 @@ func mockPeerServer(ctx context.Context, id int, listener net.Listener, msgCh ch
 	}
 }
 
-func TestRawpeerReadWrite(t *testing.T) {
+func TestConcurrentReadWrite(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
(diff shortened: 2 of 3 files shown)


marcopeereboom (changes requested):
Let's fix this for real. Why is the write stepping on each other?

We need a mutex somewhere.

service/tbc/peer/rawpeer/rawpeer.go
```diff
@@ -110,8 +110,8 @@ func (r *RawPeer) Id() int {
 
 func (r *RawPeer) Write(timeout time.Duration, msg wire.Message) error {
 	r.mtx.Lock()
+	defer r.mtx.Unlock()
 	conn := r.conn
-	r.mtx.Unlock()
```
marcopeereboom, line 114: There is no way this is right compared to the defer. With the defer we execute everything in the same shot. With this, we conn can and will be intertwined. We need more diagnostics on this.

marcopeereboom (changes requested):

service/tbc/peer/rawpeer/rawpeer.go
marcopeereboom: Derp, I read that exactly the opposite way.

marcopeereboom (approved):
This PR adheres to the readers read, writers write idiom. I recall reasoning through that when I wrote it and argued the long mutex away in writes. That was a mistake in my end. Apologies to the poor CI that suffered through this.

ClaytonNorthey92 (approved):
nice! thanks for handling this

service/tbc/peer/rawpeer/rawpeer_test.go
```diff
@@ -4,4 +4,133 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err := p1.Write(0, wire.NewMsgPing(1)); err != nil {
+		panic(err)
+	}
+
+	writerNum := 2
+	WriteCount := 100
```
joshuasing, line 93: nit: I think these could be constants - Also, is there any reason `WriteCount` starts with a capitalised letter? 🤔 
```suggestion
	const (
		writeNum = 2
		writeCount = 100	
	)
```

service/tbc/peer/rawpeer/rawpeer_test.go
```diff
@@ -4,4 +4,133 @@
...
+
+	fakeCh := chainhash.Hash{}
```
joshuasing, line 95: nit: It is usually more idiomatic to write empty/zero-value variables as:
```
var fakeCh chainhash.Hash
```

joshuasing (approved):
Nice catch! Two nit comments, both non-blocking.

service/tbc/peer/rawpeer/rawpeer_test.go
AL-CT: Yep makes sense. And no, just an accident 😄

service/tbc/peer/rawpeer/rawpeer_test.go
AL-CT: Gotcha! Fixed 👍

PR #443, opened 2025-03-13
optimize(rawdb):use t.TempDir replace os.MkdirTemp to simplify test temp dir manager

healthyyyoung:
**Summary**

**Changes**


marcopeereboom:
While I am sensitive to this change we are not going to merge this. The remove flag, in this form, exists to aid during debug where the directory must remain behind. Thanks for the PR though, appreciate it.

PR #450, opened 2025-03-19
popm: don't instantly noop if bfg response channel is busy

### LINKED ISSUE #422: popm, bfg: pop miner will timeout if bfg sends responses too fast
### Confirmation

### Describe the bug

As discussed in #421, pop miner appears to have a potential race condition when it sends a request to bfg during keystone mining. If bfg sends a response before the pop miner has begun listening for it, the miner will get stuck until timing out (after which it will try to mine a keystone again), regardless of how many times you retry sending the response.

### Expected behaviour

The miner should be able see the responses, either ones sent to the conn before they were listening, or retries sent after the fact.

### Environment

_No response_

### Additional Information

This is rare, as network overhead for BFG will make it very unlikely for this to happen. It may also be hard to spot in real-world scenarios, as the pop miner will simply time out that mining attempt and start a new one.
Finally, the BFG code will likely receive a big overhaul, so this might get fixed as a consequence.

AL-CT:
**Summary**
Fix #422 

**Changes**
Prior, whenever the channel that would receive bfg responses in pop miner was busy, we would no-op. this happened often because the channel was not buffered. this would happen silently as well. add a timeout to sending messages to the non-buffered channel and log and error if we time out

### DIFF
--- a/service/popm/popm.go
+++ b/service/popm/popm.go
@@ -742,7 +742,7 @@ func (m *Miner) handleBFGCallCompletion(parrentCtx context.Context, conn *protoc
 	select {
 	case bc.ch <- payload:
 		log.Tracef("handleBFGCallCompletion returned: %v", spew.Sdump(payload))
-	case <-time.After(DefaultBFGRequestTimeout):
+	case <-time.After(m.requestTimeout):
 		log.Errorf("handleBFGCallCompletion: response time out %v", cmd)
 	}
 }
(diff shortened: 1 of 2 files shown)


ClaytonNorthey92 (approved):

service/popm/popm.go
```diff
@@ -742,7 +742,8 @@ func (m *Miner) handleBFGCallCompletion(parrentCtx context.Context, conn *protoc
 	select {
 	case bc.ch <- payload:
 		log.Tracef("handleBFGCallCompletion returned: %v", spew.Sdump(payload))
-	default:
+	case <-time.After(m.requestTimeout):
```
ClaytonNorthey92, line 745: hey @AL-CT one more thing we might want to add here is a `case` for the context being "done" 

perhaps something like (double check this)
```go
case <-ctx.Done():
```

not a big deal, since we will timeout anyway, but we probably should have this

joshuasing (approved):
Awesome!

PR #451, opened 2025-03-20
Localnet update

ClaytonNorthey92:
**Summary**
update localnet to latest hashes, also replace pop point address with governance token address as we have it temporarily set to pop points but there is no way to configure this in optimism.  so let's replace it in the file until it's changed back

**Changes**
* update op-geth and op-node
* find-and-replace rewards address


jcvernaleo (approved):
OK

AL-CT (approved):

PR #456, opened 2025-03-24
BFG: save db records in processBitcoinBlock in a single tx

ClaytonNorthey92:
**Summary**
when saving records from processing bitcoin blocks, prior they were not done atomically (in a db tx).  this means that if one failed others may succeed.  this change makes it all-or-nothing; we either save everything we want from a btc block or we don't, so we don't half-save.

**Changes**
see summary

### DIFF
--- a/database/bfgd/database.go
+++ b/database/bfgd/database.go
@@ -6,6 +6,7 @@ package bfgd
 
 import (
 	"context"
+	"database/sql"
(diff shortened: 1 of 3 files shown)


marcopeereboom (approved):

AL-CT (approved):
LGTM

PR #461, opened 2025-03-25
tbc: add bound paralell outpoint fetch fixup strategy.

marcopeereboom:
The batched fixup strategy is too slow on cloud VMs and the unbound parallel strategy spawns too many threads when used in conjunction with op-geth. This commit maxes out the number of parallel fetch requests to 128 per block.

**Summary**

**Changes**

### DIFF
--- a/service/tbc/crawler.go
+++ b/service/tbc/crawler.go
@@ -490,8 +490,16 @@ func (s *Server) unprocessUtxos(ctx context.Context, txs []*btcutil.Tx, utxos ma
 	return nil
 }
 
-func (s *Server) fetchOPParallel(ctx context.Context, w *sync.WaitGroup, op tbcd.Outpoint, utxos map[tbcd.Outpoint]tbcd.CacheOutput) {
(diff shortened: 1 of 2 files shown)


AL-CT (approved):
LGTM

service/tbc/crawler.go
```diff
@@ -590,6 +598,56 @@ func (s *Server) fixupCacheBatched(ctx context.Context, b *btcutil.Block, utxos
...
+	slots := 128
+	c := make(chan struct{}, slots)
+	defer close(c)
+	for i := 0; i < slots; i++ {
+		select {
+		case <-ctx.Done():
+			return nil
+		case c <- struct{}{}:
+		default:
+			return fmt.Errorf("shouldn't happen")
```
joshuasing, line 612: Not sure why the linter didn't catch this...
```suggestion
			return errors.New("shouldn't happen")
```

marcopeereboom: i swear i will write a benchmark one day to show this isn't worth it :-P

ClaytonNorthey92 (approved):

PR #477, opened 2025-03-31
localnode: remove bfgd as a dependency of bssd

ClaytonNorthey92:
**Summary**
bssd can TECHNICALLY run without bfgd, so don't declare it as a dependency.  this makes it easier to run HEMI-only nodes

**Changes**
see summary


jcvernaleo (approved):
OK

joshuasing (approved):

PR #479, opened 2025-04-02
Improved Documentation, New `hemi-min` Profile for Lighter Nodes, Updated Docker Images, Misc. Config Updates

max-sanchez:


localnode/docker-compose.yml
```diff
@@ -235,9 +235,9 @@ services:
   # Hemi op-node
   op-node:
     # yamllint disable-line rule:line-length
-    image: "ghcr.io/hemilabs/op-node:ee580dd@sha256:c3c2bcabb0da02e6aca462a2756ef78d8a40e569281de1f4dac9a9eb8fa80359"
+    image: "ghcr.io/hemilabs/op-node:e79d444@sha256:c2757007112d81c2db1f0ad52cf12e1a3f2bf8313df434d9533abbf29b8b9d04"
```
ClaytonNorthey92, line 238: let's use the latest op-node, ghcr.io/hemilabs/op-node:9dc3004@sha256:fbbed98c8d9f8b1de1eb426b84b38c5df4fe2e04e9c757f4734dc1009d253b3e

localnode/docker-compose_mainnet.yml
```diff
@@ -233,9 +233,9 @@ services:
   # Hemi op-node
   op-node:
     # yamllint disable-line rule:line-length
-    image: "ghcr.io/hemilabs/op-node:84b895e@sha256:df29c59680c33cdda616f67662c9e75bdeea60f79d91937ee0b60bf0d0f48243"
+    image: "ghcr.io/hemilabs/op-node:e79d444@sha256:c2757007112d81c2db1f0ad52cf12e1a3f2bf8313df434d9533abbf29b8b9d04"
```
ClaytonNorthey92: same as other comment, 
```
ghcr.io/hemilabs/op-node:9dc3004@sha256:fbbed98c8d9f8b1de1eb426b84b38c5df4fe2e04e9c757f4734dc1009d253b3e
```

localnode/docker-compose_mainnet.yml
```diff
@@ -266,6 +266,10 @@ services:
...
+      - "--p2p.static=/ip4/34.90.21.246/tcp/9003/p2p/16Uiu2HAmNGgNTgiFBAqH58aCT3iXWMnetYMtQgH21Ydjq2R7QRbt"
+      - "--p2p.ban.peers=false"
```
ClaytonNorthey92, line 270: does this mean that _we never ban peers in mainnet_? is that what we want? 🤔

ClaytonNorthey92: I guess a user could always change this....

localnode/gen-files.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
```
joshuasing, line 1: Ideally, I think it would be best if we could continue using `/bin/sh` and make the script POSIX compliant for compatibility on as many platforms as possible. I can do this if you want.

joshuasing (changes requested):

joshuasing (approved):

localnode/docker-compose_mainnet.yml
max-sanchez: op-node does some aggressive node pruning and generates a lot of false positives, this is something I'd like to remove eventually after we improve this logic.

jcvernaleo (approved):
OK

PR #482, opened 2025-04-09
Update popm.go

### LINKED ISSUE #481: minRelayTxFee to calculate IsDust is too high
### Confirmation

### Describe the bug

The minrelayfee setting to calculate if an output should be consider dust is too high. This should be set to 1,000 instead of 10,000. Because of this an output with a value of 5460 or less is already seen as dust instead of 546 or less. 

For example this transaction is paying too much: https://mempool.space/nl/tx/0fa95415a6dfdeae278d6a6abce5a18cf5e2a9b343ea46d0a1cd5943f18e9196

### Expected behaviour

Set the minrelayfee to calculate dust to 1000.

### Environment

Version: 
Operating System: 
Architecture: 

### Additional Information

_No response_

Leonardiae:
The minrelayfee has been set to high causing outputs being marked as dust too soon.

**Summary**
Change the minrelayfee to calculate dust from 10,000 to 1,000.

"Fixes #481 "

### DIFF
--- a/service/popm/popm.go
+++ b/service/popm/popm.go
@@ -383,7 +383,7 @@ func (m *Miner) mineKeystone(ctx context.Context, ks *hemi.L2Keystone) error {
 	}
 
 	// Build transaction.
-	btx, err := createTx(ks, btcHeight, utxo, payToScript, feeAmount, 10000)
+	btx, err := createTx(ks, btcHeight, utxo, payToScript, feeAmount, 1000)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


joshuasing (approved):

AL-CT (approved):
Thanks!

ClaytonNorthey92 (approved):

joshuasing (approved):
Thank you!

PR #486, opened 2025-04-14
tbc: fix fork tests failing if tbc is slow to insert blocks

AL-CT:
**Summary**
Inside our fork tests, we send blocks to TBC and then check the db for tests. However, TBC might take longer than expected to insert the blocks, so we must wait for the blocks to be inserted before checking the db.

**Changes**
Explicitly check that all blocks were inserted before proceeding with tests.

### DIFF
--- a/service/tbc/tbcfork_test.go
+++ b/service/tbc/tbcfork_test.go
@@ -1282,7 +1282,7 @@ func TestFork(t *testing.T) {
 	}
 
 	// Wait for tbc to insert all blocks
-	hasBlocks := false
+	var hasBlocks bool
 	for !hasBlocks {
 		hasBlocks, err = s.hasAllBlocks(ctx, n.blocksAtHeight)
 		if err != nil {
@@ -1484,7 +1484,7 @@ func TestIndexNoFork(t *testing.T) {
 	}
 
 	// Wait for tbc to insert all blocks
-	hasBlocks := false
+	var hasBlocks bool
 	for !hasBlocks {
 		hasBlocks, err = s.hasAllBlocks(ctx, n.blocksAtHeight)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


service/tbc/tbcfork_test.go
```diff
@@ -2748,6 +2811,16 @@ func TestForkCanonicity(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// Wait for tbc to insert all blocks
+	hasBlocks := false
```
ClaytonNorthey92, line 2815: nit: since this functionality is repeated a few times, it could be a function

ClaytonNorthey92 (approved):

service/tbc/tbcfork_test.go
AL-CT: The actual checking is inside a function, but decided to leave part of it outside and check the error that function returns so the code is less messy in breaking out of the outer for loop, and to potentially prevent logging in T after the test fails from the context timeline exceeding (which would result in a panic).

Still, if you think it's best, I can probably change it a bit so that that is addressed, just might be a little less readable at a glance 👍

service/tbc/tbcfork_test.go
```diff
@@ -1268,6 +1280,17 @@ func TestFork(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	// Wait for tbc to insert all blocks
+	hasBlocks := false
```
joshuasing, line 1285: I think this can be simplified to (and for below), unless there is a reason not to:
```suggestion
	var hasBlocks bool
```

Go seems to generally prefer using the empty value vs doing it like this, but they both have the same behaviour. I usually prefer empty, but we mix both a lot so it's up to you :)

joshuasing (approved):

joshuasing:
Thanks for debugging this :D

PR #487, opened 2025-04-14
all: enable linting test sources

joshuasing:
**Summary**
Previously, we were not linting any test files. This enables linting test sources, and fixes all linting issues currently reported for the tests.

**Changes**
- Enable golangci-lint linting test sources
- Ignore `websocket.Close` for the `bodyclose` linter, as closing the response body is handled by the websocket library.
- Fix issues reported by golangci-lint for tests (mostly duplicated packages, unused code, etc.)

There were a few ineffective assignments, two of which I was unable to determine why they were there - they both have `nolint` comments. Please let me know if you think they should be removed or kept.

I have commented out unused functions, however I think removing them would be best if we do not plan on using them in the near future - @marcopeereboom?

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,6 +1,6 @@
 version: "2"
 run:
-  tests: false
+  tests: true
   go: "1.23"
 
 issues:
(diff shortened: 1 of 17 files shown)


marcopeereboom (approved):
Some of these linters are hokey.

AL-CT (approved):
LGTM

PR #492, opened 2025-04-22
tbc: avoid race with r.index on Close

marcopeereboom:
**Summary**
Saw a panic on rad database close close for using `r.index` after it was reset to nil. Use a bool to indicate if a db is open instead of relying on r.index thus removing the possibility of a race.

**Changes**


joshuasing (approved):

ClaytonNorthey92 (approved):

jcvernaleo (approved):
OK

PR #494, opened 2025-04-30
fix escaping quotes that were re-recreating geth datadir

ClaytonNorthey92:
**Summary**
prior, the way we were escaping quotes was preventing the entrypoint.sh file to recognize an existing datadir, fix those

**Changes**
see summary


jcvernaleo (approved):
OK

marcopeereboom (approved):

joshuasing (approved):

PR #502, opened 2025-05-10
refactor: use the built-in max/min to simplify the code

eveneast:
**Summary**

In Go 1.21, the standard library includes built-in [max/min](https://pkg.go.dev/builtin@go1.21.0#max) function, which can greatly simplify the code.

**Changes**


marcopeereboom:
Thanks but after internal debate we have decided that the `-4` gets lost in readability when using min. Appreciate the pr though.

PR #507, opened 2025-05-19
add a non-sequencing node to localnet and test it

### LINKED ISSUE #508: Add non-sequencing test cases to localnet tests
### Confirmation

### Problem

As of now, localnet only runs a single sequencer.  We're missing test cases for non-sequencing nodes.

### Suggested solution

talking to @max-sanchez , we want to add the following:

* Non-sequencing and sequencing nodes stay at the same tip (could have a check that the non-sequencing nodes are never more than 2 blocks behind, and their tip matches the equivalent block hash from the sequencer)
* Non-sequencing and sequencing nodes have the same hVM tip when they are at the same L2 block tip (you can call the 0x44 precompile which will return the hVM indexed tip)
* Non-sequencing and sequencing nodes have the same state root (optimism_outputAtBlock) for the entire chain
* Non-sequencing and sequencing nodes both are able to re-derive the chain from L1 (which means the "safe" and "final" blocks should progress)
* Transactions sent to any of the non-sequencing op-geth nodes correctly propagate over op-geth P2P and get sequenced into blocks within <X> (20-30) seconds.

### Additional Information

_No response_

ClaytonNorthey92:
**Summary**
added test cases for the following:
* assert that non-sequencing tip is close to the sequencer
* assert that non-sequencing hvm btc tip is the same as the sequencer
* walk the l2 chain and assert that the output root is the same all of
  the way down, ignore sync status
* assert that safe and final blocks are progressing
* repeat tests for both sequencer and non-sequencer (deploying
  contracts, sending txs, bridging, etc)

fixes #508 

**Changes**
See summary.


e2e/monitor/main_test.go
```diff
@@ -1130,5 +1207,148 @@ func waitForTxReceipt(t *testing.T, ctx context.Context, client *ethclient.Clien
...
+
+		resBody2, err := io.ReadAll(res2.Body)
+		if err != nil {
+			t.Fatalf("error reading response body from non-sequencer: %s", err)
+		}
+
+		res.Body.Close()
+		res2.Body.Close()
+
+		t.Logf("will parse reponses\n%s\n%s", string(resBody), string(resBody2))
```
ClaytonNorthey92: this is a bit noisy, probably not necessary

jcvernaleo (approved):
OK

max-sanchez (approved):

PR #510, opened 2025-05-21
Disregard keystones that are too far in the future

ClaytonNorthey92:
**Summary**
if someone broadcasts an invalid keystone, specifically one that with a very high l2 block number, all finalities for lower keystones will be incorrec.

add a configuration for a "baseline l2 block" to calculate the expected now + 10 minute block height.  if we have keystones above that, do not include them in finality calculations.>

**Changes**
see summary


service/bfg/bfg.go
```diff
@@ -1474,6 +1477,30 @@ func (s *Server) refreshCacheAndNotifiyL2Keystones(pctx context.Context) {
...
+
+	// set cut off 10 minutes in the future
+	// calculate the number of expected blocks between then and our baseline
+	// block
+
+	cutoff := time.Now().Add(10 * time.Minute).Unix()
+	expectedTimeElapsedInBlocks := (cutoff - s.cfg.BaselineL2BlockTimestamp) / hemi.L2BlockTimeSeconds
+	expectedHighestBlock := s.cfg.BaselineL2BlockHeight + expectedTimeElapsedInBlocks
+
+	log.Tracef("the time is cutt off at %d, we expected %d to be the highest block", cutoff, expectedHighestBlock)
```
jcvernaleo, line 1495: I think you've got too many letters in `cut`

jcvernaleo (changes requested):
One super minor typo.

service/bfg/bfg.go
ClaytonNorthey92: ah yep...fixing

jcvernaleo (approved):
OK

max-sanchez (approved):

PR #515, opened 2025-05-23
do not include invalid keystones in effectiveHeight

ClaytonNorthey92:
**Summary**
when calculating effectiveHeight, we're not taking into account the ignore (we only do so when querying for blocks), add that to the effectiveHeight calculation

add test case for this

**Changes**
make sure that the invalid keystones are taken into account with effectiveHeight

### DIFF
--- a/database/bfgd/database_ext_test.go
+++ b/database/bfgd/database_ext_test.go
@@ -1707,7 +1707,6 @@ func TestL2BtcFinalitiesByL2KeystoneWithCutoff(t *testing.T) {
 }
 
 func TestL2BtcFinalitiesByL2KeystoneWithCutoffReverseMining(t *testing.T) {
-
 	// if a newer keystone gets mined before gets mined before an older
 	// one, take that into account with effective height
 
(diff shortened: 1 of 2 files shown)


database/bfgd/database_ext_test.go
```diff
@@ -1700,6 +1700,54 @@ func TestL2BtcFinalitiesByL2KeystoneWithCutoff(t *testing.T) {
...
 		t.Fatalf("unexpected diff %s", diff)
 	}
+
+	if finalities[0].EffectiveHeight != 8987 {
+		t.Fatalf("incorrect effective height %d", finalities[0].EffectiveHeight)
+	}
+}
+
+func TestL2BtcFinalitiesByL2KeystoneWithCutoffReverseMining(t *testing.T) {
+	// if a newer keystone gets mined before gets mined before an older
```
jcvernaleo, line 1710: I think so e diplicate words

jcvernaleo (changes requested):
Minor comment issue, ortherwise looks good

database/bfgd/database_ext_test.go
ClaytonNorthey92: ah thanks

jcvernaleo (approved):
ok

max-sanchez (approved):

PR #516, opened 2025-05-23
bfg window from tip vs starting block

ClaytonNorthey92:
**Summary**
bfg doesn't need to store a lot of the btc chain; it just needs enough to fetch pop payouts and bitcoin finality.

**Changes**
allow a configurable "window-from-tip" to be set in bfg, this is how many blocks it keeps track of behind tip.  this should allow us to keep our current queries as the inefficiency comes from the large amount of data stored that we do not use.

simplified the query a bit


ClaytonNorthey92:
Queryrow->queryrowcontext

PR #396, opened 2025-02-05
all: rewrite bfgd and popmd, remove bssd

AL-CT:
**Summary**
Now that we have arrived at mainnet it is time to pay off the technical debt we incurred. For all the right reasons, bfg became a dumping ground for functionality our users needed. TBC has caught up and implements the primitives that bfg needs so that it can drastically shrink in scope and functionality.

**Changes**
* Remove bssd altogether. Having it at the time made sense but that has gone away as the code progressed. PoP payouts have moved to op-geth.
* Implement keystone tracking in tbc.
* Implement packages that handle bitcoin wallet functionality.
* Implement various backends for bitcoin knowledge in bfg.
* Rewrite pop miner to use new infrastructure.

Screenshot 2025-06-05 at 16 28 07

Screenshot 2025-06-05 at 16 28 17

Screenshot 2025-06-05 at 16 28 22

Screenshot 2025-06-05 at 16 28 30

Screenshot 2025-06-05 at 16 28 36

Screenshot 2025-06-05 at 16 30 23

PR #521, opened 2025-05-29
tbcd: implement testnet4

marcopeereboom:
**Summary**

**Changes**


jcvernaleo:
Should this be marker as draft maybe until #396 is in and it gets rebased?

joshuasing:
Need to rebase once #396 is merged.

database/tbcd/level/upgrade.go
```diff
@@ -348,7 +348,7 @@ func (l *ldb) v3(ctx context.Context) error {
 	// First let's make sure we have enough disk space to process this.
 	var need uint64
 	switch l.cfg.Network {
-	case "testnet3":
+	case "testnet3", "testnet4":
```
ClaytonNorthey92: as of now, the readme says:
> **`testnet4`** requires approximately 10 GiB of disk space.

but this states that we need `100 GiB`, I think we should make these consistent

service/popm/popm.go
```diff
@@ -82,15 +82,12 @@ type Config struct {
...
-		Network:                "testnet3",
-		PrometheusNamespace:    appName,
-		OpgethURL:              defaultOpgethURL,
-		BitcoinConfirmations:   defaultBitcoinConfirmations,
-		BitcoinSource:          bitcoinSourceTBC,
-		BitcoinURL:             tbcgozer.DefaultURL,
-		opgethReconnectTimeout: defaultOpgethReconnectTimeout,
-		l2KeystoneMaxAge:       defaultL2KeystoneMaxAge,
-		l2KeystonePollTimeout:  defaultL2KeystonePollTimeout,
+		Network:              "mainnet",
```
ClaytonNorthey92, line 85: is `mainnet` supposed to be the default?

marcopeereboom: Yes. This was flipped around in all other daemons.

database/tbcd/level/upgrade.go
marcopeereboom: Fixed

cmd/tbcd/README.md
```diff
@@ -17,6 +17,7 @@ endpoint.**
 **As of June 2025:**
```
joshuasing, line 17: It is now July.

```suggestion
**As of July 2025:**
```

marcopeereboom: Meh. It was accurate in June 2025 and is still accurate.

ClaytonNorthey92 (approved):

PR #523, opened 2025-05-30
docs: update ubuntu version

gap-editor:
In NODE_RUNNING.md, there's a reference to "Ubuntu 24.04" which doesn't exist yet (the latest LTS is 22.04). This should be corrected.


jcvernaleo:
24.04 is the current latest LTS
https://ubuntu.com/blog/tag/ubuntu-24-04-lts

PR #529, opened 2025-06-03
all: update go to 1.24 (toolchain 1.24.4)

joshuasing:
**Summary**
Update Go version to `1.24` (and toolchain `go1.24.4`).

**Changes**
- Update required go version in `go.mod` to `1.24`.
- Update go toolchain version in `go.mod` to `1.24.4`.
- Update golang Docker images to `1.24.4-alpine3.22` and `1.24.4-bookworm`.
- Update tests to use `(*testing.T).Context()` instead of `context.Background()` or `context.TODO()`.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,7 +1,6 @@
 version: "2"
 run:
   tests: true
-  go: "1.23"
 
 issues:
   max-issues-per-linter: 0
--- a/api/auth/secp256k1_test.go
+++ b/api/auth/secp256k1_test.go
@@ -5,7 +5,6 @@
 package auth
 
 import (
-	"context"
 	"crypto/sha256"
 	"encoding/hex"
 	"net/http"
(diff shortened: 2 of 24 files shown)


.golangci.yaml
```diff
@@ -1,7 +1,6 @@
 version: "2"
 run:
   tests: true
-  go: "1.23"
```
jcvernaleo, line 4: Are we okay not specifying a version here at all?

joshuasing: I think this is okay. - `golangci-lint` will use the Go version from `go.mod` by default. We had to set this as the toolchain version was previously set to `go1.24` but we used `go 1.23`.

Per https://golangci-lint.run/usage/configuration/#run-configuration:
```
  # Define the Go version limit.
  # Default: use Go version from the go.mod file, fallback on the env var `GOVERSION`, fallback on 1.22.
  go: '1.23'
  ```

jcvernaleo: Using the version from go.mod is actually even better than explicitly setting it in my opinion.

jcvernaleo (approved):
OK

ClaytonNorthey92 (approved):

PR #533, opened 2025-06-04
re-added e2e tests

ClaytonNorthey92:
**Summary**
re-added a much-smaller e2e_ext_test.go file. this only tests finality by l2 keystone abrev hash.  testing for less than superfinality, equal to superfinality, and greater than superfinality

**Changes**
see summary

### DIFF
--- a/service/bfg/bfg.go
+++ b/service/bfg/bfg.go
@@ -197,6 +197,7 @@ func (s *Server) callOpgeth(ctx context.Context, request any) (any, error) {
 			if err != nil {
 				return nil, fmt.Errorf("error calling opgeth: %w", err)
 			}
+
 			return &resp, nil
 		default:
 			return nil, fmt.Errorf("unknown opgeth command: %T", request)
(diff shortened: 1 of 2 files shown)


service/bfg/bfg.go
```diff
@@ -283,6 +285,12 @@ func (s *Server) handleKeystoneFinality(w http.ResponseWriter, r *http.Request)
 	// Cycle through each response and replace finality value for the best
 	// finality value of its descendants or itself
 	fin := &bfgapi.L2KeystoneBitcoinFinalityResponse{}
+
+	// shouldn't this always be one if we get here? since we query by the hash?
+	if len(resp.L2Keystones) > 0 {
+		fin.L2Keystone = resp.L2Keystones[0]
+	}
```
AL-CT: I think this is fine, but the finality value is going to be 0 regardless if it's not found in TBC, so does it make a difference to have the keystone set if all the other fields are at their 0 value? I think it's fine either way, but we might be able to infer more information if that is left empty (such as the fact that no keystone was posted to the BTC chain)

ClaytonNorthey92: hmm so I think that we _should_ fill out the keystone's fields if it's found in op-geth, but not tbc.  because it's a valid keystone (the keystone is _found_) but it isn't in the blockchain yet.

does that make sense?

PR #539, opened 2025-06-10
tbc: add v4 keystone height hash index

marcopeereboom:
**Summary**
Add height hash index to keystone index so that we can have a notion of what the last super duper giga final keystone is.


database/tbcd/level/level.go
```diff
@@ -1938,16 +1955,99 @@ func encodeKeystoneToSlice(ks tbcd.Keystone) []byte {
...
+	if int64(height)+int64(depth) > math.MaxUint32 {
+		return nil, errors.New("the overflow that matters")
+	}
+	if int64(height)+int64(depth) <= 0 {
+		return nil, errors.New("underflow")
+	}
+
+	kssDB := l.pool[level.KeystonesDB]
+	i := kssDB.NewIterator(keystoneHeightRange(height, depth), nil)
+	defer func() { i.Release() }()
```
joshuasing:
```suggestion
	defer i.Release()
```

database/tbcd/level/level.go
```diff
@@ -1938,16 +1955,99 @@ func encodeKeystoneToSlice(ks tbcd.Keystone) []byte {
...
+
+	kssList := make([]tbcd.Keystone, 0, 16)
+	for i.Next() {
+		_, hash := decodeKeystoneHeightHash(i.Key())
+		eks, err := kssDB.Get(hash[:], nil)
+		if err != nil {
+			// mismatch between heighthash and hash indexes
+			panic(fmt.Sprintf("data corruption: %v", err))
```
joshuasing:
```suggestion
			panic(fmt.Errorf("data corruption: %w", err))
```

database/tbcd/level/level.go
```diff
@@ -1938,16 +1955,105 @@ func encodeKeystoneToSlice(ks tbcd.Keystone) []byte {
...
+	}
+
+	kssDB := l.pool[level.KeystonesDB]
+	i := kssDB.NewIterator(keystoneHeightRange(start, d), nil)
+	defer func() { i.Release() }()
+
+	kssList := make([]tbcd.Keystone, 0, 16)
+	for i.Next() {
+		_, hash := decodeKeystoneHeightHash(i.Key())
+		eks, err := kssDB.Get(hash[:], nil)
```
ClaytonNorthey92, line 2041: if `i` is an iterator over `kssDB`, then we can use `i.Value()` here instead , that way we're not chasing a red herring on an error that shouldn't happen

database/tbcd/level/level.go
```diff
@@ -1938,16 +1955,105 @@ func encodeKeystoneToSlice(ks tbcd.Keystone) []byte {
...
+		if err != nil {
+			// mismatch between heighthash and hash indexes
+			panic(fmt.Sprintf("data corruption: %v", err))
```
ClaytonNorthey92: I would like to get rid of this error altogether, see other comment

database/tbcd/level/level.go
```diff
@@ -1938,16 +1955,105 @@ func encodeKeystoneToSlice(ks tbcd.Keystone) []byte {
...
+		deks := decodeKeystone(eks)
+		kssList = append(kssList, deks)
+	}
+	if len(kssList) == 0 {
+		return nil, database.NotFoundError(fmt.Sprintf("no first occurrence "+
+			"keystones range: %v < %v",
+			min(start, end), max(start, end)))
+	}
+
+	return kssList, i.Error()
```
ClaytonNorthey92: let's not return a non-nil `kssList` _and_ an error; that's a bit confusing.  let's explicitly check for `i.Error()` and return that if appropriate

service/tbc/testdata/testdatabasev3.tar.gz
ClaytonNorthey92, line 1: let's get rid of this; we should be able to define test data in tests themselves.  instead of this file, let's create v3 data in a test, then ensure that that data is upgraded

service/tbc/tbc_test.go
```diff
@@ -386,6 +386,99 @@ func TestDbUpgrade(t *testing.T) {
 	}
 }
 
+func TestDbUpgradeV4(t *testing.T) {
+	home := t.TempDir()
+	t.Logf("temp: %v", home)
+
+	err := extract("testdata/testdatabasev3.tar.gz", home)
```
ClaytonNorthey92: I put a comment on the file, but we should not be using a data file for this.  let's create data in the "v3" format in the test and ensure it gets upgrade.

service/tbc/rpc.go
```diff
@@ -640,6 +647,45 @@ func (s *Server) handleBlockInsertRawRequest(ctx context.Context, req *tbcapi.Bl
...
+		e := protocol.NewInternalError(err)
+		return &tbcapi.KeystonesByHeightResponse{
+			BTCTipHeight: bhb.Height,
+			Error:        e.ProtocolError(),
+		}, e
+	}
+
+	aks := make([]*hemi.L2KeystoneAbrev, 0, len(kssList))
+	for _, k := range kssList {
+		aks = append(aks,
```
ClaytonNorthey92: nit: I don't think that we need to `append` here if we already know the size (i.e. `len(kssList)`

service/tbc/tbc_test.go
ClaytonNorthey92: for example, I see 5 error return cases in the the upgrade function.  each of those should have its own test

database/tbcd/level/upgrade.go
```diff
@@ -546,3 +551,66 @@ func (l *ldb) v3(ctx context.Context) error {
...
+		value := i.Value()
+		if len(value) != keystoneSize {
+			if len(value) == keystoneSizeV3 {
+				// Keystone without height encoded.
+				var u [keystoneSize]byte
+				copy(u[4:], value[:])
+				value = u[:]
+			} else {
+				// Not a keystone.
+				continue
```
ClaytonNorthey92, line 581: isn't this the _keystones database_? are we ignoring an error here? shouldn't everything be a keystone here? or do we store different types in the same db?

if so, so we prefix those types at all? this silent continue feels like a footgun

ClaytonNorthey92 (changes requested):
the db upgrade test should be re-written; we're not testing individual error cases as far as I can tell. we should also get rid of that large test file. 

I also think that we're silencing possible large issues with a `continue`.

a few more things that I feel need addressing. I left comments

database/tbcd/level/level.go
AL-CT: In this new upgrade we store keystones using a key in the format of `[keystone_hash]`, as well as using a height hash key in the format of `H[block_height][keystone_hash]`. The height hash stores nothing, and instead we decode the key to get the hash of the keystone it points to.

database/tbcd/level/level.go
AL-CT: Related to previous comment

database/tbcd/level/upgrade.go
AL-CT: A few things:
1. This upgrade changes the size of keystones, so we check if their size is different
2. The new height hash can have an index but not value
3. I believe some entries always exist with no value but @marcopeereboom may have to confirm

With that said, I think adding a log here is probably safe though could be noisy.

database/tbcd/level/level.go
ClaytonNorthey92: ok I see now, disregard

database/tbcd/level/upgrade.go
ClaytonNorthey92: ok, this feels a bit odd to me, but I get it. let's not sweat it

database/tbcd/level/level.go
AL-CT: fixed 👍

database/tbcd/level/level.go
```diff
@@ -1938,16 +1955,109 @@ func encodeKeystoneToSlice(ks tbcd.Keystone) []byte {
...
+		panic("not a keystone height hash index")
+	}
+	height = binary.BigEndian.Uint32(v[1 : 1+4])
+	h := &hash
+	err := h.SetBytes(v[5:])
+	if err != nil {
+		panic(err)
+	}
+	return
+}
```
joshuasing: I think this can be made more readable by removing the named return values. In general, we should avoid named return values unless they are absolutely needed (e.g. err handling in defer).

Also inline the err assignment and check from hash.SetBytes, and use errors for panic instead of spew.

```suggestion
func decodeKeystoneHeightHash(v []byte) (uint32, chainhash.Hash) {
	if len(v) != keystoneHeightHashSize {
		panic(fmt.Errorf("invalid height hash size: %x", v))
	}
	if v[0] != 'h' {
		panic(fmt.Errorf("not a keystone height hash index: %x", v))
	}

	var hash chainhash.Hash
	if err := hash.SetBytes(v[5:]); err != nil {
		panic(err)
	}
	return binary.BigEndian.Uint32(v[1 : 1+4]), h
}
```

service/tbc/crawler.go
```diff
@@ -1375,7 +1376,14 @@ func (s *Server) TxIndexer(ctx context.Context, endHash chainhash.Hash) error {
 	return fmt.Errorf("invalid direction: %v", direction)
 }
 
-func processKeystones(blockHash *chainhash.Hash, txs []*btcutil.Tx, direction int, kssCache map[chainhash.Hash]tbcd.Keystone) error {
+func processKeystones(block *btcutil.Block, direction int, kssCache map[chainhash.Hash]tbcd.Keystone) error {
+	if block.Height() == btcutil.BlockHeightUnknown {
+		panic("diagnostic: block height not set")
+	}
```
joshuasing, line 1382: Was this added for debugging purposes? If so, should it be removed prior to merging?

service/tbc/rpc.go
joshuasing: As the length of the slice is `0`, we do need `append` here - however, I think it would be more readable and cleaner if we changed this to:
```
aks := make([]*hemi.L2KeystoneAbrev, len(kssList))
for i, k := range kssList {
	aks[i] = // ...
```

service/tbc/tbc.go
```diff
@@ -1735,6 +1735,19 @@ func (s *Server) BlockByHash(ctx context.Context, hash chainhash.Hash) (*btcutil
...
 }
 
+// KeystonesByHeight returns the first occurance found of keystones
+// at a given height + range. The given height is excluded.
+func (s *Server) KeystonesByHeight(ctx context.Context, height uint32, depth int) ([]tbcd.Keystone, error) {
+	log.Tracef("KeystonesByHeight")
+	defer log.Tracef("KeystonesByHeight exit")
+
+	if s.cfg.ExternalHeaderMode {
+		return nil, errors.New("cannot call KeystonesByHeight on TBC running in External Header mode")
```
joshuasing, line 1745: In the future, I think we should consider making an error type for "not allowed in external-header mode"

service/tbc/tbc_test.go
AL-CT: Added an extra test to validate the errors

service/tbc/rpc.go
AL-CT: Fixed 👍

database/tbcd/level/level.go
AL-CT: Fixed 👍

database/tbcd/level/level.go
AL-CT: Fixed 👍

database/tbcd/level/level.go
AL-CT: I agree. Fixed 👍

service/tbc/crawler.go
AL-CT: I believe this needs to remain here, as if the keystone does not have the height, the height hash index will not work properly, and things may break down the line.

service/tbc/tbc.go
AL-CT: I agree with this. Will make an issue to add this once everything is merged

database/tbcd/level/upgrade.go
```diff
@@ -364,14 +368,14 @@ func (l *ldb) v3(ctx context.Context) error {
...
-		}
-	}
+	//if !l.cfg.nonInteractive {
+	//	log.Infof("This operation will take a long time. " +
+	//		"Press ctrl-c within 30 seconds to abort upgrade!")
+	//	select {
+	//	case <-ctx.Done():
+	//	case <-time.Tick(30 * time.Second):
+	//	}
+	//}
```
joshuasing, line 378: Was this intentionally disabled?

joshuasing (approved):

database/tbcd/level/upgrade.go
marcopeereboom: no!

PR #547, opened 2025-06-13
fixed CI pipeline with test+code changes

ClaytonNorthey92:
**Summary**
see title

**Changes**

* run Go pipeline on any branch, so we can test prs from one non-main branch to another
* increase Gozer default request timeout
* make EnsureCanConnectTCP more robust by allow retries
* revert back to use FreePort
* increase timeouts of tests that needed it
* in popm, change package-scoped variables to be server-scoped and protect with mutex
* update TestTickingPopMiner to not rely on timing, but explicitly set the expiration to "now" and let pop miner expire the keystones
* a little more robust error handling
* some data race fixes, fixed with mutexes

### DIFF
--- a/e2e/e2e_ext_test.go
+++ b/e2e/e2e_ext_test.go
@@ -67,7 +67,8 @@ func EnsureCanConnect(t *testing.T, url string, timeout time.Duration) error {
 }
 
 func EnsureCanConnectTCP(t *testing.T, addr string, timeout time.Duration) error {
-	timeoutTicker := time.NewTicker(timeout)
+	ctx, cancel := context.WithTimeout(t.Context(), timeout)
+	defer cancel()
 
 	for {
 		select {
@@ -75,16 +76,13 @@ func EnsureCanConnectTCP(t *testing.T, addr string, timeout time.Duration) error
 			conn, err := net.DialTimeout("tcp", addr, 100*time.Millisecond)
 			if err != nil {
 				t.Logf("error dialing: %s", err)
-				conn.Close()
 				continue
-			} else {
-				conn.Close()
-				return nil
 			}
-		case <-t.Context().Done():
-			return t.Context().Err()
-		case <-timeoutTicker.C:
-			return fmt.Errorf("could not connect to tcp port: timeout")
+
+			conn.Close()
+			return nil
+		case <-ctx.Done():
+			return ctx.Err()
 		}
 	}
 }


AL-CT (approved):

e2e/e2e_ext_test.go
```diff
@@ -67,34 +67,38 @@ func EnsureCanConnect(t *testing.T, url string, timeout time.Duration) error {
...
 
-	conn.Close()
-	return nil
+	for {
+		select {
+		case <-time.After(100 * time.Millisecond):
+			conn, err := net.DialTimeout("tcp", addr, 100*time.Millisecond)
+			if err != nil {
+				t.Logf("error dialing: %s", err)
+				conn.Close()
```
marcopeereboom: this certainly is not needed

service/bfg/bfg_test.go
```diff
@@ -91,7 +91,7 @@ func TestBFG(t *testing.T) {
 	}()
 
 	// receive messages and errors from opgeth and tbc
-	if err = messageListener(ctx, expectedMsg, errCh, msgCh); err != nil {
+	if err = messageListener(t, ctx, expectedMsg, errCh, msgCh); err != nil {
```
marcopeereboom: you should not pass context here if you pass t

service/bfg/bfg_test.go
```diff
@@ -171,15 +171,15 @@ func TestKeystoneFinalityInheritance(t *testing.T) {
...
```
marcopeereboom: same

service/popm/popm.go
```diff
@@ -410,6 +415,7 @@ func (s *Server) hydrateKeystones(ctx context.Context) error {
 		s.mtx.Unlock()
 		return errors.New("already hydrated")
 	}
+	log.Infof("setting keystones hydrating to %d", len(keystones))
```
marcopeereboom: this is excessive logging

service/popm/popm.go
```diff
@@ -153,9 +153,11 @@ func NewServer(cfg *Config) (*Server, error) {
 	}
 
 	s := &Server{
-		cfg:            cfg,
-		retryThreshold: uint32(cfg.RetryMineThreshold) * hemi.KeystoneHeaderPeriod,
-		workC:          make(chan struct{}, 2),
+		cfg:                    cfg,
+		retryThreshold:         uint32(cfg.RetryMineThreshold) * hemi.KeystoneHeaderPeriod,
+		workC:                  make(chan struct{}, 2),
+		opgethReconnectTimeout: 5 * time.Second,
```
marcopeereboom: these should be const or variables, not set here as immutable constants.

service/popm/popm.go
```diff
@@ -545,6 +551,10 @@ func (s *Server) opgeth(ctx context.Context) {
 	log.Tracef("opgeth")
 	defer log.Tracef("opgeth exit")
 
+	s.mtx.Lock()
+	opgethReconnectTimeoutTmp := s.opgethReconnectTimeout
```
marcopeereboom: i am not following what this is for. The mutex is to prevent data races on maps etc, not for config values.

service/popm/popm_test.go
```diff
@@ -122,6 +117,11 @@ func TestTickingPopMiner(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// keystone expiration is forced below
+	s.mtx.Lock()
+	s.l2KeystoneMaxAge = mock.DefaultNtfnDuration * 10000000
```
marcopeereboom: I don't understand what this does. This needs to be either documented or made into a const that describes it.

marcopeereboom (approved):

PR #557, opened 2025-06-30
refactor: use the built-in max/min to simplify the code

mountdisk:
**Summary**

In Go 1.21, the standard library includes built-in [max/min](https://pkg.go.dev/builtin@go1.21.0#max) function, which can greatly simplify the code.

**Changes**


AL-CT:
Thank you for the PR! However, we have had a different PR suggest the same change [here](https://github.com/hemilabs/heminetwork/pull/502). As with the previous PR, we believe `-4` gets lost in readability when using min.

PR #564, opened 2025-07-05
fix typo in Update blockstream.go

moshderte:
This PR fixes a misleading error message in the Tip() function. The function uses strconv.ParseInt, but the error message previously said "parse uint", which is incorrect and could confuse during debugging.

Updated the error message from "parse uint" to "parse int" in Tip().


AL-CT (approved):
Thanks!

joshuasing (approved):
Thanks!

PR #565, opened 2025-07-06
cmd/keygen: remove duplicate flag.Parse call

moshderte:
This PR makes the following changes and improvements to the keygen command-line tool:

Removed duplicate flag.Parse() call in main()
Calling flag.Parse() more than once can cause a panic in newer Go versions (see flag.Parse() docs). The redundant second call was removed to prevent unexpected behavior.


AL-CT (approved):
Thanks!

joshuasing (approved):
Thanks!

PR #568, opened 2025-07-09
hproxy: a much simpler proxyd replacement

marcopeereboom:
**Summary**
proxyd tried to solve all problems and per the laws of physics, it made everything worse.

hproxy is a dead simple replacement that throws opgeth commands over the fence and maintains a list of concurrent nodes.

**Changes**
No changes, all greenfield.

### DIFF
--- a/service/hproxy/ethereum.go
+++ b/service/hproxy/ethereum.go
@@ -8,10 +8,10 @@ import (
 	"bytes"
 	"context"
 	"crypto/rand"
-	"encoding/binary"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
-	"io"
+	"fmt"
 	"net/http"
 )
 
@@ -27,24 +27,35 @@ var _ Proxy = EthereumProxy{}
 
 const EthereumVersion = "2.0"
 
-type EthereumCall struct {
-	Method  string `json:"method"`
-	Params  []any  `json:"params"`
-	ID      uint64 `json:"id"`
+// EthereumRequest is a JSON-RPC request object.
+// https://www.jsonrpc.org/specification
+type EthereumRequest struct {
 	Version string `json:"jsonrpc"`
+	Method  string `json:"method"`
+	Params  []any  `json:"params,omitempty"`
+	ID      any    `json:"id"`
 }
 
-func ethID() uint64 {
+// EthereumResponse is a JSON-RPC response object.
+// https://www.jsonrpc.org/specification
+type EthereumResponse struct {
+	Version string          `json:"jsonrpc"`
+	Result  json.RawMessage `json:"result"`
+	Error   any             `json:"error"`
+	ID      any             `json:"id"`
+}
+
+func ethID() string {
 	buf := make([]byte, 8)
 	_, err := rand.Read(buf)
 	if err != nil {
 		panic(err)
 	}
-	return binary.LittleEndian.Uint64(buf)
+	return hex.EncodeToString(buf)
 }
 
-func CallEthereum(ctx context.Context, c *http.Client, url, method string, params []any) (io.ReadCloser, error) {
-	ec := EthereumCall{
+func CallEthereum(ctx context.Context, c *http.Client, url, method string, params ...any) (*EthereumResponse, error) {
+	ec := EthereumRequest{
 		Method:  method,
 		Params:  params,
 		ID:      ethID(),
@@ -63,6 +74,7 @@ func CallEthereum(ctx context.Context, c *http.Client, url, method string, param
 	if err != nil {
 		return nil, err
 	}
+	defer resp.Body.Close()
 
 	switch resp.StatusCode {
 	case http.StatusOK:
@@ -70,7 +82,18 @@ func CallEthereum(ctx context.Context, c *http.Client, url, method string, param
 		return nil, errors.New(http.StatusText(resp.StatusCode))
 	}
 
-	return resp.Body, nil
+	var res EthereumResponse
+	if err := json.NewDecoder(resp.Body).Decode(&res); err != nil {
+		return nil, err
+	}
+	if res.Version != EthereumVersion {
+		return nil, fmt.Errorf("unexpected version: %s", res.Version)
+	}
+	if res.ID != ec.ID {
+		return nil, fmt.Errorf("invalid response id: %s instead of %s",
+			res.ID, ec.ID)
+	}
+	return &res, nil
 }
 
 func (e EthereumProxy) Poke(ctx context.Context) error {
--- a/service/hproxy/ethereum.go
+++ b/service/hproxy/ethereum.go
@@ -1,0 +1,83 @@
+// Copyright (c) 2025 Hemi Labs, Inc.
+// Use of this source code is governed by the MIT License,
+// which can be found in the LICENSE file.
+
+package hproxy
+
+import (
+	"bytes"
+	"context"
+	"crypto/rand"
+	"encoding/binary"
+	"encoding/json"
+	"errors"
+	"io"
+	"net/http"
+)
+
+type Proxy interface {
+	Poke(ctx context.Context) error
+}
+
+type EthereumProxy struct {
+	f func(ctx context.Context) error
+}
+
+var _ Proxy = EthereumProxy{}
+
+const EthereumVersion = "2.0"
+
+type EthereumCall struct {
+	Method  string `json:"method"`
+	Params  []any  `json:"params"`
+	ID      uint64 `json:"id"`
+	Version string `json:"jsonrpc"`
+}
+
+func ethID() uint64 {
+	buf := make([]byte, 8)
+	_, err := rand.Read(buf)
+	if err != nil {
+		panic(err)
+	}
+	return binary.LittleEndian.Uint64(buf)
+}
+
+func CallEthereum(ctx context.Context, c *http.Client, url, method string, params []any) ([]byte, error) {
+	ec := EthereumCall{
+		Method:  method,
+		Params:  params,
+		ID:      ethID(),
+		Version: EthereumVersion,
+	}
+	jec, err := json.Marshal(ec)
+	if err != nil {
+		return nil, err
+	}
+	req, err := http.NewRequestWithContext(ctx, method, url, bytes.NewBuffer(jec))
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Add("Content-Type", "application/json")
+	resp, err := c.Do(req)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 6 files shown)


service/hproxy/ethereum.go
```diff
@@ -0,0 +1,83 @@
...
+	buf := make([]byte, 8)
+	_, err := rand.Read(buf)
+	if err != nil {
+		panic(err)
+	}
+	return binary.LittleEndian.Uint64(buf)
+}
+
+func CallEthereum(ctx context.Context, c *http.Client, url, method string, params []any) ([]byte, error) {
+	ec := EthereumCall{
```
ClaytonNorthey92: let's use the [ethclient](https://pkg.go.dev/github.com/ethereum/go-ethereum/ethclient) package (we can use our fork).  this logic is already defined there

service/hproxy/hproxy.go
```diff
@@ -0,0 +1,960 @@
...
+	state  HVMState // Do NOT directly set, use utility functions!
+}
+
+func (s *Server) handleProxyRequest(w http.ResponseWriter, r *http.Request) {
+	log.Tracef("handleProxyRequest: %v", r.RemoteAddr)
+	defer log.Tracef("handleProxyRequest exit: %v", r.RemoteAddr)
+
+	// Select host to call
+	s.mtx.Lock()
+	id := -1
```
ClaytonNorthey92, line 467: nit: I'd prefer a separate `found` boolean

service/hproxy/hproxy.go
```diff
@@ -0,0 +1,960 @@
...
+	poker := s.hvmHandlers[id].poker
+	s.mtx.Unlock()
+	defer func() {
+		s.mtx.Lock()
+		s.hvmHandlers[id].poking = false
+		s.mtx.Unlock()
+	}()
+
+	err := poker.Poke(ctx)
+	if err != nil {
```
ClaytonNorthey92, line 806: let's add a trace log here to debug why nodes are being marked "unhealthy", when needed

service/hproxy/hproxy.go
```diff
@@ -0,0 +1,960 @@
...
+		}
+
+		j := make(map[string]any)
+		err = json.Unmarshal(resp, &j)
+		if err != nil {
+			return err
+		}
+
+		// Make sure we have "result"
+		if _, ok := j["result"]; !ok {
```
ClaytonNorthey92: we need to do more to determine "unhealthy", for example what if the node isn't getting unsafe blocks and not progressing?

service/hproxy/hproxy.go
```diff
@@ -0,0 +1,960 @@
...
+	}
+
+	// Find suitable candidate
+	if id == -1 {
+		var leastConnections uint = math.MaxUint
+		for k := range s.hvmHandlers {
+			if s.hvmHandlers[k].state == StateHealthy {
+				if s.hvmHandlers[k].connections < leastConnections {
+					leastConnections = s.hvmHandlers[k].connections
+					id = k
```
ClaytonNorthey92, line 486: nit: we should `break` here when we find a healthy node, so we don't continue iterating

ClaytonNorthey92 (commented):
hey @marcopeereboom I have added a few nitpicks plus a few things I would strongly ask that we add.

The latter are:
* a smarted "healthy" check; there are many scenarios where even with a `result` , the node isn't what we consider "healthy". we should address low-hanging fruit here
* we should add trace logging when we mark a node unhealthy, that way we can turn it on and debug why the node is unhealthy in production (otherwise it'll just be silent and we won't be able to tell what's going on)

additionally, I don't see any special handling for "writes", as was discussed in the requirements doc.  ex. `eth_sendTransaction` should go to all nodes, according to requirements

service/hproxy/hproxy.go
```diff
@@ -0,0 +1,960 @@
...
+
+type Node struct {
+	NodeURL string `json:"node_url"`
+}
+
+type NodeError struct {
+	NodeURL string `json:"node_url"`
+	Error   string `json:"error"`
+}
+
```
joshuasing, line 38: Structs ending with `Error` should usually implement the `error` type. I think this also works better here, because then we can log the error without also needing to log the node URL separately:
```suggestion
	Message   string `json:"error"`
}

func (ne NodeError) Error() string {
	return fmt.Sprintf("%s: %s", ne.NodeURL, ne.Message)
}
```

ClaytonNorthey92 (changes requested):
added some comments prior, forgot to select "request changes"

service/hproxy/hproxy.go
marcopeereboom: -1 is the sentinel, it can’t get out of sync. I’ll keep it as is.

service/hproxy/hproxy.go
marcopeereboom: We don’t know if the node we find first is the indeed the node with the least amount of connections. We must finish the loop, only shortcut is finding 0. Also, there won’t ever be more than like 5 VMs to look at.

ClaytonNorthey92: ah I see. makes sense.

also it's "the least _number_ of connections"

service/hproxy/hproxy.go
```diff
@@ -0,0 +1,960 @@
...
+			Enable:   true,
+			Idle:     7 * time.Second,
+			Interval: 7 * time.Second,
+			Count:    2,
+		},
+	}
+	ctx, cancel := context.WithTimeout(pctx, defaultDialTimeout)
+	defer cancel()
+	return d.DialContext(ctx, network, addr)
+}
```
joshuasing, line 97: `Timeout` can be set in Dialer, which `DialContext` uses to add a timeout to the provided context.
```suggestion
func dialer(ctx context.Context, network, addr string) (net.Conn, error) {
	log.Tracef("handleProxyDial: %v %v", network, addr)
	defer log.Tracef("handleProxyDial exit: %v %v", network, addr)

	defaultDialTimeout := 5 * time.Second
	d := &net.Dialer{
		Timeout:   defaultDialTimeout,
		KeepAliveConfig: net.KeepAliveConfig{
			Enable:   true,
			Idle:     7 * time.Second,
			Interval: 7 * time.Second,
			Count:    2,
		},
	}
	return d.DialContext(ctx, network, addr)
}
```

service/hproxy/hproxy.go
```diff
@@ -0,0 +1,960 @@
...
+		Network:             "mainnet",
+		PrometheusNamespace: appName,
+		RequestTimeout:      DefaultRequestTimeout,
+	}
+}
+
+type Server struct {
+	mtx  sync.RWMutex
+	wg   sync.WaitGroup
+	gctx context.Context // We need the global context in client requests
```
joshuasing: Are we able to use the request context (which is created with `http.Server.BaseContext`) instead of storing the context here?

service/hproxy/hproxy.go
```diff
@@ -0,0 +1,960 @@
...
+
+func (s *Server) nodeRemove(node string) error {
+	u, err := url.Parse(node)
+	if err != nil {
+		return fmt.Errorf("invalid url: %w", err)
+	}
+	switch u.Scheme {
+	case "http", "https":
+	default:
+		return fmt.Errorf("unsuported scheme: %v", u.Scheme)
```
joshuasing: nit: typo
```suggestion
		return fmt.Errorf("unsupported scheme: %v", u.Scheme)
```

service/hproxy/ethereum.go
```diff
@@ -0,0 +1,83 @@
...
+		Method:  method,
+		Params:  params,
+		ID:      ethID(),
+		Version: EthereumVersion,
+	}
+	jec, err := json.Marshal(ec)
+	if err != nil {
+		return nil, err
+	}
+	req, err := http.NewRequestWithContext(ctx, method, url, bytes.NewBuffer(jec))
```
joshuasing: This is using the Ethereum JSON-RPC method as the HTTP request method.
```suggestion
	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewBuffer(jec))
```

service/hproxy/hproxy.go
marcopeereboom: Fixed

service/hproxy/ethereum.go
marcopeereboom: I deliberately used the most basic and small curl-like mechanism. I don't think we need to grow the binary with giant piles of ethereum code unless we need additional functionality.

The code is equivalent to:
```
curl  http://localhost:8545  -X POST   -H "Content-Type: application/json"   --data '{"method":"eth_blockNumber","params":[],"id":1,"jsonrpc":"2.0"}' 
 
```

service/hproxy/hproxy.go
marcopeereboom: I agree with this but I wan't to test this on deployed testnet to get a better feel how it performs/works.  Tracked in #571

service/hproxy/hproxy.go
marcopeereboom: I tried that and it had side-effects and interacted with other timeouts. I read through the go code and in there they suggested to use a context as done here.

service/hproxy/ethereum.go
marcopeereboom: Fixed

service/hproxy/hproxy.go
marcopeereboom: Either we store the http server in the Server struct or the global context. I picked the global context but can be convinced to store the http server pointer instead. What do you prefer?

service/hproxy/hproxy.go
marcopeereboom: fixed in several spots.

service/hproxy/hproxy.go
marcopeereboom: While I agree, this isn't really a formal shared api. This is more a "here's a map and one of the entries is called error and if it's not empty bad things happened".  There is no contract, if we want to do that we should implement errors that are interpreted by the machine vs a human.

I am willing to revisit this in the future if we deem this necessary. For now I'd like to leave this alone. If you feel strongly about this open an issue with a formal api proposal.

It looks like this using curl:
```
[{"node_url":"http://34.90.253.98","error":""}]
```
and
```
[{"node_url":"http://34.90.253.98","error":"duplicate"}]
```

service/hproxy/hproxy.go
marcopeereboom: Moved httpServer into Server so that we can call it from the handler. Thanks.

ClaytonNorthey92 (approved):

Issue #566, opened 2025-07-08
audit: grep code for time.After use


marcopeereboom:
### Confirmation

- [x] I have checked for similar issues.

### Describe the bug

We must remove all time.After calls from the code base. They do not work across long hardware sleeps. Sweep code and replace them with time.Ticker.

### Expected behaviour

good luck

### Environment

Version: 
Operating System: 
Architecture:

PR #572, opened 2025-07-14
all: replace time.After with time.Tick

AL-CT:
**Summary**
This PR aims to fix issue #566 by replacing the use of `time.After` with `time.Tick`, as the latter will perform better at recovering from system hibernation or suspensions, due to the periodic nature of Ticker vs the punctual nature of Timer.

**Changes**
Replace all instances of `time.After` with `time.Tick`.

**Notes**
As per the `time.Tick` documentation:
```
Before Go 1.23, this documentation warned that the underlying
[Ticker] would never be recovered by the garbage collector, and that
if efficiency was a concern, code should use NewTicker instead and
call [Ticker.Stop] when the ticker is no longer needed.
As of Go 1.23, the garbage collector can recover unreferenced
tickers, even if they haven't been stopped.
The Stop method is no longer necessary to help the garbage collector.
There is no longer any reason to prefer NewTicker when Tick will do.
```

Since our use of Ticker is meant for recovery in case of long hardware sleeps yet should remain a punctual (not recurring) check for time elapsed, there is no need to use `time.NewTicker` over `time.Tick`

### DIFF
--- a/database/postgres/postgres.go
+++ b/database/postgres/postgres.go
@@ -178,7 +178,6 @@ func (p *Database) ntfnListenHandler(ctx context.Context) {
 			// processed at the same time.
 			go p.ntfnEventHandler(ctx, pqn)
 
-		/// XXX ticker check
 		case <-time.Tick(60 * time.Second):
 			go func() {
 				// log.Tracef("ntfnHandler: ping")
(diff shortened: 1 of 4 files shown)


marcopeereboom (approved):
Fuck yeah!

joshuasing (changes requested):
Blocking until we have throughly reviewed each `time.After` use and its replacement

joshuasing (approved):

PR #588, opened 2025-07-16
all: move utility packages under a 'util' package

joshuasing:
**Summary**
Move utility packages under a `util` package, and update import paths.
This is to keep the root directory from getting exponentially messy as we keep adding util packages.

As this PR is breaking the API, it should be included only in v2.

**Changes**
- Create `util` package
- Move `testutil`, `ttl` and `version` packages under the `util` package.
- Fix import paths.


marcopeereboom (commented):
I actively dislike a generic package where tiny one line functions land. They always become a dumping ground. Either something is a package or isn’t. The second part is needlessly moving stuff into directories. For example the version package is a full citizen of the repo, why would it live in an obscure directory?

I am ok moving like rawdb into database and ttl maybe into tbc but I honestly don’t see what this gains.

joshuasing:
Closing for now

PR #593, opened 2025-07-16
readme: rewrite and bring up-to-date

joshuasing:
**Summary**
Rewrite the `README.md` file with up-to-date information.

**Changes**

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,6 @@ This repository also contains several utility binaries.
 | Name                       | Description                   |
 |----------------------------|-------------------------------|
 | [`btctool`](./cmd/btctool) | Bitcoin utility tool          |
-| [`extool`](./cmd/extool)   | ElectrumX/Electrs client tool |
 | [`hemictl`](./cmd/hemictl) | Hemi network controller CLI   |
 | [`keygen`](./cmd/keygen)   | Keypair generation tool       |
 
--- a/README.md
+++ b/README.md
@@ -1,77 +1,57 @@
-# ⚡️ The Hemi Network
+# Hemi Network
+
+[![Go Reference](https://pkg.go.dev/badge/github.com/hemilabs/heminetwork.svg)](https://pkg.go.dev/github.com/hemilabs/heminetwork)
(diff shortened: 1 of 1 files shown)


marcopeereboom (approved):
I like this a lot but we should remove electrs mentions and simply removing it since we killed support for it altogether.

README.md
```diff
@@ -1,77 +1,67 @@
...
+|----------------------------|-------------------------------|
+| [`btctool`](./cmd/btctool) | Bitcoin utility tool          |
+| [`hemictl`](./cmd/hemictl) | Hemi network controller CLI   |
+| [`keygen`](./cmd/keygen)   | Keypair generation tool       |
+
+## Installation
+
+### Binaries
+
+Pre-built binaries for Linux, macOS, Windows and OpenBSD are available
```
jcvernaleo: I will always object to something that is missing an Oxford comma.

README.md
```diff
@@ -82,135 +72,42 @@ Pre-built binaries are available on the [Releases Page](https://github.com/hemil
...
+Discord server](https://discord.gg/hemixyz).
 
-When running BFG, you'll want the following env variables set:
+#### Security Vulnerabilities
 
-* `BFG_BITCOIN_URL`: the `TBC`'s websocket url that you will connect to
+If you discover vulnerabilities in Hemi, we encourage responsible disclosure of the vulnerability so that we can take
+steps to resolve the vulnerability as quickly as possible. We ask you to help us better protect Hemi and our users by
+reporting vulnerabilities through HackerOne. **Never report security vulnerabilities publicly, especially on GitHub
+issues.**
```
jcvernaleo, line 103: Should we also mention the secops team email here?  I'm not totally sure.

joshuasing: I think we chose to exclude this except for:
> If you have any questions or concerns about this policy or this project's security practices, please do not hesitate to contact us at [security@hemi.xyz](mailto:security@hemi.xyz).

in the security policy document to prevent people from doing that instead of the VDP report, although I don't mind either way. Up to you.

README.md
joshuasing: Whoops, good catch. Fixed!

README.md
jcvernaleo: Let's not add it then.  I wasn't totally sure but I think you are right.

jcvernaleo (approved):
OK

PR #596, opened 2025-07-18
popm: add static fee override

AL-CT:
**Summary**
Currently, fee calculation for pop miners is done by the gozer. However, the only functional gozer is TBC, which requires the mempool to be enabled in order for fee estimation to work. This can be both expensive and potentially unreliable. As such, an option to override fee estimation with a static fee was added.

**Changes**
Introduce a way to set a static fee for the pop miner.
Changed the wallet transaction creation methods to allow float sats/byte fees.
Changed config.go to be able to parse float values in env vars.

### DIFF
--- a/service/popm/popm.go
+++ b/service/popm/popm.go
@@ -16,12 +16,14 @@ import (
 	"github.com/btcsuite/btcd/btcutil/hdkeychain"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	btcmempool "github.com/btcsuite/btcd/mempool"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/juju/loggo"
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/hemilabs/heminetwork/api/gethapi"
+	"github.com/hemilabs/heminetwork/api/tbcapi"
 	"github.com/hemilabs/heminetwork/bitcoin/wallet"
 	"github.com/hemilabs/heminetwork/bitcoin/wallet/gozer"
 	"github.com/hemilabs/heminetwork/bitcoin/wallet/gozer/blockstream"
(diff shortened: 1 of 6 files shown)


service/popm/popm.go
```diff
@@ -332,6 +338,30 @@ func (s *Server) latestKeystones(ctx context.Context, count int) (*gethapi.L2Key
...
+		feeEstimates, err := s.gozer.FeeEstimates(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("fee estimates: %w", err)
+		}
+		feeAmount, err = gozer.FeeByConfirmations(s.cfg.BitcoinConfirmations, feeEstimates)
+		if err != nil {
+			return nil, fmt.Errorf("fee by confirmations: %w", err)
+		}
+	}
+	return feeAmount, nil
```
joshuasing: I think this can be simplified as:
```suggestion
	if s.cfg.StaticFee != 0 {
		return  &tbcapi.FeeEstimate{
			Blocks:      s.cfg.BitcoinConfirmations,
			SatsPerByte: float64(s.cfg.StaticFee),
		}
	}

	// Estimate BTC fees.
	feeEstimates, err := s.gozer.FeeEstimates(ctx)
	if err != nil {
		return nil, fmt.Errorf("fee estimates: %w", err)
	}
	feeAmount, err := gozer.FeeByConfirmations(s.cfg.BitcoinConfirmations, feeEstimates)
	if err != nil {
		return nil, fmt.Errorf("fee by confirmations: %w", err)
	}
	return feeAmount, nil
```

joshuasing (changes requested):

service/popm/popm.go
AL-CT: Oops, you're right! What I get for copy pasting stuff around

service/popm/popm.go
```diff
@@ -51,6 +53,9 @@ const (
 	defaultL2KeystoneMaxAge       = 4 * time.Hour
 	defaultL2KeystonePollTimeout  = 13 * time.Second
 	defaultL2KeystoneRetryTimeout = 15 * time.Second
+
+	// convert btcd minimum fee from sat/kB to sat/byte
+	minRelayFee = uint(btcmempool.DefaultMinRelayTxFee) / 1000
```
marcopeereboom: This should be moved to tbc mempool.

marcopeereboom (changes requested):

service/popm/popm.go
AL-CT: Made it into a local constant 👍

marcopeereboom (approved):

Issue #524, opened 2025-06-02
Protocol: audit error types


AL-CT:
The current protocol api error types don't seem to represent all possible states. Namely:

- Internal errors are errors that occurred within the application and are not caused by a client (e.g. validation or input errors). The actual error message should not be sent to clients, as it is internal to the application, and may be server-operator specific.
- Request errors are usually something caused by a client, e.g. validation or input errors, and therefore should not be logged server-side and do not contain an embedded trace.

However, neither may necessarily be the best fit for specific errors (e.g., database.NoFoundErrors), in which a call is valid, but the server could not attend to the request, yet _should_ communicate why it could not to the caller. 

We should consider the current error types and discuss whether these should be altered in some way.

PR #608, opened 2025-07-22
protocol: add NotFoundError

AL-CT:
**Summary**
Addresses issue #524 

**Changes**
Simply adds a `NotFoundError` function meant to uniformize errors representing valid requests that the server cannot fulfill due to missing or unknown resources.


marcopeereboom (approved):
We need to rip this bandaid. Let’s do it.

ClaytonNorthey92 (approved):

PR #611, opened 2025-07-23
all: fix several fragile tests

marcopeereboom:
**Summary**
We have a bunch of fragile test that are impeding development. This fixes a few by doing things in order (write first, then tell waiter) and block on waiter. This should solve several problems but others remain. Remember kids, don't use channels.

**Changes**
Fix order of operations and wait when needed.
Fix several fragile bits in e2e.


testutil/mock/geth.go
```diff
@@ -116,23 +116,26 @@ func (f *OpGethMockHandler) mockOpGethHandleFunc(w http.ResponseWriter, r *http.
...
-			return f.pctx.Err()
-		case f.msgCh <- msg.Method:
-		default:
-			// discard message if channel is blocked
-		}
+		// Tell caller we got it but delay a bit. This is super hokey
+		// and should be rewritten but the net is that we need to do
+		// the write prior to the msg being sent.
+		go func() {
+			time.Sleep(250 * time.Millisecond)
```
AL-CT: Doesn't this introduce a different race? Can we move this to the end of the function and not have a sleep?

testutil/mock/geth.go
```diff
@@ -212,7 +215,7 @@ func (f *OpGethMockHandler) mockOpGethHandleFunc(w http.ResponseWriter, r *http.
 
 			abrevHash, err := chainhash.NewHashFromStr(shash)
 			if err != nil {
-				return err
+				panic(err)
```
AL-CT: I would rather have these return an error, as they are sent to the caller and handled there

testutil/mock/mock.go
```diff
@@ -88,22 +89,24 @@ func (f *mockHandler) Start() {
 	f.mtx.Lock()
 	defer f.mtx.Unlock()
 	f.isRunning = true
-	log.Tracef("%v: server started", f.name)
+	log.Infof("%v: server started", f.name)
```
AL-CT, line 114: Was this a test change? If so, can we change it back to TRACE? 🤔

testutil/mock/tbc.go
```diff
@@ -107,32 +107,36 @@ func (f *TBCMockHandler) mockTBCHandleFunc(w http.ResponseWriter, r *http.Reques
...
-			return f.pctx.Err()
-		case f.msgCh <- string(cmd):
-		default:
-			// discard message if channel is blocked
-		}
+		// Tell caller we got it but delay a bit. This is super hokey
+		// and should be rewritten but the net is that we need to do
+		// the write prior to the msg being sent.
+		go func() {
+			time.Sleep(250 * time.Millisecond)
```
AL-CT: Same as above

service/bfg/bfg.go
```diff
@@ -336,8 +336,10 @@ func (s *Server) shortCircuitFinality(ctx context.Context, kss *hemi.L2Keystone,
 }
 
 func (s *Server) handleKeystoneFinality(w http.ResponseWriter, r *http.Request) {
+	log.Infof("handleKeystoneFinality: %v", r.RemoteAddr)
```
AL-CT: Was this a test change? Can we revert it?

service/bfg/bfg.go
```diff
@@ -352,7 +354,9 @@ func (s *Server) handleKeystoneFinality(w http.ResponseWriter, r *http.Request)
 	// firstLoop := true
 	for {
 		// Call opgeth to retrieve keystones
+		log.Infof("handleKeystoneFinality: calling opgeth")
 		resp, err := s.opgethL2KeystoneValidity(r.Context(), *hash, defaultKeystoneCount)
+		log.Infof("handleKeystoneFinality: opgeth %v", err)
```
AL-CT: Was this a test change? Can we revert it?

testutil/mock/geth.go
marcopeereboom: This is shitty but the smallest change I could make to get the point cross. The issue is that we wrote (sometimes not because of default) to the other side PRIOR to the websocket write. Thus the other end moved forward assuming the write had happened. This is hokey for sure.

service/bfg/bfg.go
marcopeereboom: fixed

service/bfg/bfg.go
marcopeereboom: fixed

testutil/mock/geth.go
AL-CT: Gotcha. While I think that due to how the tests were, the order shouldn't matter, I agree this is correct and the order should be reversed. The default I am iffy on, because we might not want to be forced to listen for the channels in tests, but it's also not a hill I will die on

testutil/mock/mock.go
marcopeereboom: I prefer to leave it as is.

testutil/mock/tbc.go
marcopeereboom: Same comment as above.

testutil/mock/geth.go
marcopeereboom: Doesn't hit. So we should leave it a panic and change it later if we need it to be an error.

Makefile
```diff
@@ -76,7 +76,7 @@ race:
 	go test -v -race ./...
 
 test:
-	go test -test.timeout=20m ./...
+	go test -v -test.timeout=20m ./...
```
AL-CT, line 79: Was this a test change or final? Probably shouldn't change it here

marcopeereboom: other make test has it, that's why i added it there

marcopeereboom: It was decided to keep it.

joshuasing (approved):

PR #612, opened 2025-07-23
popm: add exponential backoff reconnection logic to opgeth

### LINKED ISSUE #569: popm: add exponential backoff reconnection logic
### Confirmation

### Problem

As discussed in the meeting regarding #396 we should introduce logic for pop miners that disconnect from opgeth to attempt to reconnect at increasingly longer intervals, up to a limit (e.g., 15 seconds), with jitter.

### Suggested solution

Write code :)

### Additional Information

_No response_

tosynthegeek:
## Changes
- Modified `Server.opgeth()` method to implement exponential backoff instead of fixed timeout
- Uses existing `opgethReconnectTimeout` config as base delay
- Exponential growth: `baseDelay * 2^(attempt-1)` for first 3 attempts
- Maximum delay capped at 15 seconds (as specified in issue)
- Adds jitter (0-10% of delay) 
- Resets attempt counter on successful reconnection
- Maintains infinite retry behavior as well

Fixes #569

### DIFF
--- a/service/popm/popm.go
+++ b/service/popm/popm.go
@@ -590,8 +590,8 @@ func (s *Server) opgeth(ctx context.Context) {
 			delay = maxDelay
 		}
 
-		jitter := delay / 10
-		delay += time.Duration(mathrand.Int64N(int64(jitter)*2) - int64(jitter))
+		jitter := int64(delay / 10)
+		delay += time.Duration(mathrand.Int64N(jitter*2) - jitter)
 
 		log.Debugf("reconnecting to: %v in %v", s.cfg.OpgethURL, delay)
 


service/popm/popm.go
```diff
@@ -563,22 +563,48 @@ func (s *Server) opgeth(ctx context.Context) {
 	log.Tracef("opgeth")
 	defer log.Tracef("opgeth exit")
 
+	const maxExponentialAttempts = 3
```
AL-CT: Since we're already capping this with `maxDelay`, can we remove this const?

service/popm/popm.go
```diff
@@ -563,22 +563,48 @@ func (s *Server) opgeth(ctx context.Context) {
...
+
+		jitter := time.Duration(time.Now().UnixNano() % int64(delay/10))
+		delay += jitter	
+
+		log.Debugf("reconnecting to: %v in %v", s.cfg.OpgethURL, delay)
+
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(delay):
```
AL-CT: nit: we changed the codebase to use `time.Tick` over `time.After`, so would prefer we keep that here as well

AL-CT:
```suggestion
		case <-time.Tick(delay):
```

AL-CT (changes requested):
Thank you for the commit! Looks good, just left two minor nits. Appreciate the hard work 👍

service/popm/popm.go
tosynthegeek: > Since we're already capping this with `maxDelay`, can we remove this const?

Resolved, please check @AL-CT

service/popm/popm.go
tosynthegeek: > nit: we changed the codebase to use `time.Tick` over `time.After`, so would prefer we keep that here as well

Fixed as well

service/popm/popm.go
AL-CT: Thanks!

AL-CT (approved):

joshuasing (approved):
Thank you!

I have pushed commits to fix a few small issues:
 - Linting issue (whitespace at the end of a line)
 - Use a random number when creating jitter (and make it +/- jitter instead of only +jitter).
 - Make `attempt` zero-based

PR #619, opened 2025-07-28
bfg, popm: add health checks

### LINKED ISSUE #148: all: add health check listeners to each daemon
### Confirmation

### Problem

There is currently no easy way to monitor the health of the daemons when running in a containerised environment where health (readiness and liveness) matters.

### Suggested solution

It would be nice to add a health HTTP listener to the daemons that exports a small amount of health information and performs async health checks to check the overall status of the daemon.

If the daemon is unhealthy (i.e. not able to operate correctly), it should report such so that the container can be marked as unhealthy and be restarted.

### Additional Information

_No response_

marcopeereboom:
**Summary**
Fixes #148 

**Changes**
Add health checks to popm and bfg.

### DIFF
--- a/bitcoin/wallet/gozer/blockstream/blockstream.go
+++ b/bitcoin/wallet/gozer/blockstream/blockstream.go
@@ -71,8 +71,8 @@ func (bs *blockstreamGozer) BestHeightHashTime(ctx context.Context) (uint64, *ch
 		return 0, nil, timestamp, err
 	}
 	if t, ok := bi["timestamp"]; ok {
-		if ts, ok := t.(int64); ok && ts > 0 {
-			timestamp = time.Unix(ts, 0)
+		if ts, ok := t.(float64); ok && ts > 0 {
+			timestamp = time.Unix(int64(ts), 0)
 		} else {
 			return 0, nil, timestamp, fmt.Errorf("invalid timestamp")
 		}
--- a/service/bfg/bfg.go
+++ b/service/bfg/bfg.go
@@ -318,7 +318,7 @@ func (s *Server) opgethL2KeystoneValidity(ctx context.Context, hash chainhash.Ha
 	return resp, nil
 }
 
-func (s *Server) gethBestHeightHash(ctx context.Context) (uint64, *chainhash.Hash, time.Time, error) {
+func (s *Server) gethBestHeightHash(pctx context.Context) (uint64, *chainhash.Hash, time.Time, error) {
 	log.Tracef("gethBestHeightHash")
 	defer log.Tracef("gethBestHeightHash exit")
 
@@ -329,6 +329,9 @@ func (s *Server) gethBestHeightHash(ctx context.Context) (uint64, *chainhash.Has
 	if err != nil {
 		return 0, nil, t, err
 	}
+
+	ctx, cancel := context.WithTimeout(pctx, gethapi.DefaultCommandTimeout)
+	defer cancel()
 	header, err := geth.HeaderByNumber(ctx, height)
 	if err != nil {
 		return 0, nil, t, fmt.Errorf("error calling opgeth: %w", err)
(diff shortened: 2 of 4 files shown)


bitcoin/wallet/gozer/blockstream/blockstream.go
```diff
@@ -46,20 +47,43 @@ func (bs *blockstreamGozer) Connected() bool {
...
-		return 0, err
+		return 0, nil, timestamp, err
+	}
+	if t, ok := bi["timestamp"]; ok {
+		if ts, ok := t.(int64); ok && ts > 0 {
+			timestamp = time.Unix(ts, 0)
+		} else {
+			return 0, nil, timestamp, fmt.Errorf("invalid timestamp")
+		}
+	}
```
AL-CT: Shouldn't we fail if the block info doesn't have `timestamp` information? Or at least not just continue.

bitcoin/wallet/gozer/blockstream/blockstream.go
```diff
@@ -46,20 +47,43 @@ func (bs *blockstreamGozer) Connected() bool {
...
-	var height uint64
-	err = json.Unmarshal(rawHeight, &height)
+	var bi map[string]any
+	err = json.Unmarshal(blockInfo, &bi)
 	if err != nil {
-		return 0, err
+		return 0, nil, timestamp, err
+	}
+	if t, ok := bi["timestamp"]; ok {
+		if ts, ok := t.(int64); ok && ts > 0 {
```
AL-CT: As per the[ json.Unmarshal docs](https://pkg.go.dev/encoding/json#Unmarshal), I believe JSON numbers are cast to float64, so casting to int64 probably doesn't work here.

service/bfg/bfg.go
```diff
@@ -251,6 +268,28 @@ func (s *Server) opgethL2KeystoneValidity(ctx context.Context, hash chainhash.Ha
...
 }
 
+func (s *Server) gethBestHeightHash(ctx context.Context) (uint64, *chainhash.Hash, time.Time, error) {
+	log.Tracef("gethBestHeightHash")
+	defer log.Tracef("gethBestHeightHash exit")
+
+	var t time.Time
+
+	height := big.NewInt(int64(rpc.LatestBlockNumber))
+	header, err := s.opgethClient.HeaderByNumber(ctx, height)
```
AL-CT: I think it might be a good idea to create a child context with a timeout in case this call takes a long time

AL-CT: Probably should also do this on the `callOpgeth` method, or just move this there

service/popm/popm.go
```diff
@@ -658,17 +687,140 @@ func (s *Server) mine(ctx context.Context) error {
...
+}
+
+func (s *Server) gethBestHeightHash(ctx context.Context) (uint64, *chainhash.Hash, time.Time, error) {
+	log.Tracef("gethBestHeightHash")
+	defer log.Tracef("gethBestHeightHash exit")
+
+	var t time.Time
+
+	height := big.NewInt(int64(rpc.LatestBlockNumber))
+	header, err := s.opgethClient.HeaderByNumber(ctx, height)
```
AL-CT, line 736: Same as above regarding the timeout context

bitcoin/wallet/gozer/blockstream/blockstream.go
marcopeereboom: fixed

bitcoin/wallet/gozer/blockstream/blockstream.go
marcopeereboom: works

service/bfg/bfg.go
marcopeereboom: fixed

service/popm/popm.go
marcopeereboom: Fixed

AL-CT (approved):

PR #620, opened 2025-07-28
localnode: dynamically generate private keys and addresses

### LINKED ISSUE #618: localnet: remove hardcoded addresses for pop miner
### Confirmation

### Problem

PoP miner addresses for both BTC and ETH are derived from the same private key. These addresses are used in localnet e2e testing. However, they are hardcoded in a few spots that could otherwise be dynamically derived from the private key.

### Suggested solution

Change the spots in the code that use hardcoded values for the addresses. Namely [here](https://github.com/hemilabs/heminetwork/blob/e2e1395825880a1d857149ad9209e02d160cf3ad/e2e/.env#L4), [here](https://github.com/hemilabs/heminetwork/blob/e2e1395825880a1d857149ad9209e02d160cf3ad/e2e/monitor/main_test.go#L51), and [here](https://github.com/hemilabs/heminetwork/blob/e2e1395825880a1d857149ad9209e02d160cf3ad/e2e/monitor/main.go#L320C1-L320C88).

### Additional Information

_No response_

AL-CT:
Fixes #618
Localnet tests passing [here](https://github.com/hemilabs/heminetwork/actions/runs/16593817373/job/46935605434)

**Changes:**
Removes hardcoded private key and btc / eth addresses in `e2e` and replaces them with dynamically generated values, passed through env variables.


e2e/monitor/main_test.go
```diff
@@ -46,12 +47,14 @@ import (
 )
 
 const (
-	localnetPrivateKey = "dfe61681b31b12b04f239bc0692965c61ffc79244ed9736ffa1a72d00a23a530"
 	retries            = 10
-	btcAddress         = "mw47rj9rG25J67G6W8bbjRayRQjWN5ZSEG"
+	localnetPrivateKey = "dfe61681b31b12b04f239bc0692965c61ffc79244ed9736ffa1a72d00a23a530"
```
ClaytonNorthey92, line 53: nit: we may want to leave a comment as to why this is still hard-coded; if I am not mistaken, it's the eth dev pre-funded private key and shouldn't change with our changes.  it's used to create txs

ClaytonNorthey92 (approved):
nice 💪

e2e/monitor/main_test.go
joshuasing: Agreed - adding a comment here would be best to prevent any future confusion.

AL-CT: Fixed 👍

marcopeereboom (changes requested):
Why aren't you using keygen instead of making a new gen.go?

marcopeereboom (approved):

PR #635, opened 2025-08-01
mock: limit number of new keystone notifications

### LINKED ISSUE #632: popm: fragile test: TestTickingPopMiner
### Confirmation

### Describe the bug

example of fragile test failing here https://github.com/hemilabs/heminetwork/actions/runs/16657958021/job/47147698978?pr=631

### Expected behaviour

tests should not be fragile

### Environment

Version: 
Operating System: 
Architecture: 

### Additional Information

_No response_

AL-CT:
**Summary**
Once a pop miner sends a subscription request to `mock opgeth`, it will periodically send new keystone notifications until the connection ends. This can cause some issues with tests, as popm will constantly be running the same logic, which requires a lock, and may prevent other parts of the code from executing.

Should fix #632

**Changes**
Be explicit about how many keystones are "available" from the list used by `mock opgeth`, and only send notifications until all keystones become available.

Also added some stronger logging for `TestTickingPopMiner` as that test seems particularly fragile.

### DIFF
--- a/api/auth/secp256k1.go
+++ b/api/auth/secp256k1.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2024 Hemi Labs, Inc.
+// Copyright (c) 2024-2025 Hemi Labs, Inc.
 // Use of this source code is governed by the MIT License,
 // which can be found in the LICENSE file.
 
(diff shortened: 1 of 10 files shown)


marcopeereboom (changes requested):
I want to have a go at this before approving. This feels artificial.

PR #660, opened 2025-08-19
tbc: add "not allowed in external-header mode" error type

### LINKED ISSUE #551: tbc: add "not allowed in external-header mode" error type
### Confirmation

### Problem

There are multiple instances of methods in tbc that are not allowed when in external-header mode. We should create an error type for this. [This](https://github.com/hemilabs/heminetwork/blob/af45269e28a743a6004009858ae0e99593cdab86/service/tbc/tbc.go#L1888) and [this](https://github.com/hemilabs/heminetwork/blob/af45269e28a743a6004009858ae0e99593cdab86/service/tbc/tbc.go#L1913) are good examples.

### Suggested solution

Create a new error type.

### Additional Information

_No response_

Dzmitryy1812:
Adds a typed error for operations not allowed while TBC runs in External Header Mode, enabling precise errors.Is checks and clearer messages. Fixes #551.

Add ExternalHeaderNotAllowedError in service/tbc/tbc.go with Error() and Is() to support errors.Is(err, &ExternalHeaderNotAllowedError{}).
Return ExternalHeaderNotAllowedError when External Header Mode is enabled for:
- BlockByHash
- KeystonesByHeight
- BlocksMissing
- BalanceByAddress
- BalanceByScriptHash
- UtxosByAddress
- ScriptHashAvailableToSpend
- SpentOutputsByTxId
- BlockInTxIndex
- BlockHashByTxId
- TxById
- TxBroadcastAllToPeer
- TxBroadcast
- FeesByBlockHash
- FullBlockAvailable
- Run
Update tests in service/tbc/tbc_test.go to assert the new error type.

### DIFF
--- a/service/tbc/tbc.go
+++ b/service/tbc/tbc.go
@@ -1823,7 +1823,7 @@ func (s *Server) BlockByHash(ctx context.Context, hash chainhash.Hash) (*btcutil
 	defer log.Tracef("BlockByHash exit")
 
 	if s.cfg.ExternalHeaderMode {
-		return nil, &ExternalHeaderNotAllowedError{Op: "BlockByHash"}
+		return nil, NewExternalHeaderNotAllowedError("BlockByHash")
 	}
 
 	return s.db.BlockByHash(ctx, hash)
@@ -1836,7 +1836,7 @@ func (s *Server) KeystonesByHeight(ctx context.Context, height uint32, depth int
 	defer log.Tracef("KeystonesByHeight exit")
 
 	if s.cfg.ExternalHeaderMode {
-		return nil, &ExternalHeaderNotAllowedError{Op: "KeystonesByHeight"}
+		return nil, NewExternalHeaderNotAllowedError("KeystonesByHeight")
 	}
 
 	return s.db.KeystonesByHeight(ctx, height, depth)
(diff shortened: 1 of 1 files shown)


service/tbc/tbc.go
```diff
@@ -103,6 +103,26 @@ var (
 	fixupStrategy = 3 // Do not touch unless your name is marco
 )
 
+// ExternalHeaderNotAllowedError is returned when an operation is invoked
+// that is not allowed while TBC is running in External Header Mode.
+type ExternalHeaderNotAllowedError struct {
+	Operation string
```
joshuasing: I think `Operation` could be shortened to `Op` (this is common in some errors from Go stdlib).
```suggestion
	Op string
```

service/tbc/tbc_test.go
joshuasing, line 1: Could the underscores `_` be removed from the test function names?

joshuasing (changes requested):
Thanks for taking the time to contribute!

joshuasing:
My change suggestion only changes the struct field name. Please make sure to update usage.

service/tbc/tbc_test.go
Dzmitryy1812: Thanks, I fixed it.

Dzmitryy1812:
corrected it

service/tbc/tbc.go
```diff
@@ -1795,7 +1823,7 @@ func (s *Server) BlockByHash(ctx context.Context, hash chainhash.Hash) (*btcutil
 	defer log.Tracef("BlockByHash exit")
 
 	if s.cfg.ExternalHeaderMode {
-		return nil, errors.New("cannot call BlockByHash on TBC running in External Header mode")
+		return nil, &ExternalHeaderNotAllowedError{Op: "BlockByHash"}
```
joshuasing: Now that we have a creator function, `NewExternalHeaderNotAllowedError` - we can use it instead of creating the error directly (repeat for each `&ExternalHeaderNotAllowedError{Op: "Name"}` use)
```suggestion
		return nil, NewExternalHeaderNotAllowedError("BlockByHash")
```

PR #662, opened 2025-08-20
tbc: add indexer abstraction layer

marcopeereboom:
This abstraction is needed to find the subtle differences between the various indexers. The surrounding code is 99.9% the same and the copy paste to add a new indexer is very error prone. Finally bite the bullet and move on making this code use the same code path.

After this pile of shit has settled we can make a proper indexer interface.


service/tbc/indexer.go
```diff
@@ -196,25 +228,24 @@ func wind(ctx context.Context, i Indexer, startBH, endBH *tbcd.BlockHeader) erro
 
 	// Allocate here so that we don't waste space when not indexing.
 	// XXX the cache *really* should become a list methinks.
-	es := i.cache().Generic()
-	defer i.cache().Clear()
+	defer i.cacheFlush()
```
marcopeereboom: I don't think this is right. The cache MUST be allocated on the stack to help gc. It is also has a very large capacity so when not indexing we need to recoup that memory.

service/tbc/indexer.go
```diff
@@ -84,18 +85,69 @@ type Indexer interface {
...
-	cacheFlush()
+	geometry geometryParams
+	genesis  *HashHeight
 
-	// XXX i would be tickled if we can get rid of these inside the interface
-	geometry() geometryParams                          // Return geometry parameters
-	commit(context.Context, int, chainhash.Hash) error // Commit index cache to disk
-	genesis() *HashHeight                              // Genesis override, like hemi
+	p     indexer
+	cache indexerCache
```
marcopeereboom: Again, cache needs to live on the stack.

service/tbc/indexer.go
```diff
@@ -449,7 +449,7 @@ func (c *indexerCommon) parseBlocks(ctx context.Context, endHash *chainhash.Hash
 		}
 
 		// Move to next block
-		hh, err = nextCanonicalBlockheader(ctx, c.geometry, endHash, hh)
```
marcopeereboom: haha you are an odd one. Previous commit you made the code wide by adding `Index` to an interface that is always just Index and now you remove the hint by shortening the name of what we are passing in.

service/tbc/keystoneindex.go
```diff
@@ -31,26 +30,23 @@ var (
...
 		enabled: enabled,
-		geometry: geometryParams{
-			db:    db,
-			chain: chain,
-		},
-		p:     ki,
-		cache: ki.cache,
+		g:       g,
+		p:       ki,
+		cache:   ki.cache,
```
marcopeereboom: hmmm, does this mean the cache goes out of scope when the indexer exits?

service/tbc/tbc.go
```diff
@@ -2725,11 +2725,11 @@ func (s *Server) dbOpen(ctx context.Context) error {
...
 		s.fixupCache = s.fixupCacheChannel
 	}
-	s.ui = NewUtxoIndexer(s.g.chain, s.cfg.MaxCachedTxs, s.g.db, s.fixupCache)
-	s.ti = NewTxIndexer(s.g.chain, s.cfg.MaxCachedTxs, s.g.db)
-	if s.cfg.HemiIndex {
-		s.ki = NewKeystoneIndexer(s.g.chain, s.cfg.MaxCachedKeystones,
-			s.g.db, s.cfg.HemiIndex, s.hemiGenesis)
+	s.ui = NewUtxoIndexer(s.g, s.cfg.MaxCachedTxs, s.fixupCache)
+	s.ti = NewTxIndexer(s.g, s.cfg.MaxCachedTxs)
+	if s.cfg.HemiIndex { // TODO: nil if not enabled, but we pass enabled into the indexer?
```
marcopeereboom: I don't understand this comment.

joshuasing: `HemiIndex` is always true inside this `if` statement, however we are still passing it into the Keystone indexer and the Indexer has an `Enabled() bool` function.

marcopeereboom: hmmm ok, do as you please here. Just make sure to test in tbcd.

service/tbc/indexer.go
```diff
@@ -0,0 +1,563 @@
...
+func (c *Cache[K, V]) Map() map[K]V {
+	return c.m
+}
+
+type Indexer interface {
+	// Enabled returns whether the indexer is enabled.
+	Enabled() bool
+
+	// Indexing returns whether the indexer is currently indexing.
+	Indexing() bool
```
joshuasing, line 67:
```suggestion
	// IsEnabled returns whether the indexer is enabled.
	IsEnabled() bool

	// IsIndexing returns whether the indexer is currently indexing.
	IsIndexing() bool
```

service/tbc/indexer.go
```diff
@@ -0,0 +1,563 @@
...
+		cached, _, pct := cache.Stats()
+		log.Infof("%v indexer blocks processed %v in %v cached %v (%v%%) avg/blk %v",
+			c, blocksProcessed, time.Since(start), cached, pct, cached/blocksProcessed)
+
+		// Flush to disk
+		start = time.Now()
+		if err := c.p.commit(ctx, 1, last.Hash, cache); err != nil {
+			return fmt.Errorf("block %v update: %w", c, err)
+		}
+		// TODO: do we need to clear the cache here?
```
joshuasing: @marcopeereboom Please check this

service/tbc/indexer.go
```diff
@@ -0,0 +1,563 @@
...
+		cached, _, pct := cache.Stats()
+		log.Infof("%v unwinder blocks processed %v in %v cached %v cache (%v%%) avg/blk %v",
+			c, blocksProcessed, time.Since(start), cached, pct, cached/blocksProcessed)
+
+		// Flush to disk
+		start = time.Now()
+		if err = c.p.commit(ctx, -1, last.Hash, cache); err != nil {
+			return fmt.Errorf("block %v update: %w", c, err)
+		}
+		// TODO: do we need to clear the cache here?
```
joshuasing: @marcopeereboom Please check this

service/tbc/tbc.go
```diff
@@ -135,6 +135,36 @@ func init() {
 	}
 }
 
+type HashHeight struct {
+	Hash      chainhash.Hash `json:"hash"`
+	Height    uint64         `json:"height"`
+	Timestamp int64          `json:"timestamp"` // optional
```
joshuasing:
```suggestion
	Timestamp int64          `json:"timestamp,omitempty"` // optional
```

service/tbc/tbc.go
joshuasing: Need to fix this.

service/tbc/tbc.go
```diff
@@ -2734,7 +2920,10 @@ func (s *Server) Run(pctx context.Context) error {
 		return fmt.Errorf("df: %w", err)
 	}
 	if df != 0 {
-		blockPerDay := uint64(24 * time.Hour / s.chainParams.TargetTimePerBlock)
+		if s.g.chain == nil {
+			panic("gfy")
+		}
```
joshuasing: Was this temporary or for debugging? If not, we should use a useful error message here.

service/tbc/utxoindex.go
```diff
@@ -0,0 +1,341 @@
...
+		// This happens when a transaction is created and spent in the
+		// same block.
+		// XXX this is probably too loud but log for investigation and
+		// remove later.
+		log.Debugf("db missing pkscript: %v", op)
+		return
+	}
+	s.mtx.Lock()
+	utxos[op] = tbcd.NewDeleteCacheOutput(*sh, op.TxIndex())
+	s.mtx.Unlock()
```
joshuasing, line 203: Why does this use the server's mutex?

service/tbc/utxoindex.go
```diff
@@ -0,0 +1,341 @@
...
+				utxos[op] = tbcd.NewDeleteCacheOutput(tbcd.NewScriptHashFromScript(txOut.PkScript),
+					op.TxIndex())
+			}
+		}
+	}
+
+	return nil
+}
+
+func (s *Server) fetchOPParallel(ctx context.Context, c chan struct{}, w *sync.WaitGroup, op tbcd.Outpoint, utxos map[tbcd.Outpoint]tbcd.CacheOutput) {
```
joshuasing, line 181: Is there a reason for all of the below functions to be receivers on the `Server` struct? I believe these should all be separated unless they are explicitly required by the Server as well as the UTXO indexer. If they are required by both, ideally they should not be receivers.

The only usage of `s` I am seeing is for `s.mtx`, which I don't believe is necessary in most cases.

joshuasing (changes requested):

service/tbc/indexer.go
marcopeereboom: No java :-P

service/tbc/tbc.go
marcopeereboom: Dealer's choice.

service/tbc/indexer.go
marcopeereboom: Fixed.

service/tbc/indexer.go
marcopeereboom: Fixed

service/tbc/tbc.go
marcopeereboom: Removed.

service/tbc/utxoindex.go
marcopeereboom: I agree with this but let's do this in a separate PR.

service/tbc/utxoindex.go
marcopeereboom: left a XXX to do this in a separate PR.

service/tbc/geometry.go
```diff
@@ -0,0 +1,404 @@
...
+func (e NotLinearError) Error() string {
+	return string(e)
+}
+
+func (e NotLinearError) Is(target error) bool {
+	_, ok := target.(NotLinearError)
+	return ok
+}
+
+var ErrNotLinear = NotLinearError("not linear")
```
AL-CT, line 114: I don't think this is used outside of tests. Probably remove it or instead remove NotLinearError and instead:
```
var ErrNotLinear = errors.New("not linear")
```
And wrap ErrNotLinear using fmt.Errorf for specific use cases

service/tbc/geometry.go
```diff
@@ -0,0 +1,404 @@
...
+		if bhb.Hash.IsEqual(genesisHash) {
+			return false, nil
+		}
+		if bhb.Hash.IsEqual(&bh.Hash) {
+			return true, nil
+		}
+	}
+}
+
+// findCommonParent find the common parrent between the two provided
```
AL-CT:
```suggestion
// findCommonParent find the common parent between the two provided
```

service/tbc/indexer.go
```diff
@@ -0,0 +1,563 @@
...
+func (aie AlreadyIndexingError) Error() string {
+	return string(aie)
+}
+
+func (aie AlreadyIndexingError) Is(target error) bool {
+	_, ok := target.(AlreadyIndexingError)
+	return ok
+}
+
+var ErrAlreadyIndexing = AlreadyIndexingError("already indexing")
```
AL-CT: ErrAlreadyIndexing is the only instance of AlreadyIndexingErr
```suggestion
var ErrAlreadyIndexing = errors.New("already indexing")
```

service/tbc/indexer.go
```diff
@@ -0,0 +1,563 @@
...
+func (c *indexerCommon) evaluateBlockHeaderIndex(bh *tbcd.BlockHeader, err error) (*tbcd.BlockHeader, error) {
+	if err != nil {
+		if !errors.Is(err, database.ErrNotFound) {
+			return nil, err
+		}
+		bh = &tbcd.BlockHeader{
+			Hash:   *c.g.chain.GenesisHash,
+			Height: 0,
+			Header: h2b(&c.g.chain.GenesisBlock.Header),
+		}
```
AL-CT, line 172: Should this be hemiGenesis for keystoneIndexer?

marcopeereboom: yeah someone broke that methinks

marcopeereboom: This threw me for a loop but this is below it:
```
        } else {
                // Some indexers use a different genesis, e.g. keystones. Will
                // be nil if there is no override.
                if override := c.genesis; override != nil {
                        hh = override
                }
        }
```

service/tbc/geometry.go
marcopeereboom: No, we may need it.

PR #665, opened 2025-08-29
Add database interface for upcomming ZK work

marcopeereboom:
**Summary**
The upcoming ZK work will require extremely heavy database workloads. Implement a generic database layer that can be used with various key/value stores, including caches and in-memory stores. This way we can fine-tune whichever type of storage we need to whichever workload we encounter.

A nice side effect of this is that we can do some neat things. For example, we can create a database that replicates itself.


marcopeereboom:
This has been split into various PRs

PR #675, opened 2025-09-10
service/popm: replace exponential backoff

### LINKED ISSUE #664: popm: exponential backoff overflow
### Confirmation

### Describe the bug

PR #612 added logic for [exponential backoff to popm](https://github.com/hemilabs/heminetwork/blob/67084179716fe1ca530f42a23c529e555faeecc9/service/popm/popm.go#L655). However, the method used (byte-shifting), will eventually force delay to overflow, leading to delay being set to a negative number (preventing Tick() from ever continuing) or even 0:
```
delay := baseDelay * (1 << attempt)
if delay > maxDelay {
	delay = maxDelay
}
```
This is exacerbated by the jitter code, which will panic once the argument for Int64N() becomes too high:
```
jitter := int64(delay / 10)
delay += time.Duration(mathrand.Int64N(jitter*2) - jitter)
```

### Expected behaviour

popm should retry indefinitely without any issues

### Environment

Version: 
Operating System: 
Architecture: 

### Additional Information

_No response_

AL-CT:
**Summary**
Fixes #664 

**Changes**
As previously discussed, replaces the exponential backoff with a semi-random backoff.

On disconnect:
1. Attempt reconnect after `delay` + `jitter` (between 0 and 0.5 of `delay`)
2. Increase `delay` [between `minInc` to `minInc` * 3], up to `maxDelay`
3. If reconnect successful, reset `delay`

This should prevent network congestion whilst removing the complexity and problems of the previous algorithm.

### DIFF
--- a/service/popm/popm.go
+++ b/service/popm/popm.go
@@ -58,6 +58,8 @@ const (
 	defaultL2KeystonePollTimeout  = 13 * time.Second
 	defaultL2KeystoneRetryTimeout = 15 * time.Second
 
+	defaultMaxReconnectDelay = 15 * time.Second
+
 	minRelayFee = 1                // sats/byte
 	maxBlockAge = 30 * time.Second // XXX make this configurable?
 )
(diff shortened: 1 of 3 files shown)


service/popm/popm.go
```diff
@@ -631,10 +631,11 @@ func (s *Server) opgeth(ctx context.Context) {
...
 
-	const maxDelay = 15 * time.Second
-	baseDelay := s.cfg.opgethReconnectTimeout
-
-	var attempt int
+	const (
+		maxBaseDelay = int64(15 * time.Second)
+		minInc       = int64(1 * time.Second)
+	)
+	baseDelay := int64(s.cfg.opgethReconnectTimeout)
```
ClaytonNorthey92: hey @AL-CT , why is this a signed number? could we use unsigned?

ClaytonNorthey92 (changes requested):
one question, let's also add a test.

timing is a bit difficult in tests, but I think that this is a good example: https://github.com/hemilabs/dave/pull/11/files#diff-660f03c7c68e5e818ca4927ba8bcf1cad4c031d90c006c6c9ac16678630d7670R302

marcopeereboom (changes requested):
Please just make this random.

service/popm/popm.go
AL-CT: Both done 👍

PR #677, opened 2025-09-15
tbc: add ZK indexers

marcopeereboom:
**Summary**
Add a journal of bitcoin transactions, blocks, etc. This enables future ZK provability of the bitcoin blockchain.

**Changes**
All greenfield.


database/tbcd/level/level.go
```diff
@@ -2193,6 +2194,218 @@ func (l *ldb) BlockHeaderByTxIndex(ctx context.Context) (*tbcd.BlockHeader, erro
...
+		}
+		so := tbcd.ZKSpentOutput{
+			ScriptHash:        tbcd.NewScriptHashFromBytesP(k[:32]),
+			BlockHeight:       binary.BigEndian.Uint32(k[32:]),
+			BlockHash:         bytes2hash(k[32+4 : 32+4+32]),
+			TxID:              bytes2hash(k[32+4+32 : 32+4+32+32]),
+			PrevOutpointHash:  bytes2hash(k[32+4+32+32 : 32+4+32+32+32]),
+			PrevOutpointIndex: binary.BigEndian.Uint32(k[32+4+32+32+32:]),
+			TxInIndex:         binary.BigEndian.Uint32(k[32+4+32+32+32+4:]),
+		}
```
joshuasing, line 2285: Could we turn these into constants, so they are easier to understand and so we're less likely to make mistakes in the future? Also, this would allow us to document what each of the number parts are.

_**Same for `tbcd.ZKSpendingOutpoint` and `tbcd.ZKSpendableOutput` below**_

service/tbc/tbc.go
```diff
@@ -3100,6 +3167,16 @@ func (s *Server) Run(pctx context.Context) error {
...
+		if s.cfg.ZKIndex {
+			bh, _ := s.zki.IndexerAt(ctx)
+			log.Infof("ZK utxo index %v @ %v", bh.Height, bh.Hash)
+		}
+
+		//// XXX remove
+		//err := s.SyncIndexersToBest(ctx)
+		//if err != nil {
+		//	panic(err)
+		//}
```
joshuasing: Should this be removed?

service/tbc/zkindexer.go
```diff
@@ -0,0 +1,235 @@
...
+		// Handle balance
+		switch direction {
+		case 1:
+			cache[tbcd.ZKIndexKey(sh[:])] = tbcd.BESubUint64(balance,
+				value)
+		case -1:
+			cache[tbcd.ZKIndexKey(sh[:])] = tbcd.BEAddUint64(balance,
+				value)
+		default:
+			panic("wtf")
```
joshuasing: We should be using actual error messages for these:
```suggestion
			panic(fmt.Errorf("bug: invalid direction: %v", direction))
```

***Same for other `panic("wtf")` calls***

service/tbc/zkindexer_test.go
```diff
@@ -0,0 +1,200 @@
...
+		t.Fatalf("expected %v, got %v", balance, rv)
+	}
+
+	// SpentOut
+	out, err := db.ZKSpentOutputs(ctx, sh)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(out) != 1 {
+		t.Fatalf("expected 1 spent output, got %d", len(out))
```
joshuasing: Go generally prefers "got before want", and identifying the function in tests, per:
https://go.dev/wiki/TestComments#got-before-want
https://go.dev/wiki/TestComments#identify-the-function

This change makes the failure easier to debug and a lot more descriptive, whilst being more idiomatic:
```suggestion
		t.Fatalf("ZKSpentOutputs out len = %d, want 1", len(out))
```

---

Go avoids "expected, got" or "expected, actual", and opts for "got X, want Z" (https://go.dev/wiki/TestComments#got-before-want) or better `MyFunc(%q) = %v, want %v` (https://go.dev/wiki/TestComments#identify-the-function)

Also, it is generally preferred to use `t.Error` instead of `t.Fatal` when possible (as long as the test can actually continue) and/or using `t.Run` to create sub-tests to avoid failing at the first error, making it easier to figure out what broke.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Add ZK indexers.
```
joshuasing, line 12: I think this needs to be a lot more descriptive. Can we document exactly why these were added, and what they will be used for?

Also best if we can include `([#677](https://github.com/hemilabs/heminetwork/pull/677))`

database/tbcd/level/level.go
marcopeereboom: No because then it is no longer readable. There are times where magic numbers are preferred, this is one of those times. 32 is a hash, 4 is a height and 8 is a value. This is common practice when hand rolling encoders so that you can easily spot numbers that are off. Virtually no one ever reads a constant value, they simply assume it is correct.

service/tbc/zkindexer.go
marcopeereboom: meh but committed a lame diagnostic message

marcopeereboom: sure, fixed

CHANGELOG.md
marcopeereboom: This is the first iteration of the indexers and it will change. I am not going to speculate over and over this. It now has a place holder so that we don't forget when churning changelog.

service/tbc/tbc.go
marcopeereboom: Updated the comment explaining it.

service/tbc/zkindexer_test.go
marcopeereboom: I am not a huge fan of the continue to run a test and having to spend hours digging through logs to find it (or not when dealing with a go routine panic). This is especially true using dogshit like github where you can't search through the logs. While I don't disagree with the neatness of this, spending human time is always worse than machine time. Fail hard and often is IMO the superior of the two modes when it comes to test code. On top of that, this test should not ever fail once it has been written.

database/tbcd/level/level.go
joshuasing: > 32 is a hash, 4 is a height and 8 is a value

This isn't immediately obvious to someone who is not familiar with this code. Can we add a comment like this explaining what the numbers are?

service/tbc/zkindexer_test.go
AL-CT: Thanks 👍 change the order of the outputs, but will keep t.Fatal for now. I have used t.Error in some of the larger tests with subtests, or when we want to know multiple points of failure. In this instance though, if the general encoding of zkKeys got changed for example, I imagine dozens of these might be triggered, so I think it might end up just being noise for whoever fixes it.

database/tbcd/level/level.go
marcopeereboom: I am going to disagree that this isn't self-evident. The word says `ScriptHash` and is 32, the other is `BlockHeight` and is 4 etc. Encoders are notoriously hard to read and often write-once code; this is also the reason most people use computer generated code for this but these are database keys and thus we must make it readable for humans. When I look at the structure and a hex dump i can immediately see where what is at.

Sometimes over-documenting has the opposite effect and tends to become stale over time, code changes but no ever changes the comments. Unfortunately, not everyone is as diligent as you.

service/tbc/tbc.go
```diff
@@ -2656,10 +2683,37 @@ func (s *Server) synced(ctx context.Context) (si SyncInfo) {
...
+		// Perform additional zk indexer tests.
+		zkBH, err := s.zki.IndexerAt(ctx)
+		if err != nil {
+			zkBH = &tbcd.BlockHeader{}
+		}
+		zkHH := &HashHeight{
+			Hash:   zkBH.Hash,
+			Height: zkBH.Height,
+		}
+		si.ZK = *zkHH
```
joshuasing, line 2698: Is there any reason to create a `zkHH` variable pointer here and then dereference it?
```suggestion
		si.ZK = HashHeight{
			Hash:   zkBH.Hash,
			Height: zkBH.Height,
		}
```

service/tbc/tbc.go
```diff
@@ -2656,10 +2683,37 @@ func (s *Server) synced(ctx context.Context) (si SyncInfo) {
...
+		// If keystone and zk indexers are disabled we are synced.
+		if !s.cfg.HemiIndex && !s.cfg.ZKIndex {
 			si.Synced = true
 			return
 		}
+		if !(s.cfg.HemiIndex && si.Keystone.Hash.IsEqual(&bhb.Hash)) {
+			// Keystone index not synced
+			return
+		}
+		if !(s.cfg.ZKIndex && si.ZK.Hash.IsEqual(&bhb.Hash)) {
```
joshuasing, line 2712: De Morgan's law (usually I wouldn't comment, but here I think this is actually more readable):

```suggestion
		if !s.cfg.HemiIndex || !si.Keystone.Hash.IsEqual(&bhb.Hash) {
			// Keystone index not synced
			return
		}
		if !s.cfg.ZKIndex || !si.ZK.Hash.IsEqual(&bhb.Hash) {
```

marcopeereboom: Never. !() always wins over mixing ! ||

service/tbc/tbc.go
marcopeereboom: Yes, it uses the return value of  `zkBH, err := s.zki.IndexerAt(ctx)` to create the variable. We just reuse it later if an error occurred. It's dereference there or elsewhere.

cmd/tbcd/tbcd.go
```diff
@@ -85,16 +85,22 @@ var (
...
 			Value:        &cfg.MaxCachedKeystones,
-			DefaultValue: int(1e6),
+			DefaultValue: int(2e6),
 			Help:         "maximum cached keystones during indexing",
 			Print:        config.PrintAll,
 		},
 		"TBC_MAX_CACHED_TXS": config.Config{
 			Value:        &cfg.MaxCachedTxs,
-			DefaultValue: int(1e6),
+			DefaultValue: int(2e6),
```
joshuasing: Is there a reason to do this change in this pull request? Would we be able to do this in a separate PR with a separate changelog entry?

service/tbc/indexer.go
```diff
@@ -455,6 +455,13 @@ func (c *indexerCommon) parseBlocks(ctx context.Context, endHash *chainhash.Hash
 			break
 		}
 
+		// Check for sigterm.
+		select {
+		case <-ctx.Done():
+			return 0, last, ctx.Err()
+		default:
+		}
```
joshuasing, line 463: This can be simplified as:

```suggestion
		if err := ctx.Err(); err != nil {
			return 0, last, err
		}
```

AL-CT (approved):

PR #691, opened 2025-09-30
hproxy: add method whitelisting

AL-CT:
**Summary**
Creates an ethereum RPC method whitelist to hproxy (e.g., `eth_getBlockByNumber`).

**Changes**
Adds a new field to the hproxy server cfg where each allowed method must be explicitly passed.

**Notes**
We decode each request into an `EthereumRequest`, checking the method. This prevents requests that do not comply to the ethereum RPC message structure to be denied. I think this is fine, and maybe even preferred, as other parts of the code also make the assumption that all nodes are Ethereum Nodes.


service/hproxy/hproxy.go
```diff
@@ -566,6 +601,26 @@ func (s *Server) handleProxyRequest(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		return err
+	}
+	r.Body = io.NopCloser(bytes.NewReader(data))
+
+	var j EthereumRequest
+	if err := json.NewDecoder(bytes.NewReader(data)).Decode(&j); err != nil {
+		return errors.New("unknown request type")
+	}
+	if _, ok := s.whitelist[j.Method]; !ok {
```
ClaytonNorthey92, line 653: nit: a common practice is to make the map of type `map[string]bool`, since `false` is the default for bools, you could then clean this up a bit like
```go
if !s.whitelist[j.Method] { ... }
```

service/hproxy/hproxy.go
```diff
@@ -170,6 +188,8 @@ type Server struct {
 	hvmHandlers []HVMHandler       // hvm nodes
 	clients     map[string]*client // [ip_address]server_id
 
+	whitelist map[string]interface{} // method filter
```
joshuasing: I think we could use an empty struct (`struct{}`) here to make sure this doesn't use any resources: https://dave.cheney.net/2014/03/25/the-empty-struct
```suggestion
	whitelist map[string]struct{} // method filter
```

service/hproxy/hproxy.go
```diff
@@ -566,6 +601,26 @@ func (s *Server) handleProxyRequest(w http.ResponseWriter, r *http.Request) {
 //	}
 // }
 
+func (s *Server) filterRequest(r *http.Request) error {
+	// copy and reset body
+	const byteLimit = 5 * 1024 * 1024 // 5 MiB
```
marcopeereboom: const on top of file.

service/hproxy/hproxy.go
```diff
@@ -198,6 +218,7 @@ func NewServer(cfg *Config) (*Server, error) {
 			Name:      "proxy_calls",
 			Help:      "The total number of successful proxy calls",
 		}),
+		whitelist: make(map[string]any, len(cfg.MethodFilter)),
```
marcopeereboom: does whitelist need locks?

service/hproxy/hproxy.go
```diff
@@ -566,6 +601,26 @@ func (s *Server) handleProxyRequest(w http.ResponseWriter, r *http.Request) {
 //	}
 // }
 
+func (s *Server) filterRequest(r *http.Request) error {
+	// copy and reset body
+	const byteLimit = 5 * 1024 * 1024 // 5 MiB
+	lr := io.LimitReader(r.Body, byteLimit)
+	data, err := io.ReadAll(lr)
+	if err != nil {
+		return err
```
marcopeereboom, line 645: does this fail if byteLimit is hit?

service/hproxy/hproxy_test.go
```diff
@@ -316,7 +338,30 @@ func TestProxy(t *testing.T) {
...
+	time.Sleep(250 * time.Millisecond)
+
+	// Send invalid command
+	err := request(t.Context(), serverID, hpCfg, "invalid")
+	if !errors.Is(err, ErrForbiddenMethod) {
+		t.Fatalf("expected %v, got %v", ErrForbiddenMethod, err)
+	}
+
+	// Send valid command
+	err = request(t.Context(), serverID, hpCfg, "valid")
```
ClaytonNorthey92: nit: this should be a separate test, perhaps a table-driven test of different methods and configs, we're also missing tests for if the list is omitted

service/hproxy/hproxy.go
```diff
@@ -462,6 +487,16 @@ func (s *Server) handleProxyRequest(w http.ResponseWriter, r *http.Request) {
 
 	startTime := time.Now()
 
+	if err := s.filterRequest(r); err != nil {
+		if errors.Is(err, ErrForbiddenMethod) {
+			w.Header().Set("Allow", strings.Join(s.cfg.MethodFilter, ", "))
+			w.WriteHeader(http.StatusMethodNotAllowed)
+			return
+		}
+		w.WriteHeader(http.StatusBadRequest)
```
ClaytonNorthey92: could we log the error out here? if people start complaining "hey we're only getting `StatusBadRequest`s!", we need to debug

service/hproxy/hproxy.go
```diff
@@ -462,6 +487,16 @@ func (s *Server) handleProxyRequest(w http.ResponseWriter, r *http.Request) {
 
 	startTime := time.Now()
 
+	if err := s.filterRequest(r); err != nil {
+		if errors.Is(err, ErrForbiddenMethod) {
+			w.Header().Set("Allow", strings.Join(s.cfg.MethodFilter, ", "))
+			w.WriteHeader(http.StatusMethodNotAllowed)
```
joshuasing: I am not sure HTTP Status 405 is appropriate for a JSON-RPC method not being allowed - I think it is only recommended for use when the HTTP method is not allowed.

It looks like `proxyd` writes a JSON-RPC error, which is what we should probably do: https://github.com/ethereum-optimism/infra/blob/0ef294beca024a689069fb0cbc8e31cda830ac6b/proxyd/backend.go#L49-L53

(same for the BadRequest below most likely)

service/hproxy/hproxy.go
marcopeereboom: That's a DDOS waiting to happen. We certainly should make it Debugf

service/hproxy/hproxy.go
```diff
@@ -566,6 +601,26 @@ func (s *Server) handleProxyRequest(w http.ResponseWriter, r *http.Request) {
 //	}
 // }
 
+func (s *Server) filterRequest(r *http.Request) error {
+	// copy and reset body
+	const byteLimit = 5 * 1024 * 1024 // 5 MiB
+	lr := io.LimitReader(r.Body, byteLimit)
```
joshuasing: If a request exceeds 5 MiB, the `json.NewDecoder` will likely fail to decode (e.g. unclosed `}`s). Do we want to add some kind of global request size limiter, and use that here? Then all requests exceeding that can be blocked 🤔

service/hproxy/hproxy.go
joshuasing: I think `_, ok :=` is more idiomatic and far more common in Go stdlib - please correct me if I am wrong.
This also means we can use `map[string]struct{}` which uses far less resources.

service/hproxy/hproxy.go
marcopeereboom: Agreed

service/hproxy/hproxy.go
marcopeereboom: Yes please

CHANGELOG.md
```diff
@@ -34,7 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add **hproxyd**, a simple and efficient RPC request proxy to replace
   `proxyd` ([#568](https://github.com/hemilabs/heminetwork/pull/568), [#574](https://github.com/hemilabs/heminetwork/pull/574),
-  [#576](https://github.com/hemilabs/heminetwork/pull/576)).
+  [#576](https://github.com/hemilabs/heminetwork/pull/576), [#691](https://github.com/hemilabs/heminetwork/pull/691)).
```
joshuasing: This is the changelog for v2.0.0, which this commit will not be included in.

Could we add a separate entry under the `[Unreleased]` -> `Changed` or `Added` section (see https://keepachangelog.com/en/1.1.0/ for examples)

docker/hproxyd/goreleaser.Dockerfile
```diff
@@ -33,6 +33,7 @@ COPY $TARGETPLATFORM/hproxyd /usr/local/bin/hproxyd
 # Environment variables
 ENV HPROXY_CLIENT_IDLE_TIMEOUT=""
 ENV HPROXY_HVM_URLS=""
+ENV HPROXY_METHOD_WHITELIST=""
```
joshuasing, line 36: Also needed in `Dockerfile` I think.

service/hproxy/hproxy.go
joshuasing: Also see https://github.com/hemilabs/heminetwork/pull/691/files#r2391359581

service/hproxy/hproxy.go
AL-CT: It will only read up to the amount of bytes set in the `LimitReader`. It's there to prevent any malicious requests that might be extremely large

service/hproxy/hproxy.go
AL-CT: added a justification for why not

service/hproxy/hproxy.go
AL-CT: Fixed 👍

service/hproxy/hproxy.go
AL-CT: In this case I agree with Joshua, that's generally the convention I have seen most often, and does save space 👍

service/hproxy/hproxy.go
AL-CT: Good point, fixed 👍

CHANGELOG.md
AL-CT: Oops, apologies. Fixed 👍

docker/hproxyd/goreleaser.Dockerfile
AL-CT: Yup, added 👍

service/hproxy/hproxy.go
AL-CT: As discussed, will create an [issue](https://github.com/hemilabs/heminetwork/issues/693) to make all of these responses into JSON-RPC, so it can be addressed in a separate PR. Meanwhile, will change the HTTP Status to `403` (Forbidden).

EDIT: Also added a few other possible status codes

service/hproxy/hproxy.go
AL-CT: Fixed 👍

service/hproxy/hproxy.go
marcopeereboom: But I asked if it fails :-)

service/hproxy/hproxy.go
AL-CT: Also added a few other possible status codes

service/hproxy/hproxy.go
AL-CT: Added a check for the request size, will that work? Still keeping the limit reader in case the `contentLength` is -1 and thus size is unknown.

service/hproxy/hproxy_test.go
AL-CT: Extended the test cases 👍

ClaytonNorthey92 (approved):

service/hproxy/hproxy.go
AL-CT: Oh, sorry 🙂
It will not, but if in reading less than the total it becomes unreadable, it will fail when decoding.
Also, added a different check above it

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Method whitelist to hproxy ([#691](https://github.com/hemilabs/heminetwork/pull/691))
```
joshuasing: More consistent with other entries:
```suggestion
- Add RPC request method whitelist to hproxy ([#691](https://github.com/hemilabs/heminetwork/pull/691)).
```

service/hproxy/hproxy.go
AL-CT: Replaced with MaxBytesReader so it does fail

service/hproxy/hproxy.go
AL-CT: Changed the code to use `http.MaxBytesReader`, so any request we try to read that is above the set amount, will fail as we read it, and not after.

service/hproxy/hproxy.go
```diff
@@ -566,6 +622,28 @@ func (s *Server) handleProxyRequest(w http.ResponseWriter, r *http.Request) {
 //	}
 // }
 
+func (s *Server) filterRequest(r *http.Request) error {
+	// copy and reset body
+	data, err := io.ReadAll(r.Body)
+	if err != nil {
+		maxBytesError := &http.MaxBytesError{}
+		if errors.As(err, &maxBytesError) {
```
joshuasing, line 642: I think we can do this:
```suggestion
		var maxBytesError http.MaxBytesError
		if errors.As(err, &maxBytesError) {
```

AL-CT: Apparently not: `second argument to errors.As must be a non-nil pointer to either a type that implements error, or to any interface type`

cmd/hproxyd/hproxyd.go
```diff
@@ -85,6 +91,12 @@ var (
 			Help:         "frequency that hproxy pokes nodes for health information",
 			Print:        config.PrintAll,
 		},
+		"HPROXY_REQUEST_SIZE": config.Config{
+			Value:        &cfg.MaxRequestSize,
+			DefaultValue: cfg.MaxRequestSize,
```
marcopeereboom, line 96: Is this humanized? if not we should do that.

AL-CT: Fixed 👍

service/hproxy/hproxy.go
joshuasing: Interesting, okay. Thanks!

PR #704, opened 2025-10-15
localnet: updated deprecated flag in forge when building op-stack image for localnet

ClaytonNorthey92:
**Summary**
the `--deny-warnings` flag is deprecated, and the optimism files that are built fail linting.  

**Changes**
update the flag to `--deny never` to allow building to succeed.


ClaytonNorthey92:
localnet test passing here https://github.com/hemilabs/heminetwork/actions/runs/18541527579/job/52850040517

Issue #137, opened 2024-05-30
Add notifications to TBC (and tbcapi)


joshuasing:
### Confirmation

- [X] I have checked for similar issues.

### Problem

There is currently no way of being notified when TBC receives a new block, and other events.

### Suggested solution

It would be useful to add notifications to TBC, both in the Go API and the WebSocket-based RPC API.

The TBC Go API should allow users to call a function with a channel or callback function that is then passed notifications whenever events happen in TBC.

Additionally, RPC clients should be able to send a command to subscribe to certain types of notifications (possibly with filters or something in the future) and then start receiving those notifications as events happen in TBC.

PR #725, opened 2025-10-22
tbc: add notifications

AL-CT:
**Summary**
Mostly addresses #137

Adds a notifier to TBC, which can be subscribed to in order to receive notifications when certain events occur within TBC.
Fixed several tests in the `tbc` package that relied on `time.sleep` for concurrent synchronization.

I think more tests can be changed to utilize this, but I will create an issue to address them in a separate PR.

**Changes**
- Create a `Notifier` in `tbc` that can be subscribed to in order to receive notifications from TBC.
- Wrap specific database calls in TBC to ensure notifications are sent.
- Fix several fragile tests in the `tbc` package by using the notification system.

### DIFF
--- a/service/tbc/notifier.go
+++ b/service/tbc/notifier.go
@@ -97,6 +97,13 @@ func NewNotifier(blocking bool) *Notifier {
 	return &n
 }
 
+func (n *Notifier) HasListeners() bool {
+	n.mtx.Lock()
+	defer n.mtx.Unlock()
+
+	return len(n.listeners) > 0
+}
+
 func (n *Notifier) Subscribe(pctx context.Context, capacity uint64) (*Listener, error) {
 	lctx, cancel := context.WithCancel(pctx)
 	l := &Listener{
--- a/service/tbc/tbc.go
+++ b/service/tbc/tbc.go
@@ -1624,9 +1624,11 @@ func (s *Server) insertBlock(ctx context.Context, block *btcutil.Block) (int64, 
 	if err != nil {
 		return 0, err
 	}
-	msg := NotificationBlock(*block.Hash())
-	if err := s.notifier.Notify(ctx, msg); err != nil {
-		return 0, fmt.Errorf("send new block notification: %w", err)
+	if s.notifier.HasListeners() {
+		msg := NotificationBlock(*block.Hash())
+		if err := s.notifier.Notify(ctx, msg); err != nil {
+			return 0, fmt.Errorf("send new block notification: %w", err)
+		}
 	}
 	return height, nil
 }
@@ -1636,11 +1638,13 @@ func (s *Server) insertBlockheader(ctx context.Context, headers *wire.MsgHeaders
 	if err != nil {
 		return it, cbh, lbh, n, err
 	}
-	for _, h := range headers.Headers {
-		msg := NotificationBlockheader(h.BlockHash())
-		if err := s.notifier.Notify(ctx, msg); err != nil {
-			return it, cbh, lbh, n,
-				fmt.Errorf("send new blockheader notification: %w", err)
+	if s.notifier.HasListeners() {
+		for _, h := range headers.Headers {
+			msg := NotificationBlockheader(h.BlockHash())
+			if err := s.notifier.Notify(ctx, msg); err != nil {
+				return it, cbh, lbh, n,
+					fmt.Errorf("send new blockheader notification: %w", err)
+			}
 		}
 	}
 	return it, cbh, lbh, n, err


service/tbc/tbc.go
```diff
@@ -1607,12 +1619,35 @@ func (s *Server) handleHeaders(ctx context.Context, p *rawpeer.RawPeer, msg *wir
 	return nil
 }
 
+func (s *Server) insertBlockAndNotify(ctx context.Context, block *btcutil.Block) (int64, error) {
```
marcopeereboom: 'insertBlock' is just fine

service/tbc/tbc.go
```diff
@@ -1607,12 +1619,35 @@ func (s *Server) handleHeaders(ctx context.Context, p *rawpeer.RawPeer, msg *wir
...
+	if err != nil {
+		return 0, err
+	}
+	if err := s.notifier.Notify(ctx, NotificationBlock); err != nil {
+		return 0, fmt.Errorf("send new block notification: %w", err)
+	}
+	return height, nil
+}
+
+func (s *Server) insertBlockheaderAndNotify(ctx context.Context, headers *wire.MsgHeaders, batchHook tbcd.BatchHook) (tbcd.InsertType, *tbcd.BlockHeader, *tbcd.BlockHeader, int, error) {
```
marcopeereboom: Same here, drop 'AndNotify'

service/tbc/notifier.go
```diff
@@ -0,0 +1,110 @@
...
+				cancel:   cancel,
+			}
+			n.listeners[string(nid[:])] = &l
+			return &l, nil
+		}
+	}
+}
+
+func (n *Notifier) Notify(ctx context.Context, message notification) error {
+	n.mtx.Lock()
```
marcopeereboom, line 143: maybe a bit of a weird idea but maybe use pctx here instead? Defining ctx inside the struct is always weird and generally should be avoided, maybe the p makes that more obvious?

service/tbc/notifier.go
```diff
@@ -0,0 +1,110 @@
...
+	defer n.mtx.Unlock()
+
+	for _, l := range n.listeners {
```
marcopeereboom, line 146: this is clever but if you have 10 blockers and the first is slow...

Maybe that was the intent all along but this needs documenting.

service/tbc/notifier.go
```diff
@@ -0,0 +1,110 @@
...
+package tbc
+
+import (
+	"context"
+	"crypto/rand"
+	"io"
+	"sync"
+)
+
+type notification string
```
marcopeereboom: Should this be modeled after context? Thus chan+error?

service/tbc/notifier.go
```diff
@@ -0,0 +1,110 @@
...
+
+const (
+	NotificationBlock       notification = "block inserted"
+	NotificationBlockheader notification = "blockheader inserted"
+)
```
marcopeereboom: This needs some thought. I do think we need a notification type that has a String that returns a human readable notification. But we do need context. Like `block inserted 4398578394579843578943579843`.

We can make this:
```
type Notification Struct {
  Type string // the const above
  ID string // Or chainhash or any?  string may be nice to leave interpretation to caller
  Timestamp time.Time // Always :-)
  Error error // nil when shit's good
}
```

marcopeereboom (changes requested):
The only thing I really tripped over was the java style naming convention but have a look at the suggestions and use the ones you find helpful.

service/tbc/notifier.go
```diff
@@ -0,0 +1,110 @@
...
+	l.cancel()
+
+	l.notifier.mtx.Lock()
+	defer l.notifier.mtx.Unlock()
+
+	delete(l.notifier.listeners, l.id)
+}
+
+func (l *Listener) Listen() <-chan notification {
+	return l.ch
```
ClaytonNorthey92: what happens if two gorouintes each call `.Listen()`? they may both expect notifications but only one of them will receive them

example:
https://go.dev/play/p/q9v9YQm4yup

I think we should either: 
1. return an error if more than one goroutine  calls `.Listen()`
2. allow multiple goroutines to call `.Listen()`

what do you think?

service/tbc/notifier.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (n *Notifier) Notify(ctx context.Context, message notification) error {
+	n.mtx.Lock()
+	defer n.mtx.Unlock()
+
+	for _, l := range n.listeners {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case l.ch <- message:
```
ClaytonNorthey92, line 150: will this _block_ if nothing is listening on this channel? thus blocking the following listeners in the slice?  if so, I think we should have a `case default`, and if there's nobody listening on the channel then they miss it 🤷

service/tbc/notifier.go
marcopeereboom: I like this. Panic is even fine :-)

AL-CT: That's a good point. I'll add a comment saying that `Listen` can't be called by multiple goroutines, and panic / return error if you do 👍

ClaytonNorthey92: @AL-CT you're the owner, you decide what's best! I would be on board with a panic as @marcopeereboom suggested 👍

service/tbc/notifier.go
AL-CT: I decided against going with `default` because we might want a notifier to block or not, depending on the situation:

When creating a notifier, you explicitly say whether it is blocking or not, considering you should only make it blocking during tests and not in production (hence not making that something you can change from `tbcd`, only by manually setting it in the TBC configs).

If it is blocking, sending a notification will block if it can't send notifications to a listener, no matter how many there are. This is on purpose, so that you can control the flow of tests. 

If the Notifier isn't blocking, then the case below that (a call to `block()`), will ensure that this continues, essentially working as default. Note that if the Notifier is blocking, the `block()` call will only be activated if the listener is marked for deletion, in which case we skip it even if it is not actively listening for messages.

service/tbc/tbc.go
AL-CT: Fixed 👍

service/tbc/tbc.go
AL-CT: Fixed 👍

service/tbc/notifier.go
AL-CT: Yup, agreed, will change 👍

service/tbc/notifier.go
ClaytonNorthey92: ah ok I see it now, thanks!

service/tbc/notifier.go
AL-CT: Changed `Listen` to return the next message from the channel, and panic when called by multiple goroutines. Could have just made it an error, but if you're using `Listen` with multiple goroutines then you likely made a mistake in the code, and should instead make another listener.

service/tbc/notifier.go
AL-CT: Changed the notification structure 👍

service/tbc/notifier.go
AL-CT: I added some docu for this 👍

service/tbc/notifier.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+type Listener struct {
+	id string
+	ch chan Notification
+
+	listening atomic.Bool
+
+	ctx      context.Context
+	callback func()
```
joshuasing: Maybe `unsubscribe` would be a clearer name for this 🤔 
```suggestion
	unsubscribe func()
```

service/tbc/notifier.go
```diff
@@ -0,0 +1,157 @@
...
+
+func (n *Notifier) Subscribe(pctx context.Context, capacity uint64) (*Listener, error) {
+	n.mtx.Lock()
+	defer n.mtx.Unlock()
+
+	var nid [16]byte
+	for {
+		if _, err := io.ReadFull(rand.Reader, nid[:]); err != nil {
+			return nil, err
+		}
```
joshuasing: Using `string(nid[:])` may result in invalid characters. Instead, maybe we could add a `genID()` helper function which generates this ID and uses hex-encoded bytes (could also use base64, etc.).

```go
func genID() string {
	buf := make([]byte, 24)
	if _, err := rand.Read(buf); err != nil {
		panic(fmt.Errorf("read random: %w", err))
	}
	return hex.EncodeToString(buf)
}
```

It is safe (and sometimes recommended) to use `panic` if `rand.Read` fails, as the function will always panic internally if an error occurs:
```go
	if err != nil {
		fatal("crypto/rand: failed to read random data (see https://go.dev/issue/66821): " + err.Error())
		panic("unreachable") // To be sure.
	}
```

service/tbc/notifier.go
```diff
@@ -0,0 +1,157 @@
...
+		if _, ok := n.listeners[string(nid[:])]; !ok {
+			lctx, cancel := context.WithCancel(pctx)
+			l := Listener{
```
joshuasing: If `l` is created as a pointer here, we can avoid `&l` twice below I think.
```suggestion
			l := &Listener{
```

service/tbc/notifier.go
AL-CT: Fair point, fixed 👍

service/tbc/notifier.go
AL-CT: Oh, interesting. Great suggestion, thank you! Fixed 👍

service/tbc/notifier.go
AL-CT: Fixed 👍

service/tbc/notifier.go
AL-CT: I believe this is addressed by the Notification struct change

service/tbc/tbc.go
```diff
@@ -1607,12 +1619,39 @@ func (s *Server) handleHeaders(ctx context.Context, p *rawpeer.RawPeer, msg *wir
 	return nil
 }
 
+func (s *Server) insertBlock(ctx context.Context, block *btcutil.Block) (int64, error) {
+	height, err := s.g.db.BlockInsert(ctx, block)
+	if err != nil {
+		return 0, err
+	}
+	msg := NotificationBlock(*block.Hash())
```
marcopeereboom: Here and below we should look ifnotifications are required and exit early if they are not. We shouldn;t be creating a message that os tossed because there is no one home.

Prob something like if `len(listeners) == 0`

marcopeereboom (approved):

service/tbc/tbc.go
AL-CT: Fixed 👍 added a method since the listeners map is protected

Issue #718, opened 2025-10-20
Flakey tests


joshuasing:
This issue is to track flakey tests across the repository.

joshuasing:
https://github.com/hemilabs/heminetwork/actions/runs/18662839755/job/53207190555?pr=717
```
--- FAIL: TestNotFoundError (0.05s)
    --- FAIL: TestNotFoundError/BlockKeystoneByL2KeystoneAbrevHash (0.00s)
panic: runtime error: index out of range [0] with length 0 [recovered, repanicked]

FAIL	github.com/hemilabs/heminetwork/v2/service/tbc	136.241s
```

joshuasing:
https://github.com/hemilabs/heminetwork/actions/runs/18662839755/job/53207190555?pr=717#step:5:39425
```
--- FAIL: TestDisconnectedOpgeth (45.00s)
2025-10-20 19:37:16 TRACE popm popm.go:661 opgeth exit
2025-10-20 19:37:16 INFO popm popm.go:993 pop miner has shutdown cleanly

panic: context deadline exceeded

FAIL	github.com/hemilabs/heminetwork/v2/service/popm	47.819s
```

---
Another occurrence: https://github.com/hemilabs/heminetwork/actions/runs/18662804587/job/53207080702?pr=715#step:5:1013

```
panic: context deadline exceeded

goroutine 240 [running]:
github.com/hemilabs/heminetwork/v2/service/popm.TestDisconnectedOpgeth.func1()
	/home/runner/work/heminetwork/heminetwork/service/popm/popm_test.go:337 +0x65
created by github.com/hemilabs/heminetwork/v2/service/popm.TestDisconnectedOpgeth in goroutine 237
	/home/runner/work/heminetwork/heminetwork/service/popm/popm_test.go:335 +0x5d2
FAIL	github.com/hemilabs/heminetwork/v2/service/popm	47.560s
```

---
Another: https://github.com/hemilabs/heminetwork/actions/runs/18660859115/job/53200775057?pr=710#step:5:1027

```
panic: context deadline exceeded

goroutine 267 [running]:
github.com/hemilabs/heminetwork/v2/service/popm.TestDisconnectedOpgeth.func1()
	/home/runner/work/heminetwork/heminetwork/service/popm/popm_test.go:337 +0x65
created by github.com/hemilabs/heminetwork/v2/service/popm.TestDisconnectedOpgeth in goroutine 264
[...]

PR #739, opened 2025-11-03
popm: fix TestDisconnectedOpgeth

AL-CT:
**Summary**
Fixes the test `TestDisconnectedOpgeth`, as detailed in #718.

The test was run 100 times to ensure its viability: https://github.com/hemilabs/heminetwork/actions/runs/19041422367/job/54379408025

**Changes**
- Fixed a locking issue when closing websocket connections.
- Closing the connections did not necessarily prevent requests from being handled within a for loop, so added an exit condition when server is closed.
- Refactored the test to prevent other possible edge-cases and improve logging.


service/popm/popm_test.go
```diff
@@ -351,22 +350,43 @@ func TestDisconnectedOpgeth(t *testing.T) {
...
+		case <-msgCh:
+		case rcvErr := <-errCh:
+			// miner has tried to reconnect
+			if errors.Is(rcvErr, mock.ErrConnectionClosed) {
+				t.Log("popminer attempted to reconnect")
+				opgeth.Start()
+				attempted = true
+			}
+		}
+	}
```
joshuasing: I think this can be simplified, maybe like below.
```suggestion
	var rcvErr error
	for {
		select {
		case <-ctx.Done():
			t.Fatal(ctx.Err())
		case <-msgCh:
			continue
		case rcvErr = <-errCh:
		}

		// miner has tried to reconnect
		if errors.Is(rcvErr, mock.ErrConnectionClosed) {
			t.Log("popminer attempted to reconnect")
			opgeth.Start()
			break
		}
	}
```

service/popm/popm_test.go
```diff
@@ -351,22 +350,43 @@ func TestDisconnectedOpgeth(t *testing.T) {
...
+	for flushed := false; flushed != true; {
+		select {
+		case <-ctx.Done():
+			t.Fatal(ctx.Err())
+		case <-errCh:
+		default:
+			t.Log("error messages flushed")
+			flushed = true
+		}
 	}
```
joshuasing: I think this loop can be simplified, maybe like this:
```suggestion
	for {
		select {
		case <-ctx.Done():
			t.Fatal(ctx.Err())
		case <-errCh:
			continue
		default:
		}

		t.Log("error messages flushed")
		break
}
```

joshuasing (changes requested):

PR #746, opened 2025-11-11
update geth l1 in localnet and localnode to 16.7

ClaytonNorthey92:
**Summary**
update geth l1 in localnet and localnode to 16.7, this requires a few additional small fixes:
* increase test timeout in CI to 60m, it doesn't take this long, but it takes longer than 30m (the previous value)
* implement retries where before we didn't have them; we should have been retrying before but I missed a few txs that never caused issues until now
* explicitly check for a claim already being resolved

**Changes**
see summary


jcvernaleo (approved):
OK

max-sanchez (approved):

PR #752, opened 2025-11-13
Introduce Continuum Transfunctioner daemon

marcopeereboom:
**Summary**
Add a new daemon that handles Threshold Signature Scheme(TSS) magic for MinerFI.

This is being separated into several PRs to keep them smaller and more reviewed.

In order for this to move to production we must at least add:
* Spec for continuum protocol
* HIPPO for transfunctioner


cmd/transfunctionerd/transfunctionerd.go
```diff
@@ -0,0 +1,142 @@
...
+
+	"github.com/juju/loggo"
+
+	"github.com/hemilabs/heminetwork/v2/config"
+	"github.com/hemilabs/heminetwork/v2/service/continuum"
+	"github.com/hemilabs/heminetwork/v2/version"
+)
+
+const (
+	daemonName      = "trfd"
```
joshuasing: I believe it makes it easier if `daemonName` matches the package name. I think it would be best to either:
1. Rename this package to `trfd`; or
2. Change the value to `transfunctionerd` to match the package name

service/continuum/continuum.go
```diff
@@ -0,0 +1,395 @@
...
+type Config struct {
+	Home                    string
+	LogLevel                string
+	PprofListenAddress      string
+	PrometheusListenAddress string
+	PrometheusNamespace     string
+	Secret                  string
+	ListenAddress           string
+	MaxConnections          int
+}
```
joshuasing, line 55: nit: missing blank line after struct
```suggestion
}

```

service/continuum/continuum.go
```diff
@@ -0,0 +1,395 @@
...
+	"github.com/mitchellh/go-homedir"
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/hemilabs/heminetwork/v2/service/deucalion"
+	"github.com/hemilabs/heminetwork/v2/service/pprof"
+)
+
+const (
+	logLevel = "INFO"
+	appName  = "cnt"
```
joshuasing:
```suggestion
	appName  = "continuum"
```

service/continuum/continuum.go
```diff
@@ -0,0 +1,395 @@
+// Copyright (c) 2025 Hemi Labs, Inc.
+// Use of this source code is governed by the MIT License,
+// which can be found in the LICENSE file.
+
+package continuum
```
joshuasing, line 7: In the past we have always named the cmd and service package the same (with a `d` suffix for the cmd package). I find this naming confusing, wouldn't it be better for them to be the same?

joshuasing, line 7: nit: missing package doc comment

```
// Package continuum implements ...
package continuum
```

service/continuum/continuum.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+	return hex.EncodeToString(buf)
+}
+
+func (s *Server) newSession(t *Transport) string {
+	for {
+		id := genID()
+		s.mtx.Lock()
+		if _, ok := s.sessions[id]; ok {
+			// ID is already used, retry.
```
joshuasing, line 117: bug: `s.mtx` is not unlocked before `continue`, resulting in a deadlock.
```suggestion
			// ID is already used, retry.
			s.mtx.Unlock()
```

service/continuum/protocol.go
```diff
@@ -0,0 +1,954 @@
...
+	// Encode size to prefix nonce
+	var size [4]byte
+	binary.BigEndian.PutUint32(size[:], uint32(ts))
+	nonce := t.nonce.Next()
+	blob := secretbox.Seal(append(size[1:4], nonce[:]...), cleartext, nonce,
+		t.encryptionKey)
+
+	// diagnostic
+	if ts != len(blob)-3 {
+		panic(fmt.Sprintf("encryption diagnostic: wanted %v got %v",
```
joshuasing: nit: use `error` type for panic values
```suggestion
		panic(fmt.Errorf("encryption diagnostic: wanted %v got %v",
```

service/continuum/protocol.go
```diff
@@ -0,0 +1,954 @@
...
+	m := make(map[string]string)
+	for _, v := range s {
+		kv := strings.Split(v, "=")
+		if len(kv) != 2 {
+			return nil, ErrInvalidTXTRecord
+		}
+		m[kv[0]] = kv[1]
+	}
+	return m, nil
+}
```
joshuasing, line 1003: I think we should follow the format used by DKIM (https://www.rfc-editor.org/rfc/rfc8463.txt):
`"v=DKIM1; k=ed25519; p=11qYAYKxCrfVS/7TyWQHOg7hcvPapiMlrwIaaPcHURo="`

This appears to be missing semicolon separators, and `dnsAppName` does not contain a version number.
Also it might be worth checking the length of `txt` before parsing.

joshuasing (changes requested):

AL-CT (approved):

cmd/transfunctionerd/transfunctionerd.go
marcopeereboom: Fixed

service/continuum/continuum.go
marcopeereboom: Fixed.

service/continuum/continuum.go
marcopeereboom: Fixed.

service/continuum/continuum.go
marcopeereboom: Fixed

service/continuum/continuum.go
marcopeereboom: Dang, good find. Fixed.

service/continuum/protocol.go
marcopeereboom: Fixed.

service/continuum/protocol.go
marcopeereboom: This was a crapshoot tbh. Other record's do not use no `;`. I picked it because less code really but it does make sense to me. Fixed.

service/continuum/continuum.go
marcopeereboom: Well yes but in this case the protocol is called `continuum` and the daemon is `transfunctionerd`. I do expect that `continuum` will be picked up by external parties that will run their own version. Let me have me my joke :-)

cmd/transfunctionerd/transfunctionerd.go
marcopeereboom: This is fixed @joshuasing

cmd/transfunctionerd/transfunctionerd.go
```diff
@@ -0,0 +1,142 @@
...
+
+	cfg = continuum.NewDefaultConfig()
+	cm  = config.CfgMap{
+		"TRF_HOME": config.Config{
+			Value:        &cfg.Home,
+			DefaultValue: defaultHome,
+			Help:         "data directory",
+			Print:        config.PrintAll,
+		},
+		"TRF_SECRET": config.Config{
```
joshuasing: nit: I think it is worth making it clear this is a private key in the environment variable name.
```suggestion
		"TRF_PRIVATE_KEY": config.Config{
```

service/continuum/protocol.go
```diff
@@ -0,0 +1,1058 @@
...
+//	the signature. The derived ETP must match the provided identity to
+//	prove ownership of the accompanying private key. Additionally, the
+//	identity may be stores in a DNS TXT record that is associated with the
+//	remote host.
+//	TODO: define DNS mechanism in more detail.
+//
+// In the continuum protocol every host has a long lived identity. This
+// identity is the RIPEMD160 digest of a public secp256k1 compressed key. This
+// means that every host has a long lived secp256k1 private key that is used to
+// sign various things over its life-cycle and uniquely identify itself withing
```
joshuasing: typo: withing -> within (also caught by the linter)
```suggestion
// sign various things over its life-cycle and uniquely identify itself within
```

go.mod
```diff
@@ -1,22 +1,25 @@
...
 	github.com/btcsuite/btcwallet/wallet/txsizes v1.2.5
 	github.com/coder/websocket v1.8.14
 	github.com/davecgh/go-spew v1.1.1
+	github.com/decred/dcrd/crypto/ripemd160 v1.0.2
 	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.4.0
 	github.com/docker/go-connections v0.6.0
 	github.com/dustin/go-humanize v1.0.1
 	github.com/ethereum/go-ethereum v1.16.7
 	github.com/go-test/deep v1.1.1
+	github.com/juju/loggo v1.0.0
```
joshuasing: Need to update imports to use `github.com/juju/loggo/v2`.

go.mod
```diff
@@ -1,22 +1,25 @@
 module github.com/hemilabs/heminetwork/v2
 
-go 1.25
+go 1.25.0
```
joshuasing, line 3: Specifying the patch version isn't necessary here, best to leave as-is (might have been done by vim-go or similar?):
```suggestion
go 1.25
```

joshuasing (changes requested):

cmd/transfunctionerd/transfunctionerd.go
marcopeereboom: Fixed

service/continuum/protocol.go
marcopeereboom: Fixed

go.mod
marcopeereboom: Fixed everywhere.

go.mod
marcopeereboom: I have to stress again that this is done by the linter and/or gofumpt..

marcopeereboom: per our conversation, this change is made by `go mod tidy` and thus cannot be prevented.

joshuasing (approved):

PR #774, opened 2025-11-19
Test pulling in upstream op-stack

ClaytonNorthey92:
**Summary**

**Changes**

### DIFF
--- a/e2e/docker-compose.yml
+++ b/e2e/docker-compose.yml
@@ -199,7 +199,7 @@ services:
       - "--datadir=/data"
       - "--accept-terms-of-use"
       - "--verbosity=trace"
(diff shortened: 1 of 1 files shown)


.github/workflows/localnet-test.yml
```diff
@@ -22,7 +22,6 @@ jobs:
       matrix:
         go-version: [ "1.25.x" ]
         gethl1:
-          - "ethereum/client-go:v1.15.8@sha256:a5ef22282a8d154d9f634d38cdcc233e72daef454edc6339d13b3fc456355fa2"
```
ClaytonNorthey92, line 25: this does not support the `--override.osaka` flag, removing it

PR #782, opened 2025-11-24
hemlctl, tbcd: fix typos in comments

mk0walsk:
Update tbcd README: bump Go prerequisite to 1.25 (matches go.mod/toolchain v1.25.4) to stop misleading builders
Fix comment typos in hemictl/pop.go

### DIFF
--- a/hemi/pop/pop.go
+++ b/hemi/pop/pop.go
@@ -39,7 +39,7 @@ func MinerAddressFromString(address string) (*MinerAddress, error) {
 
 // TransactionL2 rename to Transaction and fixup this code
 type TransactionL2 struct {
-	L2Keystone *hemi.L2KeystoneAbrev // XXX why is this a pointer?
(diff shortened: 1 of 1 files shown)


hemi/pop/pop.go
```diff
@@ -39,7 +39,7 @@ func MinerAddressFromString(address string) (*MinerAddress, error) {
 
 // TransactionL2 rename to Transaction and fixup this code
 type TransactionL2 struct {
-	L2Keystone *hemi.L2KeystoneAbrev // XXX wtf is this a pointer?
+	L2Keystone *hemi.L2KeystoneAbrev // XXX why is this a pointer?
 }
```
marcopeereboom, line 43: I am meh on this. I use strong language to indicate something that shouldn't be ignored.

jcvernaleo: Agree with @marcopeereboom here.  There is no reason for this change for please revert it.

joshuasing: Reverted.

PR #789, opened 2025-12-08
ci, e2e: pull in latest op-stack to test; fix critical bugs

ClaytonNorthey92:
**Summary**
* pull in latest hemi branches in our op-stack repos
* pass docker build args to ALL build steps so we're building the correct commits
* the latest foundry release is broken, so use a commit near tip on the main/master branch
* use older optimism commit to build op-deployer so we don't have to modify initial deploy in localnet
* remove older geth version from localnet workflow

fixes https://github.com/hemilabs/heminetwork/issues/790

**Changes**
see summary

### DIFF
--- a/e2e/optimism-stack.Dockerfile
+++ b/e2e/optimism-stack.Dockerfile
@@ -2,10 +2,17 @@
 # Use of this source code is governed by the MIT License,
 # which can be found in the LICENSE file.
 
-ARG OP_GETH_COMMIT=faf53b75ee93d24c93885364dbeb62bb7fc876f3
-ARG OPTIMISM_COMMIT=12dba15e5e3fdc48620a81872b381b2e79fcb62b
+ARG OP_GETH_COMMIT=12a93e2f9538378a3cdf0cd3976409246e1c0af1
+ARG OPTIMISM_COMMIT=984ab34d8f3b69ed86bdcd8d053855daf4c71bd3
+
+# commit near tip on "master" (main) branch.  the most recent release is
+# broken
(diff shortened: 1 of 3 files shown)


e2e/optimism-stack.Dockerfile
```diff
@@ -36,6 +40,20 @@ RUN cargo install just
 
 WORKDIR /git
 
+RUN git clone https://github.com/foundry-rs/foundry.git
+
+WORKDIR /git/foundry
+
+# commit near tip on "master" (main) branch.  the most recent release is
+# broken
+RUN git checkout 97fba0a51e335a174442b19d92b64df9d2ab72ab
```
joshuasing: Could we move this to an `ARG`, similar to `OP_GETH_COMMIT`?

e2e/optimism-stack.Dockerfile
```diff
@@ -36,6 +40,20 @@ RUN cargo install just
...
+RUN cargo build --release
+
+ENV PATH="${PATH}:/git/foundry/target/release"
```
joshuasing: Is it likely for unexpected binaries to end up in this directory? I think it could be better to copy the wanted binary from here into `/usr/bin/` or similar

ClaytonNorthey92: I would say it's unlikely, but doesn't hurt to copy the binary like you had suggested, so I can do that 👍

e2e/optimism-stack.Dockerfile
ClaytonNorthey92: yeah I can do that

PR #811, opened 2026-01-05
all: move various repeated test functions to testutil

AL-CT:
**Summary**
Several helper test functions that were either repeated throughout the code or that could be useful in future tests have been moved to the `testutil` package.

**Changes**
Moved the following functions:
- `MessageListener`
- `RandomBytes`
- `RandomHash`
- `String2Hash`
- `Bytes2Hash`
- `DecodeHex`
- `ErrorIsOneOf`

Also removed the use of the FreePort library from the FreePort function, and added some documentation to each function.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -117,8 +117,6 @@ github.com/docker/go-units v0.5.0 h1:69rxXcBk27SvSaaxTtLh/8llcHD8vYHT7WSdRZ/jvr4
 github.com/docker/go-units v0.5.0/go.mod h1:fgPhTUdO+D/Jk86RDLlptpiXQzgHJF7gydDDbaIK4Dk=
 github.com/dustin/go-humanize v1.0.1 h1:GzkhY7T5VNhEkwH0PVJgjz+fX1rhBrR7pRT3mDkpeCY=
 github.com/dustin/go-humanize v1.0.1/go.mod h1:Mu1zIs6XwVuF/gI1OepvI0qD18qycQx+mFykh5fBlto=
-github.com/ebitengine/purego v0.9.0 h1:mh0zpKBIXDceC63hpvPuGLiJ8ZAa3DfrFTudmfi8A4k=
-github.com/ebitengine/purego v0.9.0/go.mod h1:iIjxzd6CiRiOG0UyXP+V1+jWqUXVjPKLAI0mRfJZTmQ=
 github.com/ebitengine/purego v0.9.1 h1:a/k2f2HQU3Pi399RPW1MOaZyhKJL9w/xFpKAg4q1s0A=
 github.com/ebitengine/purego v0.9.1/go.mod h1:iIjxzd6CiRiOG0UyXP+V1+jWqUXVjPKLAI0mRfJZTmQ=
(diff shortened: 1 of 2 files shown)


database/tbcd/level/level_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2025 Hemi Labs, Inc.
+// Copyright (c) 2026 Hemi Labs, Inc.
```
joshuasing: nit: keep existing year for consistency
```suggestion
// Copyright (c) 2025-2026 Hemi Labs, Inc.
```

internal/testutil/testutil.go
```diff
@@ -85,11 +117,69 @@ func FillBytesZero(prefix string, n int) []byte {
 }
 
 // FreePort finds a port that is currently free.
-// TODO: remove use of freeport library.
-func FreePort() string {
-	port, err := freeport.GetFreePort()
+func FreePort(ctx context.Context) string {
+	lc := net.ListenConfig{}
+	l, err := lc.Listen(ctx, "tcp", ":0")
```
joshuasing: Much nicer! :D - At some point, I wonder if we can just change the daemons to use `localhost:0` directly so that there isn't a chance of something else taking the port between here and the daemon starting. I think it would just require having a function to return the listener address after the daemon starts.

One nit - probably best to listen on `localhost` to avoid exposing a port on all interfaces here:

```suggestion
	l, err := lc.Listen(ctx, "tcp", "localhost:0")
```

service/tbc/mempool_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2025 Hemi Labs, Inc.
+// Copyright (c) 2026 Hemi Labs, Inc.
```
joshuasing: nit: keep existing year for consistency (is the licenser tool changing 2025->2026 automatically?)
```suggestion
// Copyright (c) 2025-2026 Hemi Labs, Inc.
```

service/tbc/mempoolfees.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2025 Hemi Labs, Inc.
+// Copyright (c) 2026 Hemi Labs, Inc.
```
joshuasing: nit: keep existing year for consistency
```suggestion
// Copyright (c) 2025-2026 Hemi Labs, Inc.
```

internal/testutil/testutil.go
AL-CT: Done 👍
And yeah, probably should change to just use :0 everywhere, will put that on the list

database/tbcd/level/level_test.go
AL-CT: Fixed 👍

service/tbc/mempool_test.go
AL-CT: Fixed 👍
And yes, it seemed to be doing that for some reason?

service/tbc/mempoolfees.go
AL-CT: Fixed 👍

service/tbc/mempool_test.go
joshuasing: Weird... I will look into it, thanks. I think it might need a re-write lol...

jcvernaleo (approved):
OK

Issue #750, opened 2025-11-12
enhancement: generate test credentials


ClaytonNorthey92:
for some of our tests (for example, the `./e2e` tests) we have credentials hard-coded.  This is not an issue in itself, but leads to confusion and probably isn't a great practice.

track down all of the areas where test credentials (such as private keys) are hard-coded and update them to be generated instead

PR #818, opened 2026-01-07
 e2e: use different private keys to run sequencing/non-sequencing localnet tests in parallel

ClaytonNorthey92:
**Summary**
* use different private keys/accounts for sequencing and non-sequencing tests in TestL1L2Comms
* handle for errors that may now occur if the other test gets to the step first (i.e. claim already resolved, game not in progress)

**Changes**
see summary


e2e/monitor/main_test.go
```diff
@@ -99,6 +100,32 @@ func privateKeyForTest() string {
 	return "dfe61681b31b12b04f239bc0692965c61ffc79244ed9736ffa1a72d00a23a530"
 }
 
+// privateKeyForTestByIndex returns different private keys for concurrent test execution
+func privateKeyForTestByIndex(t *testing.T, index uint) string {
+	if testingFork() {
+		// For forked tests, use different keys for different indices
+		keys := []string{
+			"92db14e403b83dfe3df233f83dfa3a0d7096f21ca9b0d6d6b8d88b2b4ec1564e",
```
joshuasing, line 124: Could we move these (and any other private keys in this file) to constants, and clearly document they are _strictly for testing purposes only, and are absolutely not to be used in production ever_?- Just because otherwise we will eventually get VDP reports about them.

e.g.
```
// These are intentionally hard-coded keys used exclusively for automated
// and local testing.
//
// These values are intentionally public, and are not a security risk.
// Production use will result in immediate and irreversible loss of funds.
//
// Note to security researchers: The presence of these values is intentional
// and do not pose a security risk. Please do not report them.
const (
  testingForkKey1 = "a"
  testingForkKey2 = "b"
)
```
or similar

ClaytonNorthey92: hey @joshuasing , we could, I was saving that for this issue https://github.com/hemilabs/heminetwork/issues/750, I could do that issue in this PR if you'd like? or a separate one. up to you

Issue #819, opened 2026-01-07
feature: cache localnet builds that don't change


ClaytonNorthey92:
### Confirmation

- [x] I have checked for similar issues.

### Problem

often, when we re-run localnet tests the build of the network itself does not change and it takes up roughly 25 minutes.

### Suggested solution

cache the image built from `optimism-stack.Dockerfile` so that it persists between builds if nothing changes

PR #820, opened 2026-01-07
e2e: cache localnet builds between runs for same Dockerfile hash

ClaytonNorthey92:
**Summary**
* on each run:
  * compute hash of optimism-stack.Dockerfile
  * if this exists in the cache, use that
  * otherwise, build and save it to the cache with the file's hash so it can be retrieved later but ALSO so if the file contents change it will be re-built
* use default eviction of 7 days if I am not mistaken
* move the building of cargo into a different build step, only include the binary in the final image to save on space
* update readme

fixes https://github.com/hemilabs/heminetwork/issues/819

**Changes**
see summary

### DIFF
--- a/e2e/optimism-stack.Dockerfile
+++ b/e2e/optimism-stack.Dockerfile
@@ -2,7 +2,7 @@
 # Use of this source code is governed by the MIT License,
 # which can be found in the LICENSE file.
 
-# increment me to break the cache: 1
+# increment me to break the cache: 2
 
 ARG OP_GETH_COMMIT=12a93e2f9538378a3cdf0cd3976409246e1c0af1
 ARG OPTIMISM_COMMIT=984ab34d8f3b69ed86bdcd8d053855daf4c71bd3


.github/workflows/localnet-test.yml
```diff
@@ -5,7 +5,12 @@
 # GitHub Actions workflow to run e2e network tests.
 name: "Localnet Test"
 on:
-  workflow_dispatch: # Manually triggered
+  push:
```
ClaytonNorthey92: change this back

.github/workflows/localnet-test.yml
```diff
@@ -30,12 +30,62 @@ jobs:
...
 
+      - name: "Build optimism-stack image"
+        if: steps.cache-optimism-stack.outputs.cache-hit != 'true'
+        run: |
+          echo "Building optimism-stack image from Dockerfile (no cache or Dockerfile changed)"
+          docker build -f ./e2e/optimism-stack.Dockerfile -t optimism-stack:latest .
+
+      - name: "build localnet"
+        run: |
+          if [ "${{ steps.cache-optimism-stack.outputs.cache-hit }}" = "true" ]; then
```
joshuasing: `${{ }}` expression is evaluated before the script is executed, which can allow script injection attacks if the expression can be influenced by user-provided input.

In this case, this can't be exploited, but it is a best practice to use `env` to pass any templated strings in, just in case.

```suggestion
        env:
          OPTIMISM_STACK_CACHE_HIT: "${{ steps.cache-optimism-stack.outputs.cache-hit }}"
        run: |
          if [ "$OPTIMISM_STACK_CACHE_HIT" = "true" ]; then
```

.github/workflows/localnet-test.yml
```diff
@@ -30,12 +30,62 @@ jobs:
...
+          IMAGE_ID=$(docker images -q optimism-stack:latest)
+          if [ -z "$IMAGE_ID" ]; then
+            echo "Error: Could not find optimism-stack:latest image"
+            exit 1
+          fi
+          echo "Image ID: $IMAGE_ID"
+          IMAGE_SIZE=$(docker images --format "{{.Size}}" "$IMAGE_ID")
+          echo "Image size: $IMAGE_SIZE"
+          docker save -o optimism-stack-image.tar optimism-stack:latest
+          TAR_SIZE=$(ls -lh optimism-stack-image.tar | awk '{print $5}')
```
joshuasing: I think this could be simplified to use `du -hs`, which will output in the format `4.0K`:
```suggestion
          TAR_SIZE=$(du -hs optimism-stack-image.tar | cut -f1)
```

e2e/optimism-stack.Dockerfile
```diff
@@ -22,6 +23,16 @@ WORKDIR /git/foundry
 RUN git checkout $FOUNDRY_COMMIT
 RUN cargo build --package forge
 
+FROM golang:1.25.5-trixie@sha256:ef151f0384896831258e71065176f1e63f5a90bcbe6a98ec679a1990011a2655 AS just_build
+
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
+ENV PATH="${PATH}:/root/.cargo/bin"
+
+WORKDIR /git
+RUN git clone https://github.com/casey/just
```
joshuasing, line 33: Is this using the latest commit on the default branch? If so, I think it would be a good idea to pin the commit to ensure reproducibility.

.github/workflows/localnet-test.yml
joshuasing, line 1: There are a couple of names used in the scripts (e.g. `optimism-stack:latest`, `optimism-stack-image.tar`) that I think could be worth putting into environment variables (probably job-level) just to make sure they're consistent - then they can be used with `${{ env.NAME }}` or `$NAME` (as usual) in scripts

.github/workflows/localnet-test.yml
ClaytonNorthey92: sure, will update 👍

.github/workflows/localnet-test.yml
ClaytonNorthey92: I'll give this a try 👍

e2e/optimism-stack.Dockerfile
ClaytonNorthey92: good call, I'll update

.github/workflows/localnet-test.yml
ClaytonNorthey92: sure, I'll make this change 👍

.github/workflows/localnet-test.yml
```diff
@@ -30,12 +33,70 @@ jobs:
...
+          echo "Saving $DOCKER_IMAGE_TAG image to cache"
+          IMAGE_ID=$(docker images -q $DOCKER_IMAGE_TAG)
+          if [ -z "$IMAGE_ID" ]; then
+            echo "Error: Could not find $DOCKER_IMAGE_TAG image"
+            exit 1
+          fi
+          echo "Image ID: $IMAGE_ID"
+          IMAGE_SIZE=$(docker images --format "{{.Size}}" "$IMAGE_ID")
+          echo "Image size: $IMAGE_SIZE"
+          docker save -o optimism-stack-image.tar $DOCKER_IMAGE_TAG
```
AL-CT: How big is the tarball? The [docker documentation](https://docs.docker.com/reference/cli/docker/image/save/#save-an-image-to-a-targz-file-using-gzip) suggests it might be worthwhile to compress it with gzip using: `docker save myimage:latest | gzip > myimage_latest.tar.gz`, and it seems like you can load tar.gz files with docker without adding a manual step for decompressing.

This might add overhead, or make it faster due to upload / download, not sure. Just thought I'd mention it in case we do this for other images and need to limit how much cache space we use 👍

AL-CT (approved):
Looks good, just added a suggestion for you to consider, and agreed with the pinned commit comment that Joshua suggested 👍

.github/workflows/localnet-test.yml
joshuasing: I think `optimism-stack-image.tar.gz` could probably be moved to an env variable as well, up to you.

.github/workflows/localnet-test.yml
ClaytonNorthey92: Ah great idea, I'll make that update 👍

ClaytonNorthey92: Ah great idea, I'll make that update 👍

.github/workflows/localnet-test.yml
ClaytonNorthey92: sure, I'll update that 👍

PR #865, opened 2026-02-03
synctest: add sync testing tool for op-geth nodes

ClaytonNorthey92:
**Summary**

* added a program/module called synctest
* see README.md for full functionality and how to use

to summarize, syntest will

* start, accepting a "control" and "experimental" op-geth endpoint
* wait for experimental tbc to sync via SyncInfo
* wait for experimental l2 tip to match control l2 tip (i.e. wait for the experimental l2 node to sync)

the "control" node is a healthy node that you're testing _against_ (ex. hemi public rpc).  the "experimental" node is a node that you're testing whether it will sync or not.

**Changes**
see summary

### DIFF
--- a/synctest/synctest.go
+++ b/synctest/synctest.go
@@ -42,37 +42,37 @@ const (
 var (
 	tbcFetchHealthRetryCount = 10
 	loopDelay                = 5 * time.Second
-	log                      loggo.Logger
 	validNetworks            = []string{networkMainnet, networkTestnet, networkLocalnet}
 	validSyncmodes           = []string{syncmodeSnap, syncmodeFull}
 	errInvalidNetwork        = errors.New("invalid network")
 	errInvalidSyncmode       = errors.New("invalid syncmode")
+	log                      = loggo.GetLogger("synctest")
 )
 
 type config struct {
-	ControlOpGethEndpoint               string `env:"CONTROL_OP_GETH_ENDPOINT"`
-	ExperimentalOpGethEndpoint          string `env:"EXPERIMENTAL_OP_GETH_ENDPOINT"`
-	ExperimentalOpGethTbcHealthEndpoint string `env:"EXPERIMENTAL_OP_GETH_TBC_HEALTH_ENDPOINT"`
-	SlackOauthToken                     string `env:"SLACK_OAUTH_TOKEN"`
-	SlackChannel                        string `env:"SLACK_CHANNEL"`
-	SlackURL                            string `env:"SLACK_URL"`
-	Network                             string `env:"NETWORK"`
-	Syncmode                            string `env:"SYNCMODE"`
-	NotifyBySeconds                     uint   `env:"NOTIFY_BY_SECONDS"`
-	SkipDockerLogs                      bool   `env:"SKIP_DOCKER_LOGS"`
-	LogLevel                            string `env:"LOG_LEVEL"`
+	ControlOpGethEndpoint               string        `env:"CONTROL_OP_GETH_ENDPOINT"`
+	ExperimentalOpGethEndpoint          string        `env:"EXPERIMENTAL_OP_GETH_ENDPOINT"`
+	ExperimentalOpGethTbcHealthEndpoint string        `env:"EXPERIMENTAL_OP_GETH_TBC_HEALTH_ENDPOINT"`
+	SlackOauthToken                     string        `env:"SLACK_OAUTH_TOKEN"`
+	SlackChannel                        string        `env:"SLACK_CHANNEL"`
+	SlackURL                            string        `env:"SLACK_URL"`
+	Network                             string        `env:"NETWORK"`
+	Syncmode                            string        `env:"SYNCMODE"`
+	NotifyBy                            time.Duration `env:"NOTIFY_BY"`
+	SkipDockerLogs                      bool          `env:"SKIP_DOCKER_LOGS"`
+	LogLevel                            string        `env:"LOG_LEVEL"`
 }
 
 func DefaultConfig() *config {
 	return &config{
-		NotifyBySeconds: 60 * 30,
+		NotifyBy: 30 * time.Minute,
 	}
 }
 
 func configFromEnv() (*config, error) {
 	c := DefaultConfig()
 	envOpts := env.Options{Prefix: "SYNCTESTER_"}
-	if err := env.ParseWithOptions(&c, envOpts); err != nil {
+	if err := env.ParseWithOptions(c, envOpts); err != nil {
 		return nil, fmt.Errorf("could not parse env: %w", err)
 	}
 
@@ -103,7 +103,9 @@ func waitForSync(ctx context.Context) error {
 		return err
 	}
 
-	log = loggo.GetLogger("synctest")
+	if err := loggo.ConfigureLoggers(c.LogLevel); err != nil {
+		return err
+	}
 
 	logLevel, _ := loggo.ParseLevel(c.LogLevel)
 	if logLevel == loggo.UNSPECIFIED {
@@ -112,16 +114,14 @@ func waitForSync(ctx context.Context) error {
 
 	log.Infof("logLevel = %d, logLevel from env = %s", logLevel, c.LogLevel)
 
-	log.SetLogLevel(logLevel)
-
 	log.Infof("starting sync testing with configured endpoints: control (%s), experimental(%s)", c.ControlOpGethEndpoint, c.ExperimentalOpGethEndpoint)
 
-	lastNotifiedAt := time.Now().Add(-(time.Duration(c.NotifyBySeconds) * time.Second))
+	lastNotifiedAt := time.Now().Add(-c.NotifyBy)
 	lastSyncInfo := tbc.SyncInfo{}
 	lastl2BlockNumber := uint64(0)
 
 	for {
-		shouldNotify := time.Since(lastNotifiedAt) >= (time.Duration(c.NotifyBySeconds) * time.Second)
+		shouldNotify := time.Since(lastNotifiedAt) >= c.NotifyBy
 		if err := reportProgress(ctx, c, &lastSyncInfo, shouldNotify); err != nil {
 			return err
 		}
@@ -151,7 +151,12 @@ func waitForSync(ctx context.Context) error {
 func reportProgress(ctx context.Context, c *config, lastSyncInfo *tbc.SyncInfo, notify bool) error {
 	var syncInfo tbc.SyncInfo
 	for i := range tbcFetchHealthRetryCount {
-		resp, err := http.Get(fmt.Sprintf("%s", c.ExperimentalOpGethTbcHealthEndpoint))
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, c.ExperimentalOpGethTbcHealthEndpoint, nil)
+		if err != nil {
+			return fmt.Errorf("error creating new request: %w", err)
+		}
+
+		resp, err := http.DefaultClient.Do(req)
 		if err != nil {
 			log.Warningf("error getting tbc health: %s", err)
 			return nil
@@ -159,8 +164,10 @@ func reportProgress(ctx context.Context, c *config, lastSyncInfo *tbc.SyncInfo, 
 
 		defer resp.Body.Close()
 
-		if err := json.NewDecoder(b).Decode(&syncInfo); err != nil {
+		if err := json.NewDecoder(resp.Body).Decode(&syncInfo); err != nil {
 			return fmt.Errorf("decode tbc json: %w", err)
+		} else {
+			break
 		}
 
 		if i >= tbcFetchHealthRetryCount-1 {


synctest/synctest_test.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	t.Logf("tbc progress json: %v", string(b))
+
+	var syncInfo tbc.SyncInfo
+	if err := json.Unmarshal(b, &syncInfo); err != nil {
+		t.Fatal(err)
+	}
+
+	t.Logf(`
```
ClaytonNorthey92: check for progressing (i.e. this height is > than last height)

synctest/synctest_test.go
```diff
@@ -0,0 +1,134 @@
...
+	experimentalBlock, err := l2Experiment.BlockByNumber(ctx, big.NewInt(int64(rpc.LatestBlockNumber)))
+	if err != nil {
+		logWithHooks(t, "error getting latest experiment block: %s", err)
+		return false
+	}
+
+	logWithHooks(t,
+		"comparing hashes for latest blocks (control) %s ?= (experiment) %s", controlBlock.Hash().Hex(), experimentalBlock.Hash().Hex())
+
+	return controlBlock.Hash().Hex() == experimentalBlock.Hash().Hex()
```
ClaytonNorthey92: once we're synced, wait for a few new blocks and make sure that those stay synced as well

ClaytonNorthey92: check that each block, keystone matches from tip to genesis (reference larry function)

ClaytonNorthey92: tip - 5?

ClaytonNorthey92: state root? create state root from genesis, see if it matches control endpoint?

find another way to check keystones

.github/workflows/localnet-test.yml
```diff
@@ -109,6 +109,16 @@ jobs:
       - name: "remove more dangling"
         run: docker system prune -f
 
+
+      - name: "test sync"
+        working-directory: ./synctest
+        env: 
+          SYNCTESTER_CONTROL_OP_GETH_ENDPOINT: http://localhost:8546 # sequencing
+          SYNCTESTER_EXPERIMENTAL_OP_GETH_ENDPOINT: http://localhost:18546 # non-sequencing
+          SYNCTESTER_EXPERIMENTAL_OP_GETH_TBC_ENDPOINT: http://localhost:5555/health
```
ClaytonNorthey92: this really isn't a tbc endpoint; it's a tbc prometheus endpoint, adjust this variable name accordingly

synctest/synctest/synctest.go
```diff
@@ -0,0 +1,259 @@
...
+	experimentalOpGethEndpoint := experimentalOpGethEndpointFromEnv()
+	log.Printf("starting sync testing with configured endpoints: control (%s), experimental(%s)", controlOpGethEndpoint, experimentalOpGethEndpoint)
+
+	lastReportedAt := time.Now().Add(-shouldProgressBy)
+	lastNotifiedAt := time.Now().Add(-notifyBy)
+	lastLowestBtcBlockNumber := uint64(0)
+	lastHighestBtcBlockNumber := uint64(0)
+
+	for {
+		if time.Now().Sub(lastReportedAt) >= shouldProgressBy {
```
AL-CT: nit: should probably use `time.Since()` instead of `time.Now().Sub()`

synctest/synctest/synctest.go
```diff
@@ -0,0 +1,259 @@
...
+		utxo index %s:%d
+		keystone index %s:%d
+		blockheader index %s:%d`,
+		syncInfo.Tx.Hash, syncInfo.Tx.Height,
+		syncInfo.Utxo.Hash, syncInfo.Utxo.Height,
+		syncInfo.Keystone.Hash, syncInfo.Keystone.Height,
+		syncInfo.BlockHeader.Hash, syncInfo.BlockHeader.Height,
+	)
+
+	if lowest > *lastLowestBtcBlockNumber || highest > *lastHighestBtcBlockNumber {
```
AL-CT: I may be missing a condition, but I think this check might fail when it shouldn't. For example, with highest BTC block being `1000`:

If this is called when
```
Blockheader tip: 1000
TX indexer tip: 10
UTXO indexer tip: 0
KSS indexer tip: 0

LOWEST = 0
HIGHEST = 1000
```

Then the next time we call we have
```
Blockheader tip: 1000
TX indexer tip: 800
UTXO indexer tip: 0
KSS indexer tip: 0

LOWEST = 0
HIGHEST = 1000
```

Then this will fail, even though one indexer is progressing. Also we can't just add a `>=` instead of `>` since then if it stalls, it won't fail. The solution might be to store a map with the values for each indexer, and comparing them above 🤔

synctest/synctest/synctest.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	experimentalBlock, err := l2Experiment.HeaderByNumber(ctx, big.NewInt(int64(rpc.LatestBlockNumber)))
+	if err != nil {
+		log.Printf("error getting latest experiment block: %s", err)
+		return false
+	}
+
+	// if the experimental block is at 0, then it has not finished the tbc sync
+	if experimentalBlock.Number.Uint64() == uint64(0) {
```
AL-CT: I think it might be possible for tbc to be synced but for some reason this just doesn't progress past block 0 🤔

synctest/synctest/synctest.go
```diff
@@ -0,0 +1,259 @@
...
+			if err := reportProgress(&lastLowestBtcBlockNumber, &lastHighestBtcBlockNumber, shouldNotify); err != nil {
+				return err
+			}
+			lastReportedAt = time.Now()
+			if shouldNotify {
+				lastNotifiedAt = time.Now()
+			}
+		}
+
+		lastl2BlockNumber := uint64(0)
```
AL-CT: Think this has to be set outside of the loop so it's not always 0

synctest/synctest/synctest.go
AL-CT: We can probably use the `Synced` field of tbc `SyncInfo` to check when the syncing is done + a little buffer time just so it doesn't fail if we get here right after that variable flips to true. Might even be worth it to only start checking if the L2 tips match once that is synced, but not sure

synctest/synctest/synctest.go
ClaytonNorthey92: Yup will make this change 👍

ClaytonNorthey92: Yup will make this change 👍

synctest/synctest/synctest.go
ClaytonNorthey92: Whilst you're correct, this error is temporary (it won't fail the program), it will just return `false` i.e. "I'm not synced yet".

However, you're reminding me of another issue that exists: upon the first hvm genesis block, we will likely be stuck on the same l2 block for a while.

So we probably need another check here. It will probably need to be aware of thr hvm genesis block. Let me think about this one

synctest/synctest/synctest.go
ClaytonNorthey92: Youre right about the "lowest block", that would fail when It shouldn't, ill make an update to make it smarter.

However if tbc stalls, I _want_ this to fail. Tbc should not stall for more than 20 minutes (2x the mainnet block time). If that's happening then something is wrong imo.

HOWEVER, there may be a case where tbc "unwinds" and goes backwards. Maybe i should just check if the indexers _differ_ from the last check? I.e. did they move in any direction? I think a `deep.Equal` would do it.

What do you think?

AL-CT (approved):

.github/workflows/go.yml
```diff
@@ -87,6 +87,29 @@ jobs:
       - name: "make build"
         run: make GOCACHE="$(go env GOCACHE)" build
 
+
+  # run the testing of the synctest package in parallel with other jobs
```
ClaytonNorthey92: remove this comment

synctest/synctest/synctest.go
```diff
@@ -0,0 +1,524 @@
...
+	return nil
+}
+
+func reportProgress(ctx context.Context, lastSyncInfo *tbc.SyncInfo, notify bool) error {
+	var syncInfo tbc.SyncInfo
+	for i := range tbcFetchHealthRetryCount {
+		resp, err := http.Get(fmt.Sprintf("%s", experimentalOpGethTbcHealthEndpointFromEnv()))
+		if err != nil {
+			log.Warningf("error getting tbc health: %s", err)
+			continue
```
AL-CT: I think we don't want a continue here since it might break the loop logic by not hitting the select / exit condition

synctest/main.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"context"
+	"log"
+	"time"
+
+	"synctest"
+)
+
+func main() {
+	ctx, cancel := context.WithTimeout(context.Background(), 24*time.Hour*14)
```
AL-CT: Just reminder to remove this 👍

synctest/README.md
```diff
@@ -0,0 +1,62 @@
...
+# it's for notification clarity
+SYNCTESTER_NETWORK
+
+# the syncmode you're testing (ex. snap), you may set this to any value,
+# it's for notification clarity
+SYNCTESTER_SYNCMODE
+
+
+# (optional, default 30 minutes) how many seconds between each notification to
+# that there should be
```
AL-CT: nit: typo
```suggestion
# (optional, default 30 minutes) how many seconds there should be between each
# notification
```

synctest/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+Syntest is a package that allows a user to test when a node is synced (or not).
+Synctest considers two nodes: an *experimental* node and a *control* node.
+
+
+The experimental node is the one that you are syncing.  The control node is 
+the node you're comparing the experimental node _against_. The control node
+should be a healthy, synced node.
+
+you may run syntest like so:
```
AL-CT: nit: typo
```suggestion
you may run synctest like so:
```

synctest/README.md
```diff
@@ -0,0 +1,62 @@
+# synctest
+
+Syntest is a package that allows a user to test when a node is synced (or not).
```
AL-CT: nit: typo (to prevent AI PRs fixing them in a week 😃)
```suggestion
Synctest is a package that allows a user to test when a node is synced (or not).
```

synctest/synctest/synctest.go
ClaytonNorthey92: oh yeah good call, I'll update

synctest/main.go
ClaytonNorthey92: yes thank you 🤦

synctest/README.md
```diff
@@ -0,0 +1,62 @@
...
+```shell
+go run ./...
+```
+
+The following environment variables are used:
+
+```
+# the endpoint for the healthy op-geth you are testing against, 
+# ex. hemi public rpc 
+SYNCTESTER_CONTROL_OP_GETH_ENDPOINT"
```
jcvernaleo: Why is there a single " at the end of that line?

synctest/README.md
```diff
@@ -0,0 +1,62 @@
...
+# (optional) the slack oauth token used to send notifications
+SYNCTESTER_SLACK_OAUTH_TOKEN
+
+# (optional) the slack channel id that you want to send notifications to
+SYNCTESTER_SLACK_CHANNEL
+
+# (optional) the slack url used for testing (i.e. mocking slack)
+SYNCTESTER_SLACK_URL
+
+# the network you're testing (ex. testnet), you may set this to any value,
```
jcvernaleo: I know it is a minor point but I'd kind of prefer this only allow reasonable values for the network.

synctest/README.md
```diff
@@ -0,0 +1,62 @@
...
+# it's for notification clarity
+SYNCTESTER_NETWORK
+
+# the syncmode you're testing (ex. snap), you may set this to any value,
```
jcvernaleo: Much like the network, could we restrict this to correct values?

synctest/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+# (optional) the slack channel id that you want to send notifications to
+SYNCTESTER_SLACK_CHANNEL
+
+# (optional) the slack url used for testing (i.e. mocking slack)
+SYNCTESTER_SLACK_URL
+
+# the network you're testing (ex. testnet), you may set this to any value,
+# it's for notification clarity
+SYNCTESTER_NETWORK
```
jcvernaleo, line 42: The code looks like this one is optional despite not being listed as such here.

synctest/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+# the syncmode you're testing (ex. snap), you may set this to any value,
+# it's for notification clarity
+SYNCTESTER_SYNCMODE
```
jcvernaleo, line 46: Same, looks to be optional.

synctest/go.mod
joshuasing, line 1: Is there a particular reason to use two separate Go modules for this?

Instead, I think a single Go module could be used with this layout:
```
synctest/
  go.mod
  go.sum
  main.go
  synctest.go
  synctest_test.go
```

If the `synctest{,_test}.go` files need to be in a separate package, it would be worth considering using the `internal` package to prevent them from being used outside of the module:
```
synctest/
  internal/
    synctest/
      synctest.go
      synctest_test.go
  go.mod
  go.sum
  main.go
```

or, if the `synctest.go` API is intended for internal use, I think it would be common to leave it in the same module (e.g. delete `synctest/synctest/go.{mod,sum}`.

synctest/go.mod
```diff
@@ -0,0 +1,89 @@
+module synctestcli
```
joshuasing:
```suggestion
module github.com/hemilabs/heminetwork/synctest
```

synctest/Dockerfile
joshuasing, line 1: If possible, could we follow the same pattern as `docker/*/Dockerfile` for this Dockerfile?

It is more secure and results in a smaller image when `golang` is used as a builder stage, before then copying the resulting binary into a minimal image.

synctest/README.md
ClaytonNorthey92: great question; I am not sure, I'll remove

synctest/README.md
ClaytonNorthey92: we can do some validation in the program. I'll add it 👍 I'll add "mainnet", "testnet", and "localnet"

synctest/README.md
ClaytonNorthey92: yup can do 👍

synctest/README.md
ClaytonNorthey92: you're right.  I'll add the validation without a default fallback so then it will be required 👍

synctest/README.md
ClaytonNorthey92: yup, should be required.  I'll update similar to the other one (and validate the values)

synctest/go.mod
ClaytonNorthey92: I am forgetting why exactly, probably just fallout from "hacking" to get this to work.  I can update, I'll try the first option that you suggested first 👍

synctest/Dockerfile
ClaytonNorthey92: sure, I'll use a multistage build in the `Dockerfile` 👍

.github/workflows/localnet-test.yml
```diff
@@ -109,6 +109,31 @@ jobs:
...
+            -e SYNCTESTER_CONTROL_OP_GETH_ENDPOINT=$SYNCTESTER_CONTROL_OP_GETH_ENDPOINT \
+            -e SYNCTESTER_EXPERIMENTAL_OP_GETH_ENDPOINT=$SYNCTESTER_EXPERIMENTAL_OP_GETH_ENDPOINT \
+            -e SYNCTESTER_EXPERIMENTAL_OP_GETH_TBC_HEALTH_ENDPOINT=$SYNCTESTER_EXPERIMENTAL_OP_GETH_TBC_HEALTH_ENDPOINT \
+            -e SYNCTESTER_NETWORK=$SYNCTESTER_NETWORK \
+            -e SYNCTESTER_SYNCMODE=$SYNCTESTER_SYNCMODE \
+            -e SYNCTESTER_SKIP_DOCKER_LOGS=true \
+            --network host \
+            --entrypoint /bin/synctest \
+            synctest:latest
+
```
joshuasing, line 134: nit: extra empty line
```suggestion
```

synctest/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2026 Hemi Labs, Inc.
+# Use of this source code is governed by the MIT License,
+# which can be found in the LICENSE file.
+FROM golang:1.25.7-alpine3.23@sha256:f6751d823c26342f9506c03797d2527668d095b0a15f1862cddb4d927a7a4ced AS builder
```
joshuasing, line 5: For consistency with other files, could we add an empty line after the copyright header?
```suggestion
# Copyright (c) 2026 Hemi Labs, Inc.
# Use of this source code is governed by the MIT License,
# which can be found in the LICENSE file.

FROM golang:1.25.7-alpine3.23@sha256:f6751d823c26342f9506c03797d2527668d095b0a15f1862cddb4d927a7a4ced AS builder
```

synctest/main.go
```diff
@@ -0,0 +1,16 @@
...
+
+package main
+
+import (
+	"context"
+	"fmt"
+)
+
+func main() {
+	if err := waitForSync(context.Background()); err != nil {
```
joshuasing: Instead of using `context.Background()` here, I think we could add signal handling for cancelation, for example (suggestion below requires import `os/signal` and `syscall`):
```suggestion
func main() {
	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
	defer stop()

	go func() {
		// Stop receiving signals as soon as possible.
		<-ctx.Done()
		stop()
	}()

	if err := waitForSync(ctx); err != nil {
```

synctest/main.go
```diff
@@ -0,0 +1,16 @@
...
+		panic(fmt.Sprintf("error syncing: %s", err))
```
joshuasing: nit: pass `error` type as parameter to `panic` - also allows wrapping `err`.
```suggestion
		panic(fmt.Errorf("error syncing: %w", err))
```

synctest/README.md
```diff
@@ -0,0 +1,62 @@
+# synctest
+
+Synctest is a package that allows a user to test when a node is synced (or not).
+Synctest considers two nodes: an *experimental* node and a *control* node.
```
joshuasing, line 4: This terminology confuses me a bit. Maybe "stable" node makes more sense than a "control" node, since the "control" node isn't actually controlling anything in this process? 🤔

synctest/synctest.go
```diff
@@ -0,0 +1,560 @@
...
+	if logLevel == loggo.UNSPECIFIED {
+		logLevel = loggo.INFO
+	}
+
+	log.Infof("logLevel = %d, logLevel from env = %s", logLevel, os.Getenv(logLevelEnv))
+
+	log.SetLogLevel(logLevel)
+}
+
+func waitForSync(ctx context.Context) error {
```
joshuasing, line 100: Instead of reading the environment variables as they're needed (requiring a lot of helper functions), maybe consider adding a `Config` struct and passing it in here, then a `readConfigFromEnv` function or similar can create the config and read all of the variables at once.

I think that would also make adding variables, renaming, validation, etc. a lot easier.

jcvernaleo (approved):
Seems pretty straight forward how it works and is written.

OK

synctest/synctest.go
```diff
@@ -0,0 +1,556 @@
...
+			panic(err) // should not happen
+		}
+
+		log.Tracef("tbc progress json: %v", string(b))
+
+		if err := json.Unmarshal(b, &syncInfo); err != nil {
+			log.Warningf("error unmarshaling tbc json %s: %s", string(b), err)
+		} else {
+			break
+		}
```
joshuasing: This could be simplified to:
```suggestion
		if err := json.NewDecoder(b).Decode(&syncInfo); err != nil {
			return fmt.Errorf("decode tbc json: %w", err)
		}
```

synctest/README.md
ClaytonNorthey92: I would like to keep these names the same.  I got them from [scientific method](https://en.wikipedia.org/wiki/Scientific_method) terms.  The control group is something that you don't touch that is tested against (i.e. some "working" node), the experiment group is what you are testing against

synctest/synctest.go
ClaytonNorthey92: I can do that, but the reason that I did it this way was to make sure that the tests tested the setting and reading of environment variables.  Our other repositories don't do this (aside from localnet tests, but that doesn't include all variables).

I'll update this to be close to what you're requesting, but I want to make sure the tests still touch them in some way.  give me a few...

synctest/synctest.go
```diff
@@ -0,0 +1,512 @@
...
+
+	if !slices.Contains(validNetworks, c.Network) {
+		return nil, fmt.Errorf("%w: %s", errInvalidNetwork, c.Network)
+	}
+	if !slices.Contains(validSyncmodes, c.Syncmode) {
+		return nil, fmt.Errorf("%w: %s", errInvalidSyncmode, c.Syncmode)
+	}
+
+	return &c, nil
+}
```
joshuasing, line 98: I think we could simplify default values like this:

```suggestion
func DefaultConfig() *config {
	return &config{
		NotifyBySeconds: 60 * 30,
	}
}

func configFromEnv() (*config, error) {
	c := DefaultConfig()
	envOpts := env.Options{Prefix: "SYNCTESTER_"}
	if err := env.ParseWithOptions(&c, envOpts); err != nil {
		return nil, fmt.Errorf("could not parse env: %w", err)
	}

	if c.ControlOpGethEndpoint == "" {
		return nil, errors.New("control op-geth endpoint not set")
	}
	if c.ExperimentalOpGethEndpoint == "" {
		return nil, errors.New("experimental op-geth endpoint not set")
	}

	if c.ExperimentalOpGethTbcHealthEndpoint == "" {
		return nil, errors.New("op-geth tbc health endpoint not set")
	}

	if !slices.Contains(validNetworks, c.Network) {
		return nil, fmt.Errorf("%w: %s", errInvalidNetwork, c.Network)
	}
	if !slices.Contains(validSyncmodes, c.Syncmode) {
		return nil, fmt.Errorf("%w: %s", errInvalidSyncmode, c.Syncmode)
	}

	return c, nil
}
```

synctest/synctest.go
```diff
@@ -0,0 +1,512 @@
...
+type config struct {
+	ControlOpGethEndpoint               string `env:"CONTROL_OP_GETH_ENDPOINT"`
+	ExperimentalOpGethEndpoint          string `env:"EXPERIMENTAL_OP_GETH_ENDPOINT"`
+	ExperimentalOpGethTbcHealthEndpoint string `env:"EXPERIMENTAL_OP_GETH_TBC_HEALTH_ENDPOINT"`
+	SlackOauthToken                     string `env:"SLACK_OAUTH_TOKEN"`
+	SlackChannel                        string `env:"SLACK_CHANNEL"`
+	SlackURL                            string `env:"SLACK_URL"`
+	Network                             string `env:"NETWORK"`
+	Syncmode                            string `env:"SYNCMODE"`
+	NotifyBySeconds                     uint   `env:"NOTIFY_BY_SECONDS"`
```
joshuasing: I think it is more idiomatic to use `time.Duration` instead of seconds here (and `time.Duration` will be parsed from a string by the env library).

```
	NotifyByDuration time.Duration `env:"NOTIFY_BY"`
```

synctest/synctest.go
```diff
@@ -0,0 +1,512 @@
...
+	log = loggo.GetLogger("synctest")
+
+	logLevel, _ := loggo.ParseLevel(c.LogLevel)
+	if logLevel == loggo.UNSPECIFIED {
+		logLevel = loggo.INFO
+	}
+
+	log.Infof("logLevel = %d, logLevel from env = %s", logLevel, c.LogLevel)
+
+	log.SetLogLevel(logLevel)
```
joshuasing: The way we use loggo in other areas of the codebase is quite different. Could we follow the other pattern with `ConfigureLoggers` here?

synctest/synctest.go
```diff
@@ -0,0 +1,512 @@
...
+
+	log.Infof("starting sync testing with configured endpoints: control (%s), experimental(%s)", c.ControlOpGethEndpoint, c.ExperimentalOpGethEndpoint)
+
+	lastNotifiedAt := time.Now().Add(-(time.Duration(c.NotifyBySeconds) * time.Second))
```
joshuasing: https://github.com/hemilabs/heminetwork/pull/865/changes#r2847984703 -- This could then be simplified as (same in other places):

```suggestion
	lastNotifiedAt := time.Now().Add(-c.NotifyByDuration)
```

synctest/synctest.go
```diff
@@ -0,0 +1,512 @@
...
+		}
+	}
+
+	return nil
+}
+
+func reportProgress(ctx context.Context, c *config, lastSyncInfo *tbc.SyncInfo, notify bool) error {
+	var syncInfo tbc.SyncInfo
+	for i := range tbcFetchHealthRetryCount {
+		resp, err := http.Get(fmt.Sprintf("%s", c.ExperimentalOpGethTbcHealthEndpoint))
```
joshuasing: Would we be able to use `http.NewRequestWithContext` here (and with other http requests) so that we can include the context?

synctest/synctest.go
ClaytonNorthey92: will update 👍

synctest/synctest.go
ClaytonNorthey92: yup, will add

synctest/synctest.go
ClaytonNorthey92: will update, going to combine this with other changes

joshuasing (approved):

Issue #918, opened 2026-03-18
trust: create object


ClaytonNorthey92:
### Confirmation

- [x] I have checked for similar issues.

### Problem

create the `trust` ~daemon~ object. this ~daemon~ needs open an embedded database, it does not need an http server (since it's embed-able), and it does not need to connect to peers at this time.

allowing the setting of the network via a flag, this can be one of:
* regtest
* testnet3
* testnet4
* mainnet

don't worry about any database "indexes" or storage yet; just create the database.

### Suggested solution

use as much of the bitcoin rust crate as possible https://docs.rs/bitcoin/latest/bitcoin/

PR #928, opened 2026-03-23
trust: add initial implementation

AL-CT:
**Summary**
Initial implementation of `trust`, the rust version of `tbc`. This only serves as a starting line for the development of `trust`. As such, the underlying database currently only has generic `get`, `put`, `del`, and `has` methods for database tables, and the `trust` fields are largely unused.

Addresses #918 

**Changes**
- Create `Trust` object.
- Create `TrustDB` struct.
- Add `trust` crate modules and export APIs.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -18,6 +18,6 @@ allow = [
     "CC0-1.0",
     "ISC",
     "MIT",
-    "MITNFA", // https://spdx.org/licenses/MITNFA.html
+    "MITNFA", # https://spdx.org/licenses/MITNFA.html
     "Unicode-3.0",
 ]


crates/trust/src/trust_db.rs
```diff
@@ -0,0 +1,173 @@
...
+
+#[derive(Error, Debug)]
+pub enum TrustDBError {
+    #[error("RocksDB error: {0}")]
+    RocksDB(#[from] rocksdb::Error),
+    #[error("Key not found: {0}")]
+    NotFound(String),
+}
+
+pub type Result<T> = std::result::Result<T, TrustDBError>;
```
joshuasing, line 17: Is creating a `Result` type a common practice? Most code I have read uses the `std::result::Result` type and sets the error type in each.

AL-CT: I think I've seen it a few times, but not aliasing is probably more common. However, there are some examples in some well-know crates, and also in the std library, such as [io](https://doc.rust-lang.org/std/io/type.Result.html), [fmt](https://doc.rust-lang.org/std/fmt/type.Result.html), and [thread](https://doc.rust-lang.org/std/thread/type.Result.html). I used it just for convenience, but happy to just make it a standard result 👍

PR #932, opened 2026-03-24
crates/trust: add public set_upstream_state_id function

### LINKED ISSUE #923: trust: add SetUpstreamStateID
### Confirmation

### Problem

As per the [go-version of TBC](https://github.com/hemilabs/heminetwork/blob/40da4645e99ca23b3e8e63f0463e8ed71d52bdfa/service/tbc/tbc.go#L2639), add a call to `trust` that allows us to update the `UpstreamStateID`. This will be required for `reth`, similarly to [op-geth](https://github.com/hemilabs/op-geth/blob/8403b5c35236a886f0a77d310318a9d9228cee78/core/blockchain.go#L1550).

### Suggested solution

Follow the go implementation.

### Additional Information

_No response_

ClaytonNorthey92:
**Summary**
* add public function set_upstream_state_id to trust

fixes #923

**Changes**
see summary

### DIFF
--- a/crates/trust/src/trust.rs
+++ b/crates/trust/src/trust.rs
@@ -1,7 +1,9 @@
-use crate::trust_db::{TrustDB, TrustDBError};
+use crate::trust_db::{TrustDB, TrustDBError, TrustDBTable::MetadataCF};
 use bitcoin::Network;
 use std::path::Path;
 use thiserror::Error;
+
+static DB_METADATA_KEY_UPSTREAM_STATE_ID: &str = "upstream_state_id";
 
 #[derive(Error, Debug)]
 pub enum TrustError {
@@ -54,6 +56,16 @@ impl Trust {
 
         Ok(Self { db, cfg, network })
     }
+
+    pub fn set_upstream_state_id(&self, upstream_state_id: &[u8; 32]) {
+        self.db
+            .put(
+                &MetadataCF,
+                DB_METADATA_KEY_UPSTREAM_STATE_ID,
+                upstream_state_id,
+            )
+            .expect("could not insert into metadata table: {e}")
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add RPC request method whitelist to hproxy ([#691](https://github.com/hemilabs/heminetwork/pull/691)).
 - Add TBC notification system ([#725](https://github.com/hemilabs/heminetwork/pull/725)).
 - Add Continuum Transfuctioner protocol and daemon to handle threshold signatures ([#752](https://github.com/hemilabs/heminetwork/pull/752)).
+- Add set_upstream_state_id to trust [#932](https://github.com/hemilabs/heminetwork/pull/932)
```
joshuasing: I think we can probably leave this out of the changelog for now

crates/trust/src/trust.rs
```diff
@@ -54,6 +56,17 @@ impl Trust {
...
+
+    pub fn set_upstream_state_id(&self, upstream_state_id: &[u8; 32]) {
+        if let Err(e) = &self.db.put(
+            &MetadataCF,
+            DB_METADATA_KEY_UPSTREAM_STATE_ID,
+            upstream_state_id,
+        ) {
+            // should not happen?
+            panic!("could not insert into metadata table: {e}")
+        }
```
joshuasing: I think we probably want this to return the error further up, because this could indicate database corruption, right?

Otherwise, I think it can be simplified as:

```suggestion
		// This should always succeed.
        self.db
        	.put(&MetadataCF, DB_METADATA_KEY_UPSTREAM_STATE_ID, upstream_state_id)
        	.expect("could not insert into metadata table: {e}")
```

ClaytonNorthey92: > I think we probably want this to return the error further up, because this could indicate database corruption, right?

maybe? but, in that case should we `panic`?

> Otherwise, I think it can be simplified as:

I just did what clippy had suggested, I can see what it thinks of your suggestion

Issue #921, opened 2026-03-18
trust: add ExternalHeaderSetup


AL-CT:
### Confirmation

- [x] I have checked for similar issues.

### Problem

As per the [go-version of TBC](https://github.com/hemilabs/heminetwork/blob/40da4645e99ca23b3e8e63f0463e8ed71d52bdfa/service/tbc/tbc.go#L3265), add a setup step to `trust` that opens the database, checks the best header, or inserts genesis.

### Suggested solution

Follow the go implementation.

PR #934, opened 2026-03-25
crates/trust: add header related methods to db

AL-CT:
**Summary**
Sets up the ground work for implementing #921 by creating a `BlockHeader` struct and various related methods to `TrustDB`.

**Changes**
- `BlockHeader`:
  - Created struct.
  - Added `encode_block_header` function.
  - Added `decode_block_header` function.
- `TrustDB`:
  - Changed underlying DB to `OptimisticTransactionDB`.
  - Created `HeightHash` table.
  - Added `height_hash_to_key` function.
  - Added `key_to_height_hash` function.
  - Added `block_header_genesis_insert` method.
  - Added `block_header_best` method.
  - Added `block_header_by_hash` method.
  - Added `block_headers_by_height` method.

### DIFF
--- a/crates/trust/src/trust_db.rs
+++ b/crates/trust/src/trust_db.rs
@@ -6,6 +6,7 @@ use hex::encode;
 use primitive_types::U256;
 use rocksdb::{ColumnFamilyDescriptor, ColumnFamilyRef, OptimisticTransactionDB, Options};
 use std::path::Path;
+use std::slice::Iter;
 use std::sync::Arc;
 use thiserror::Error;
 
@@ -92,7 +93,7 @@ impl TrustDBTable {
         }
     }
 
-    pub fn iterator() -> std::slice::Iter<'static, TrustDBTable> {
+    pub fn iterator() -> Iter<'static, TrustDBTable> {
         static TABLES: [TrustDBTable; 3] = [HeadersCF, MetadataCF, HeightHashCF];
         TABLES.iter()
     }


crates/trust/src/trust_db.rs
```diff
@@ -1,42 +1,98 @@
...
+        let diffb = diff.to_big_endian();
+        enc[88..].copy_from_slice(&diffb);
+
+        enc
+    }
+
+    fn decode_block_header(enc: &[u8; BlockHeader::SIZE]) -> Self {
+        let header =
+            Header::consensus_decode(&mut &enc[8..88]).expect("encoded header should be decodable");
+        let height = u64::from_be_bytes(enc[..8].try_into().unwrap());
```
ClaytonNorthey92: if this fails, are we ok with everything crashing?

AL-CT: I think we should be. The GO code also panics there, but more importantly, if we're calling this on data that we expect to be decodable in this way and it isn't, either the code has a serious problem, or we have database corruption

crates/trust/src/trust_db.rs
```diff
@@ -1,42 +1,105 @@
...
         match self {
             HeadersCF => "blockheaders",
             MetadataCF => "metadata",
+            HeightHashCF => "heighthash",
         }
     }
 
-    pub fn iterator() -> Iter<'static, TrustDBTable> {
-        static TABLES: [TrustDBTable; 2] = [HeadersCF, MetadataCF];
+    pub fn iterator() -> std::slice::Iter<'static, TrustDBTable> {
```
joshuasing: Is there a particular reason to add `std::slice::` here? 🤔

AL-CT: There is not, I think I accidentally removed the import when testing things. Thanks for catching that 👍

Issue #919, opened 2026-03-18
trust: add AddExternalHeaders API call


ClaytonNorthey92:
### Confirmation

- [x] I have checked for similar issues.

### Problem

using the go daemon as a reference, create the AddExternalHeaders API call https://github.com/hemilabs/heminetwork/blob/main/service/tbc/tbc.go#L1437

### Suggested solution

follow the go implementation

PR #949, opened 2026-04-02
trust: add block headers insert

AL-CT:
**Summary**
Creates the DB code to address #919. Mimics the code in [TBC](https://github.com/hemilabs/heminetwork/blob/a72a3f566a9a5f260c05a4825be88e609eab9e78/database/tbcd/level/level.go#L1181).

**Changes**
Adds `block_headers_insert` to `TrustDB`.

### DIFF
--- a/crates/trust/src/trust_db.rs
+++ b/crates/trust/src/trust_db.rs
@@ -520,27 +520,27 @@ mod tests {
         TrustDB::open(tmp.path()).expect("database should open")
     }
 
-    fn create_test_header(parent_hash: BlockHash, nonce: u32) -> Header {
+    fn create_test_header(parent_hash: BlockHash, n: u32) -> Header {
         Header {
             version: bitcoin::blockdata::block::Version::ONE,
             prev_blockhash: parent_hash,
             merkle_root: bitcoin::TxMerkleNode::all_zeros(),
             time: 12345,
             bits: bitcoin::CompactTarget::from_consensus(0x1d00ffff),
-            nonce,
+            nonce: n,
         }
     }
 
     // Changing the bits can cause an overflow with invalid data, so
     // this function guarantees you only change them if necessary.
-    fn create_test_header_with_bits(parent_hash: BlockHash, nonce: u32, bits: u32) -> Header {
+    fn create_test_header_with_bits(parent_hash: BlockHash, n: u32, bits: u32) -> Header {
         Header {
             version: bitcoin::blockdata::block::Version::ONE,
             prev_blockhash: parent_hash,
             merkle_root: bitcoin::TxMerkleNode::all_zeros(),
             time: 12345,
             bits: bitcoin::CompactTarget::from_consensus(bits),
-            nonce,
+            nonce: n,
         }
     }
 
@@ -959,10 +959,10 @@ mod tests {
         db: &TrustDB,
         prev_hash: BlockHash,
         height: u64,
-        nonce: u32,
+        n: u32,
     ) -> Result<(BlockHeader, InsertType)> {
-        let header = create_test_header(prev_hash, nonce);
-        let difficulty = U256::from(nonce);
+        let header = create_test_header(prev_hash, n);
+        let difficulty = U256::from(n);
 
         if prev_hash == BlockHash::all_zeros() {
             db.block_header_genesis_insert(&header, height, difficulty)?;


crates/trust/src/trust_db.rs
```diff
@@ -78,6 +82,23 @@ impl From<&BlockHeader> for EncodedHeader {
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub enum InsertType {
+    ITChainExtend, // Normal insert, does not require further action.
+    ITChainFork,   // Chain forked, unwind and rewind indexes.
+    ITForkExtend,  // Extended a fork, does not require further action.
```
joshuasing: Since these are part of an enum, I don't think it is necessary to prefix them with `IT`.
```suggestion
    ChainExtend, // Normal insert, does not require further action.
    ChainFork,   // Chain forked, unwind and rewind indexes.
    ForkExtend,  // Extended a fork, does not require further action.
```

crates/trust/src/trust_db.rs
```diff
@@ -248,6 +271,151 @@ impl TrustDB {
...
+        // have in the database.
+        let mut x: usize = 0;
+        for rbh in headers {
+            let hash = rbh.block_hash();
+            if batch
+                .get_pinned_cf(self.get_cf(&HeadersCF), hash)?
+                .is_none()
+            {
+                break;
+            }
```
joshuasing: I get Rust is not exactly known for readability, but this is a bit too hard to read lol... maybe making it a variable above the if would be more readable? 🤔 

There's another similar multi-line `if` statement further down as well.

crates/trust/src/trust_db.rs
```diff
@@ -248,6 +271,151 @@ impl TrustDB {
...
+        &self,
+        headers: &[Header],
+        hooks: &[BatchHook],
+    ) -> Result<(InsertType, BlockHeader, BlockHeader, usize)> {
+        if headers.is_empty() {
+            return Err(TrustDBError::Other(
+                "block headers insert: invalid".to_string(),
+            ));
+        }
+        let batch = self.db.transaction();
```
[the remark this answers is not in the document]
ClaytonNorthey92: looks like it might be handled in the `Drop` implementation https://docs.rs/rocksdb/0.24.0/src/rocksdb/transactions/transaction.rs.html#898

even so, could we explicitly call `rollback` when we return with errors?

crates/trust/src/trust_db.rs
```diff
@@ -605,4 +788,244 @@ mod tests {
...
+        db: &TrustDB,
+        prev_hash: BlockHash,
+        height: u64,
+        nonce: u32,
+    ) -> Result<(BlockHeader, InsertType)> {
+        let header = create_test_header(prev_hash, nonce);
+        let difficulty = U256::from(nonce);
+
+        if prev_hash == BlockHash::all_zeros() {
+            db.block_header_genesis_insert(&header, height, difficulty)?;
```
ClaytonNorthey92, line 968: I think we should error here if we already have a genesis in the database, no?  it looks like `block_header_genesis_insert` only checks if we already have this specific genesis? but if we try to insert two different genesis blocks would it succeed? 🤔

ClaytonNorthey92: something like this, should this pass? I don't think it should but it does

```rs
    #[test]
    fn test_block_headers_insert_a_bunch_of_genesis_blocks() {
        let db = new_test_db();

        let mut genesis = create_test_header(BlockHash::all_zeros(), 0);
        db.block_header_genesis_insert(&genesis, 0, U256::from(1))
            .unwrap();

        genesis = create_test_header(BlockHash::all_zeros(), 2);
        db.block_header_genesis_insert(&genesis, 0, U256::from(2))
            .unwrap();

        genesis = create_test_header(BlockHash::all_zeros(), 3);
        db.block_header_genesis_insert(&genesis, 0, U256::from(3))
            .unwrap();
    }
```

AL-CT: This follows `TBC`, and I think arbitrarily being able to pass in genesis blocks is intentional, as `block_header_genesis_insert` essentially works in the same way as `block_headers_insert`, but without checking that their parent exists. You can also have an effective genesis block at different heights and hashes, which are (or were) determined by `op-geth`. 

Preventing multiple genesis blocks is probably possible by storing a key in the metadata, which would point to the genesis hash, but then, when inserting another one, do we replace / delete the previous one? That would require invalidating all other headers. Do we prevent the insertion? That might make it quite hard to substitute the genesis. 

The easiest solution is probably to define the expected genesis at startup, which does happen on the `Trust` side, and then managing the calling code to this method. However, we can try and add something to prevent the behavior all together, just might have to weigh the downsides 🤔

crates/trust/src/trust_db.rs
AL-CT: Good point, fixed 👍

crates/trust/src/trust_db.rs
AL-CT: Haha, I might have become too used to looking at the code and somehow convinced myself that "Yeah, this is fine". I made them into their own vars, hopefully a little easier to follow.

crates/trust/src/trust_db.rs
AL-CT: Created an issue for this: #951

crates/trust/src/trust_db.rs
```diff
@@ -351,20 +520,28 @@ mod tests {
         TrustDB::open(tmp.path()).expect("database should open")
     }
 
-    fn create_test_header(nonce: u32) -> Header {
+    fn create_test_header(parent_hash: BlockHash, nonce: u32) -> Header {
```
joshuasing: Can we rename `nonce` here to `n` or similar to (hopefully) prevent CodeQL/GHAS from having a stroke every time we update tests using it?

AL-CT: Fixed 👍

PR #950, opened 2026-04-03
tbc: fix header contiguity check

AL-CT:
**Summary**
The insertion of headers in the `TBC` database requires the headers to be contiguous, which is guaranteed by the caller. However, one of the potential calls (`handleHeaders`) had erroneous logic, which led to non-contiguous headers being allowed in. This PR addresses this flaw, and guarantees that the passed in headers **must** be contiguous before being inserted.

**Changes**
Fix the contiguity check in `handleHeaders` and add the test `TestIndexFakeHeaders` that fails if non-contiguous headers are inserted into the DB. This test passes with the new logic, but did not pass when using the original logic.

**Note**: Also tested by performing IBD up to `150000` blocks for both `mainnet` and `testnet3`.

**Localnet test passing [here](https://github.com/hemilabs/heminetwork/actions/runs/23949569849)**.

### DIFF
--- a/service/tbc/tbc.go
+++ b/service/tbc/tbc.go
@@ -1554,8 +1554,7 @@ func (s *Server) handleHeaders(ctx context.Context, p *rawpeer.RawPeer, msg *wir
 			return fmt.Errorf("cannot connect %v index %v",
 				msg.Headers[k].PrevBlock, k)
 		}
-		bh := msg.Headers[k].BlockHash()
-		pbhHash = &bh
+		pbhHash = new(msg.Headers[k].BlockHash())
 	}
 
 	// When running in normal (not External Header) mode, do not set


go.mod
```diff
@@ -23,7 +23,6 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/otiai10/copy v1.14.1
 	github.com/prometheus/client_golang v1.23.2
-	github.com/sethvargo/go-retry v0.3.0
```
jcvernaleo: Unless I'm missing something, this looks unrelated to the change, right?

AL-CT: Oh yeah, that was from me running `make`. Strange 🤔
I can revert that once all changes are made, just not sure why it wasn't removed before.

jcvernaleo: Cool, I'll ok once it is all clean and those are out.

service/tbc/tbc.go
```diff
@@ -1550,11 +1550,12 @@ func (s *Server) handleHeaders(ctx context.Context, p *rawpeer.RawPeer, msg *wir
...
 	var pbhHash *chainhash.Hash
 	for k := range msg.Headers {
-		if pbhHash != nil && pbhHash.IsEqual(&msg.Headers[k].PrevBlock) {
+		if pbhHash != nil && !pbhHash.IsEqual(&msg.Headers[k].PrevBlock) {
 			return fmt.Errorf("cannot connect %v index %v",
 				msg.Headers[k].PrevBlock, k)
 		}
-		pbhHash = &msg.Headers[k].PrevBlock
+		bh := msg.Headers[k].BlockHash()
+		pbhHash = &bh
```
joshuasing: Since Go 1.26, the built-in `new` function can be used to create a pointer to a value: https://go.dev/doc/go1.26#language

I believe this _should_ work:
```suggestion
		pbhHash = new(msg.Headers[k].BlockHash())
```

go.mod
joshuasing: I must have forgotten to run `go mod tidy` on https://github.com/hemilabs/heminetwork/pull/930 - which was the only package that used this dependency.
I have no objection to including this here, but I can also make a separate PR for it if wanted.

service/tbc/tbc.go
AL-CT: I'm really glad they added that, fixed 👍

marcopeereboom (approved):

Issue #967, opened 2026-04-13
tbc: fragile test


ClaytonNorthey92:
### Confirmation

- [x] I have checked for similar issues.

### Describe the bug

this test is failing on a PR that does not modify tbc code, indicating its fragility

```
2026-04-13 17:26:02 ERROR tbcgozer tbcgozer.go:316 handleTBCWebsocketCall *tbcapi.UTXOsByAddressRequest: context deadline exceeded
panic: context deadline exceeded [recovered, repanicked]

goroutine 3522 [running]:
sync.(*WaitGroup).Go.func1.1()
	/opt/hostedtoolcache/go/1.26.2/x64/src/sync/waitgroup.go:251 +0x45
panic({0xbefac0?, 0x13fd100?})
	/opt/hostedtoolcache/go/1.26.2/x64/src/runtime/panic.go:860 +0x13a
github.com/hemilabs/heminetwork/v2/bitcoin/wallet/gozer/tbcgozer.TestTBCGozerCalls.func1()
	/home/runner/work/heminetwork/heminetwork/bitcoin/wallet/gozer/tbcgozer/tbcgozer_test.go:239 +0x571
sync.(*WaitGroup).Go.func1()
	/opt/hostedtoolcache/go/1.26.2/x64/src/sync/waitgroup.go:258 +0x4a
created by sync.(*WaitGroup).Go in goroutine 2577
```

### Expected behaviour

tests should not be fragile

### Environment

Version: 
Operating System: 
Architecture: 

### Additional Information

run: https://github.com/hemilabs/heminetwork/actions/runs/24357099002/job/71126802759?pr=958

PR #969, opened 2026-04-15
tbcgozer: limit in-flight goroutine calls

AL-CT:
**Summary**
Addresses #967.

`tbcGozer` will wait for a request, and then spawn a goroutine that performs a call to the `TBC` server (`handleTBCWebsocketCall`). The calls are limited by a `chan` of a certain depth. However, since the processing of the call is done inside a goroutine, these requests get consumed almost instantly.

At the same time, a reader (`handleTBCWebsocketRead`) goroutine continously reads the responses from the `TBC` server, and sends it to the corresponding caller. 

However, both the caller and reader goroutines acquire a `RWLock` for the connection internally (to create and retrieve channels to route the responses to, respectively). There is only a single reader, but multiple callers, which can lead to the starvation of the reader, who must wait a very long time to acquire the lock.

The test `TestTBCGozerCalls` spawns 1000 goroutines, each performing 4 calls to the mock `TBC` server. There is also a limit to how long a caller will wait for a response (10s) before timing out. In CI, a reader would sometimes be delayed in processing responses, which would lead to certain callers timing out, waiting for a response.

**Changes**
Adds a semaphore `handleTBCWebsocketCall`, preventing unlimited calls to occur simultaneously, starving the response reader.

**NOTE**: The test was run 600 times, split between two actions runs, with no failures. Results can be seen [here](https://github.com/hemilabs/heminetwork/actions/runs/24449952738) and [here](https://github.com/hemilabs/heminetwork/actions/runs/24451002306).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix bug that allowed invalid headers to be indexed ([#950](https://github.com/hemilabs/heminetwork/pull/950))
 
-- Fix bug that let to delayed request processing in tbcgozer ([#969](https://github.com/hemilabs/heminetwork/pull/969))
+- Fix bug that led to delayed request processing in tbcgozer ([#969](https://github.com/hemilabs/heminetwork/pull/969))
 
 ## [v2.0.0]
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -39,6 +39,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix bug that led to early exit during Daemon configuration parsing ([#885](https://github.com/hemilabs/heminetwork/pull/885))
 
+- Fix bug that let to delayed request processing in tbcgozer ([#969](https://github.com/hemilabs/heminetwork/pull/969))
```
joshuasing: s/let/led/

joshuasing (approved):

CHANGELOG.md
AL-CT: Oops, thanks 👍

ClaytonNorthey92 (approved):

PR #972, opened 2026-04-19
tbcd: restore witness data in block storage and add v5 db upgrade

marcopeereboom:
tbcd was asking peers for blocks and txs using the base
(non-witness) inventory types, so under BIP-144 every reply
came back witness-stripped.  Every block body tbcd has on
disk since segwit activation is missing its witness data;
reading a tx back through the RPC shows txid == wtxid and
empty TxIn.Witness.

This PR switches the three outbound getdata call sites to
the witness variants and advertises SFNodeWitness in the
handshake.  The reply correlation continues to work because
GetData, onInvHandler, and onNotFoundHandler strip the
witness bit before tagging, keeping the pending-reply key
aligned with what the base-form inbound handlers expect.

The stored bodies are unrecoverable in place — the bytes
never arrived.  A v5 database upgrade wipes BlocksDB and
rebuilds BlocksMissingDB from the header chain so the normal
sync loop re-downloads every block with full witness data.
Headers, height-hash, transactions, outputs, keystones, and
indexer tips are witness-independent and left untouched;
re-processing the re-downloaded blocks produces identical
output, so indexers stay at their current heights.

Verified end-to-end on testnet4: after the fix, a reveal tx
at block 131115 comes back with a 3-element witness stack
and wtxid != txid, matching the reference node byte for byte.

First start on an existing home directory runs the upgrade
in place.  BlockByHash / TxById return NotFound for blocks
that have not yet been re-fetched during the resync window,
which is the same behaviour dependents already see during
cold IBD.

Operators should expect the BlocksDB to grow substantially after
the upgrade since witness data now lands on disk.  On testnet4
the blocks directory went from 3.3 G at v4 to 12 G at v5 (~3.6×),
and total home went from 9.1 G to 17 G.  Mainnet will be larger
in absolute terms, similar in ratio — post-segwit-adoption blocks
are mostly witness, especially after taproot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Improved signal handling in the daemons ([#763](https://github.com/hemilabs/heminetwork/pull/763)).
 
+- Bump tbcd database schema to v5; first start after upgrade
+  wipes the stored block bodies and re-downloads them with
+  witness data, triggered by the witness-download fix below.
+
 ### Fixed
 
 - Fix typos across the codebase
@@ -43,6 +47,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix bug that allowed invalid headers to be indexed ([#950](https://github.com/hemilabs/heminetwork/pull/950))
 
 - Fix bug that led to delayed request processing in tbcgozer ([#969](https://github.com/hemilabs/heminetwork/pull/969))
+
+- Fix tbcd requesting witness-stripped blocks and txs from
+  peers (BIP-144); on-disk blocks are now witness-inclusive
+  after a v5 upgrade plus resync.
 
 ## [v2.0.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,9 +32,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Improved signal handling in the daemons ([#763](https://github.com/hemilabs/heminetwork/pull/763)).
 
-- Bump tbcd database schema to v5; first start after upgrade
-  wipes the stored block bodies and re-downloads them with
-  witness data, triggered by the witness-download fix below.
+- Bump tbcd database schema to v5; first start after upgrade wipes the stored block bodies and re-downloads them
+  with witness data, triggered by the witness-download fix below
+  ([#972](https://github.com/hemilabs/heminetwork/pull/972)).
 
 ### Fixed
 
@@ -48,9 +48,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix bug that led to delayed request processing in tbcgozer ([#969](https://github.com/hemilabs/heminetwork/pull/969))
 
-- Fix tbcd requesting witness-stripped blocks and txs from
-  peers (BIP-144); on-disk blocks are now witness-inclusive
-  after a v5 upgrade plus resync.
+- Fix tbcd requesting witness-stripped blocks and txs from peers (BIP-144); on-disk blocks are now
+  witness-inclusive after a v5 upgrade plus resync
+  ([#972](https://github.com/hemilabs/heminetwork/pull/972)).
 
 ## [v2.0.0]
 
--- a/database/tbcd/level/level_test.go
+++ b/database/tbcd/level/level_test.go
@@ -847,6 +847,204 @@ func encodeKeystoneToSliceV1(ks tbcd.Keystone) []byte {
 	return eks[:]
 }
 
+// TestDbUpgradeV5 validates the v4 -> v5 upgrade that wipes
+// witness-stripped block bodies and rebuilds BlocksMissingDB from
+// the header chain.  The upgrade must:
+//
+//   - empty BlocksDB rawdb completely (both index and data);
+//   - populate BlocksMissingDB with one heightHashToKey entry per
+//     row in BlockHeadersDB (skipping the canonical-tip sentinel);
+//   - leave every other database untouched — specifically
+//     TransactionsDB, OutputsDB, KeystonesDB, BlockHeadersDB,
+//     HeightHashDB, and the indexer tips in MetadataDB — because
+//     those are all witness-independent;
+//   - bump the schema version from 4 to 5.
+func TestDbUpgradeV5(t *testing.T) {
+	const blkNum uint64 = 5
+
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
+
+	home := t.TempDir()
+	t.Logf("temp: %v", home)
+	cfg, err := NewConfig("upgradetest", home, "0mb", "0mb")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Phase 1: open in upgrade-skipping mode so the fresh-db
+	// version stamp can be overwritten with 4 before close.
+	// Build a short chain of headers + blocks, plant survivor
+	// rows in the tables v5 must not touch, then stamp the
+	// version down so reopen runs v5.
+	cfg.SetUpgradeOpen(true)
+	db, err := New(ctx, cfg)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	hashes := make([]chainhash.Hash, 0, blkNum)
+	heights := make(map[chainhash.Hash]uint64, blkNum)
+	prevHash := &chainhash.Hash{}
+	for i := range blkNum {
+		bh, _, err := insertBlockHeader(ctx, db, prevHash, i, i)
+		if err != nil {
+			t.Fatalf("insertBlockHeader %v: %v", i, err)
+		}
+		if err := insertBlock(ctx, db, bh); err != nil {
+			t.Fatalf("insertBlock %v: %v", i, err)
+		}
+		h := bh.BlockHash()
+		hashes = append(hashes, *h)
+		heights[*h] = bh.Height
+		prevHash = h
+	}
+
+	// Survivor rows: arbitrary bytes written to tables whose
+	// contents v5 must not touch.  The keys are deliberately
+	// unique so a post-upgrade Get confirms exact preservation.
+	survivors := map[string][2][]byte{
+		level.TransactionsDB: {
+			[]byte("v5-test-tx-key"),
+			[]byte("v5-test-tx-value"),
+		},
+		level.OutputsDB: {
+			[]byte("v5-test-utxo-key"),
+			[]byte("v5-test-utxo-value"),
+		},
+		level.KeystonesDB: {
+			[]byte("v5-test-keystone-key"),
+			[]byte("v5-test-keystone-value"),
+		},
+	}
+	for table, kv := range survivors {
+		if err := db.insertTable(table, kv[0], kv[1]); err != nil {
+			t.Fatalf("seed %v: %v", table, err)
+		}
+	}
+
+	// Sanity: BlocksDB rawdb actually holds the seeded blocks
+	// before the upgrade runs.  v5's correctness story is "we
+	// wiped what was there"; without this check a silently
+	// empty pre-state would make the post-upgrade empty
+	// assertion vacuous.
+	preBdb := db.rawPool[level.BlocksDB]
+	preIt := preBdb.DB().NewIterator(nil, nil)
+	var preCount int
+	for preIt.Next() {
+		preCount++
+	}
+	preIt.Release()
+	if err := preIt.Error(); err != nil {
+		t.Fatal(err)
+	}
+	if preCount != int(blkNum) {
+		t.Fatalf("pre-upgrade BlocksDB: got %v entries want %v",
+			preCount, blkNum)
+	}
+
+	v4 := make([]byte, 8)
+	binary.BigEndian.PutUint64(v4, 4)
+	if err := db.MetadataPut(ctx, versionKey, v4); err != nil {
+		t.Fatal(err)
+	}
+	if err := db.Close(); err != nil {
+		t.Fatal(err)
+	}
+
+	// Phase 2: reopen with the upgrade ladder enabled.  The
+	// stamped v4 triggers v5 end-to-end.
+	cfg.SetUpgradeOpen(false)
+	db2, err := New(ctx, cfg)
+	if err != nil {
+		t.Fatalf("reopen (runs v5): %v", err)
+	}
+	defer func() {
+		if err := db2.Close(); err != nil {
+			t.Fatal(err)
+		}
+	}()
+
+	// Assertion 1: BlocksDB rawdb is empty after v5.
+	bdb := db2.rawPool[level.BlocksDB]
+	it := bdb.DB().NewIterator(nil, nil)
+	var bcount int
+	for it.Next() {
+		bcount++
+	}
+	it.Release()
+	if err := it.Error(); err != nil {
+		t.Fatal(err)
+	}
+	if bcount != 0 {
+		t.Fatalf("BlocksDB not empty after v5: %v entries", bcount)
+	}
+
+	// Assertion 2: BlocksMissingDB has exactly one entry per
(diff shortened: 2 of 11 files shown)


database/tbcd/level/upgrade.go
```diff
@@ -624,3 +624,136 @@ func (l *ldb) v4(ctx context.Context) error {
...
+	if !ok || bmDB == nil {
+		return fmt.Errorf("db not found: %v", level.BlocksMissingDB)
+	}
+	clearBatch := new(leveldb.Batch)
+	clearIt := bmDB.NewIterator(nil, nil)
+	for clearIt.Next() {
+		// Copy the key because the iterator reuses its buffer
+		// across Next calls; leveldb.Batch stores a reference
+		// until Write.
+		k := append([]byte(nil), clearIt.Key()...)
```
AL-CT: I think in this scenario it's worth using clone() and pre-allocating the size of the array. It's a little less readable, but should be faster

database/tbcd/level/upgrade.go
```diff
@@ -624,3 +624,136 @@ func (l *ldb) v4(ctx context.Context) error {
...
+		return fmt.Errorf("remove blocks rawdb tree: %w", err)
+	}
+	if err := bdb.Open(); err != nil {
+		return fmt.Errorf("reopen blocks rawdb: %w", err)
+	}
+
+	// 2. Clear BlocksMissingDB.  Iterate and batch-delete rather
+	// than removing the leveldb files; the pool still holds the
+	// open handle and we want it to stay usable in step 3.
+	bmDB, ok := l.pool[level.BlocksMissingDB]
```
AL-CT: Is it not worth it to add some methods to pool that allow you to close and open individual tables, so that we can delete the files rather than iterating?

service/tbc/peer/peer_witness_test.go
```diff
@@ -0,0 +1,387 @@
...
+	return p, serverRP
+}
+
+// srvReply writes a reply after a brief yield so call() in the
+// client goroutine enters its select before the handler's
+// non-blocking send fires.  net.Pipe is synchronous, so without
+// this the reply can arrive before the caller is ready.
+func srvReply(t *testing.T, srv *rawpeer.RawPeer, msg wire.Message) {
+	t.Helper()
+	time.Sleep(time.Millisecond)
```
AL-CT: I think we should avoid `time.Sleep` in our tests. If the peer response routing can't be blocking, and the response channel can't buffered, we should use the [synctest package](https://go.dev/blog/synctest) in these tests and use `synctest.Wait()` before calling this function, ensuring the caller goroutine is blocked in the select.

database/tbcd/level/level_test.go
```diff
@@ -1352,3 +1551,162 @@ func TestBlockHeadersRemove(t *testing.T) {
...
+				// final MetadataPut(versionKey, 5) at
+				// end of v5 fails.
+				if err := db.pool[level.MetadataDB].Close(); err != nil {
+					t.Fatal(err)
+				}
+			},
+			// MetadataPut wraps through startTransaction
+			// which returns "unable to open db transaction"
+			// on a closed handle.
+			wantErrIn: "",
```
AL-CT: Without this, how do we know we received the expected error?

AL-CT (changes requested):
Just a few comments, otherwise LGTM 👍

AL-CT (approved):

database/tbcd/level/upgrade.go
marcopeereboom: Done — replaced the iterate-and-batch-delete with a close-all, delete directories, reopen-all cycle. Eliminates the O(n) key iteration entirely. No clone or pre-allocate needed since the iteration no longer exists.

database/tbcd/level/upgrade.go
marcopeereboom: Done — close-delete-reopen cycle via level.New. See 57b35b7.

service/tbc/peer/peer_witness_test.go
marcopeereboom: Done — replaced time.Sleep with testing/synctest.Wait. See b7f967b.

database/tbcd/level/level_test.go
marcopeereboom: Done — TestDbUpgradeV5Errors covers seven error paths with explicit assertions. See 9170bf2.

database/tbcd/level/level_test.go
```diff
@@ -1352,3 +1551,97 @@ func TestBlockHeadersRemove(t *testing.T) {
...
+				// pool entries so a final Close may or
+				// may not succeed.
+				_ = db.Close()
+			}()
+
+			// Seed one header + one block so step 1 has
+			// something to wipe and step 2 has a row to
+			// rebuild from.  Skip for the malformed-value
+			// case, which plants its own bogus header.
+			if c.name != "malformed blockheader value size" {
```
joshuasing: Instead of basing this on the case name, maybe a `bool` can be added to flip this behaviour for the test case?

database/tbcd/level/level_test.go
```diff
@@ -1352,3 +1551,97 @@ func TestBlockHeadersRemove(t *testing.T) {
...
+				if err := db.insertTable(level.BlockHeadersDB,
+					h[:], []byte("short")); err != nil {
+					t.Fatal(err)
+				}
+			},
+			wantErrIn: "blockheader value size",
+		},
+	}
+
+	for _, c := range cases {
```
joshuasing: Normally we tend to use different naming here, struct `test`, `tests` and `tt`:

```suggestion
	tests := []struct {
		name      string
		setup     func(t *testing.T, db *ldb)
		wantErrIn string
	}{
		{
			name: "close database fails",
			setup: func(t *testing.T, db *ldb) {
				// Pre-close a handle so that v5's
				// initial l.Close() encounters a
				// double-close error.
				if err := db.pool[level.BlocksMissingDB].Close(); err != nil {
					t.Fatal(err)
				}
			},
			wantErrIn: "close database",
		},
		{
			name: "malformed blockheader value size",
			setup: func(t *testing.T, db *ldb) {
				// Plant a row whose key looks like a
				// 32-byte hash but whose value is
				// shorter than blockheaderSize.
				h := testutil.String2Hash(
					"1111111111111111111111111111111111111111111111111111111111111111")
				if err := db.insertTable(level.BlockHeadersDB,
					h[:], []byte("short")); err != nil {
					t.Fatal(err)
				}
			},
			wantErrIn: "blockheader value size",
		},
	}

	for _, tt := range tests {
```

database/tbcd/level/upgrade.go
```diff
@@ -624,3 +624,124 @@ func (l *ldb) v4(ctx context.Context) error {
...
+		target := filepath.Join(l.cfg.Home, sub)
+		rel, err := filepath.Rel(l.cfg.Home, target)
+		if err != nil || rel == "." || rel == ".." || filepath.IsAbs(rel) {
+			return fmt.Errorf("refusing to remove %v: not a child of %v",
+				target, l.cfg.Home)
+		}
+		if err := os.RemoveAll(target); err != nil {
+			return fmt.Errorf("remove %v: %w", sub, err)
+		}
+	}
```
joshuasing, line 695: `l.cfg.Home` does not appear to be validated after being set. If it's empty or relative, `filepath.Join` could produce a relative path and `os.RemoveAll` deletes relative to the working directory.

In `level.go` (around line 191), it would be good to add a check after `homedir.Expand(filepath.Join(home, network))` - validate the result is absolute.

The traversal check here also has two issues:
- `filepath.IsAbs(rel)` is always false after a successful `filepath.Rel`.
- I think the rel check is redundant since `sub` is a constant.

Instead, I think it can be replaced with a prefix check. Also add a symlink guard if anything outside of the process creates a symlink in its place.

```suggestion
	home := filepath.Clean(l.cfg.Home)
	for _, sub := range []string{level.BlocksDB, level.BlocksMissingDB} {
		target := filepath.Join(home, sub)
		if !strings.HasPrefix(target, home+string(os.PathSeparator)) {
			return fmt.Errrof("refusing to remove %q: not a child of %q",
				target, home)
		}
		if fi, err := os.Lstat(target); err == nil && fi.Mode()&os.ModeSymlink != 0 {
			return fmt.Errorf("refusing to remove %q: is a symlink", target)
		}
		if err := os.RemoveAll(target); err != nil {
			return fmt.Errorf("remove %v: %w", sub, err)
		}
	}
```

service/tbc/peer/peer_witness_test.go
```diff
@@ -0,0 +1,398 @@
...
+		if _, ok := r.msg.(*wire.MsgNotFound); !ok {
+			t.Fatalf("expected *wire.MsgNotFound, got %T", r.msg)
+		}
+	})
+}
+
+// TestGetBlockWitness exercises the GetBlock wrapper end-to-end:
+// GetHeaders exchange followed by GetData with InvTypeWitnessBlock.
+func TestGetBlockWitness(t *testing.T) {
+	synctest.Test(t, func(t *testing.T) {
```
joshuasing, line 273: Yay synctest! :D

service/tbc/tbc_test.go
```diff
@@ -180,8 +180,8 @@ func TestDbUpgradeFull(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if version != 4 {
-		t.Fatalf("expected version 4, got %v", version)
+	if version != 5 {
+		t.Fatalf("expected version 5, got %v", version)
```
joshuasing, line 184: Could this be moved to a constant?

CHANGELOG.md
```diff
@@ -32,6 +32,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Improved signal handling in the daemons ([#763](https://github.com/hemilabs/heminetwork/pull/763)).
 
+- Bump tbcd database schema to v5; first start after upgrade
+  wipes the stored block bodies and re-downloads them with
+  witness data, triggered by the witness-download fix below.
```
joshuasing: Missing PR number and link, also over-wrapped - wrap markdown to 120 characters.

CHANGELOG.md
```diff
@@ -44,6 +48,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix bug that led to delayed request processing in tbcgozer ([#969](https://github.com/hemilabs/heminetwork/pull/969))
 
+- Fix tbcd requesting witness-stripped blocks and txs from
+  peers (BIP-144); on-disk blocks are now witness-inclusive
+  after a v5 upgrade plus resync.
```
joshuasing: Missing PR number and link, also overwrapped - wrap markdown to 120 characters.

joshuasing (changes requested):

service/tbc/tbc_test.go
marcopeereboom: All the surrounding code is the same.

database/tbcd/level/level_test.go
marcopeereboom: Done — added `seedBlock bool` field to the test struct. See 6289c0f.

database/tbcd/level/level_test.go
marcopeereboom: Done — renamed to `test`/`tests`/`tt`. See 6289c0f.

database/tbcd/level/upgrade.go
marcopeereboom: Done — replaced the broken filepath.Rel check with filepath.Clean + strings.HasPrefix prefix check + os.Lstat symlink guard. Also added filepath.IsAbs validation in NewConfig after homedir.Expand. See 6289c0f.

CHANGELOG.md
marcopeereboom: Done — added PR link, re-wrapped to 120 chars. See 6289c0f.

CHANGELOG.md
marcopeereboom: Done — added PR link, re-wrapped to 120 chars. See 6289c0f.

database/tbcd/level/upgrade.go
```diff
@@ -624,3 +625,127 @@ func (l *ldb) v4(ctx context.Context) error {
...
+//     makes returns the full set for download.
+//
+// First start after the upgrade triggers a full re-download of the
+// chain's block bodies. BlockByHash and TxById return NotFound for
+// any block that has not yet been re-fetched during that window.
+func (l *ldb) v5(ctx context.Context) error {
+	log.Tracef("v5")
+	defer log.Tracef("v5 exit")
+
+	log.Infof("Upgrading database from v4 to v5")
```
ClaytonNorthey92, line 669: is this always true? could a user upgrade from v3 to v5?  should we reflect that? or prevent it?

database/tbcd/level/upgrade.go
```diff
@@ -624,3 +625,127 @@ func (l *ldb) v4(ctx context.Context) error {
...
+				target, home)
+		}
+		if fi, err := os.Lstat(target); err == nil && fi.Mode()&os.ModeSymlink != 0 {
+			return fmt.Errorf("refusing to remove %q: is a symlink", target)
+		}
+		if err := os.RemoveAll(target); err != nil {
+			return fmt.Errorf("remove %v: %w", sub, err)
+		}
+	}
+	ld, err := level.New(ctx, level.NewDefaultConfig(l.cfg.Home))
```
ClaytonNorthey92: do we _want_ a new default config here? Or a copy of the current `l.cfg`?

database/tbcd/level/upgrade.go
```diff
@@ -624,3 +625,127 @@ func (l *ldb) v4(ctx context.Context) error {
...
+//     observe the witness
+//   - OutputsDB       kept: utxo set is computed from TxIn and
+//     TxOut, both of which precede the witness section
+//   - KeystonesDB     kept: PoP commitments live in coinbase
+//     OP_RETURNs, witness-independent
+//   - metadata indexer tips kept: re-processing the re-downloaded
+//     blocks would produce identical output, so the tips can stay
+//
+// Changes:
+//   - BlocksDB rawdb: wipe both index and data files, then reopen
```
ClaytonNorthey92, line 656: I had mentioned this on slack: we may not be able to get all of the blocks solely from the bitcoin p2p network. we have non-canonical bitcoin blocks in tbc for the hemi chain's bitcoin attributes deposited transactions.  these _very likely_ are not available over bitcoin p2p. it _might_ be better to wipe, resync, then let op-geth fill those out.

database/tbcd/level/upgrade.go
joshuasing: Based on https://github.com/hemilabs/heminetwork/pull/972/changes#diff-dc7392654ddb3d0c6bb48e13c7d49962953288d11417b656d3c887dc28758611R297-R300, I believe this function will only be called if the version is v4. I wouldn't mind a sanity check at the top of each upgrade function though.

joshuasing (approved):

database/tbcd/level/upgrade.go
marcopeereboom: The for-loop dispatcher in level.go handles sequential upgrades (v3→v4, then v4→v5 on the next iteration), so v5() is only entered when dbVersion==4. Added a defensive sanity check at the top of v5() that asserts version 4 before proceeding.

database/tbcd/level/upgrade.go
marcopeereboom: I was going to merge but this needs another look.

database/tbcd/level/upgrade.go
marcopeereboom: Valid concern. The v5 wipe re-downloads via the normal sync loop which only fetches from bitcoin p2p. Non-canonical blocks (hemi chain BAD txs) would need to come back through op-geth. Need to verify whether the sync loop handles that path already or if explicit backfill logic is needed.

database/tbcd/level/upgrade.go
marcopeereboom: Good catch — fixed. v5 now saves the existing config via a new Config() getter on level.Database and reopens with it instead of constructing a new default. See e875955.

PR #1003, opened 2026-04-29
tbc: add admin rpc

### LINKED ISSUE #983: tbc: add database controller RPC API
### Confirmation

### Problem

In order to be able to execute commands that modify the `TBC` database state (e.g., `SyncIndexersToHash`) through RPC, we need to implement these methods.

Accepting these commands likely has to be limited by authentication or a similar mechanism, so that only verified connections can perform these commands. Furthermore, these commands are often long-lived, so a "job queue" that informs the caller of the result of a request at a later date may also be necessary.

### Suggested solution

Create an authenticated route for `tbcapi` (using either the pre-existing priv key authentication API used for the original `BFG` server, or a bearer token) that allows certain connections to submit admin-only requests. Alternatively, create a separate route or http server that runs on `localhost` only and must be explicitly enabled. 

Also create a job queue that routes the result/status of a request to the caller asynchronously. 

### Additional Information

_No response_

AL-CT:
**Summary**
Fixes #983. 
Creates a JWT Token authenticated route for requests that alter `TBC` state (e.g., `SyncIndexersToHash`). As these requests may be long-lived, we also create an `AdminHub` that manages the execution of these requests, and sends push notifications during its progress.

**Changes**
- Add `tbcadminapi.go` & `admin_rpc.go` for authenticated websocket connections.
- Add `admin_hub.go` for long-lived requests lifecycle management.
- Add `Msg` field to `Notifier` notifications.


api/tbcadminapi/tbcadminapi.go
```diff
@@ -0,0 +1,148 @@
+// Copyright (c) 2024-2025 Hemi Labs, Inc.
```
joshuasing:
```suggestion
// Copyright (c) 2026 Hemi Labs, Inc.
```

service/tbc/admin_rpc.go
```diff
@@ -0,0 +1,407 @@
+// Copyright (c) 2024-2025 Hemi Labs, Inc.
```
joshuasing:
```suggestion
// Copyright (c) 2026 Hemi Labs, Inc.
```

joshuasing (approved):

api/tbcadminapi/tbcadminapi.go
AL-CT: Oops, fixed 👍

service/tbc/admin_rpc.go
AL-CT: Fixed 👍

go.mod
```diff
@@ -59,6 +59,7 @@ require (
 	github.com/go-logr/logr v1.4.3 // indirect
 	github.com/go-logr/stdr v1.2.2 // indirect
 	github.com/go-ole/go-ole v1.3.0 // indirect
+	github.com/golang-jwt/jwt/v5 v5.3.1
```
joshuasing: I think `go mod tidy` is needed, it looks like this is in the indirect `require` block because it was not detected as in-use (presumably from `go get`)

joshuasing: Not sure why CI doesn't fail on that :thinking:

jcvernaleo (approved):
OK

go.mod
AL-CT: Oh, strange. Moved 👍

joshuasing (approved):

Issue #988, opened 2026-04-23
add full sync node that will need to derive from l1


ClaytonNorthey92:
add a full sync node that will need to derive the l2 chain from l1 upon startup.

PR #1005, opened 2026-04-30
e2e: full sync node and large bitcoin re-org+block deletion in localnet

### LINKED ISSUE #989: re-org btc in localnet test so bitcoin attributes deposited transaction are not on the canonical chain
after the network runs for a bit, say after a finalized block, reorg the bitcoin block chain back far enough so that bitcoin attributes deposited transactions written to the l2 are not on the canonical bitcoin chain.

do this before nodes that need to "sync" start, so that they are forced to find a way around this (via a request to other nodes for the block that was written to the tx)

ClaytonNorthey92:
**Summary**

* add a full sync node to localnet testing, this will wait for a finalized block before starting
* invalidate a block in the bitcoin chain then "prune" everything below tip, forcing nodes to get missing blocks from l2 peers

the invalidation and pruning functions like so:

* find the highest bitcoin attributes deposited transaction in the hemi chain
* use "invalidateblock" to re-org it off of the canonical chain
* "prune" all of the blocks below tip from the bitcoind node
* start the snap-sync and full-sync nodes, they _should_ have to fetch the block from the l2 since it was re-org'd off of the chain

fixes https://github.com/hemilabs/heminetwork/issues/989 and https://github.com/hemilabs/heminetwork/issues/988e2e: full sync node and large bitcoin re-org+block deletion in localnet
* add a full sync node to localnet testing, this will wait for a finalized block before starting
* invalidate a block in the bitcoin chain then "prune" everything below tip, forcing nodes to get missing blocks from l2 peers

the invalidation and pruning functions like so:

* find the highest bitcoin attributes deposited transaction in the hemi chain
* use "invalidateblock" to re-org it off of the canonical chain
* "prune" all of the blocks below tip from the bitcoind node
* start the snap-sync and full-sync nodes, they _should_ have to fetch the block from the l2 since it was re-org'd off of the chain

fixes https://github.com/hemilabs/heminetwork/issues/989 and https://github.com/hemilabs/heminetwork/issues/988

**Changes**
see summary

### DIFF
--- a/e2e/docker-compose.yml
+++ b/e2e/docker-compose.yml
@@ -287,6 +287,7 @@ services:
       - "--p2p.no-discovery"
       - "--p2p.priv.path=/tmp/op-node-priv-key.txt"
       - "--p2p.sequencer.key=${ADMIN_PRIVATE_KEY}"
+      - "--p2p.ban.peers=false"
       - "--log.level=debug"
       - "--log.format=terminal"
       - "--override.ecotone=1725868497"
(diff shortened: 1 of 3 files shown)


e2e/docker-compose.yml
```diff
@@ -65,7 +66,7 @@ services:
       - "-rpcport=18443"
       - "-rpcconnect=bitcoind"
       - "generatetoaddress"
-      - "1"
+      - "100"
```
ClaytonNorthey92: talking to @max-sanchez , this is here because we want to mine 3+ blocks after re-org. I figured it convenient to simply generate more blocks each time that we do.

ClaytonNorthey92:
hang on with review one moment

PR #971, opened 2026-04-17
wallet: segwit signing, PrevOuts refactor, TxByID, multi-form key indexing

marcopeereboom:
Add segwit (P2WPKH, P2TR key-path) signing support to the wallet
package, refactor PrevOuts to carry amounts required by BIP-143/BIP-341
witness sighash computation, and wire TxByID into the gozer interface.

**PrevOuts refactor**

TransactionCreate and TransactionSign now use `PrevOuts` (`map[string]*wire.TxOut`)
instead of `map[string][]byte`.  Witness sighash algorithms (BIP-143 for
segwit v0, BIP-341 for taproot) require the spent output's value in the
hash preimage.  The old type carried only the pkScript; callers had to
thread amounts out-of-band.  This is a breaking API change —
TransactionCreate returns PrevOuts and TransactionSign accepts it.

**Segwit signing**

TransactionSign now dispatches to signLegacy, signP2WPKH, or
signP2TRKeyPath based on the pkScript class of each input's previous
output.  P2PKH inputs produce SignatureScript; P2WPKH and P2TR inputs
produce Witness and clear SignatureScript (TransactionCreate
pre-populates it for fee estimation).

**Multi-form key indexing**

zuul/memory PutKey now derives P2PKH, P2WPKH, and P2TR addresses from
a single compressed public key and indexes the key under all three.
LookupKeyByAddr resolves the signing key regardless of which address
form the prevout uses.

**TxByID**

Add TxByID to the Gozer interface and tbcgozer implementation.  The
server-side TxByIdRequest/TxByIdResponse RPC already exists; this wires
it into the client.

**Tests**

- P2WPKH and P2TR end-to-end: build, sign, verify through script engine
- Mixed-input (legacy + segwit) signing
- PoP transaction regression tests
- TxByID positive, negative, mock, integration, and fuzz coverage
- SignatureScript clearing regression (pre-populated scriptSig path)
- Error-path coverage for utxo pickers, missing PrevOuts, unknown keys,
  malformed outpoint keys

PR #1018, opened 2026-05-11
wallet: TSS external signing + TSS key storage

marcopeereboom:
Add external ECDSA and Schnorr signature application for TSS-signed
Bitcoin transactions, TSS key storage in zuul, and signature
verification helpers.

Depends on #971  (segwit signing PR).

**TSS key type**

Add TSSNamedKey to zuul and PutTSSKey/LookupKeyByAddr to zuul/memory.
TSS keys store only the public key (the private key is distributed
across TSS participants).  Multi-form address indexing (P2PKH, P2WPKH,
P2TR) is shared with local keys; cross-type collisions are rejected
with distinct sentinels (ErrTSSKeyOccupied, ErrLocalKeyOccupied)
wrapping ErrKeyExists.

**External signature application**

TransactionApplyECDSA signs P2PKH and P2WPKH inputs given a DER-encoded
ECDSA signature and the signer's public key.  TransactionApplySchnorr
signs P2TR key-path inputs given a 64-byte Schnorr signature.  Both
verify pubkey-to-address binding before applying.

**Verification**

VerifyECDSA and VerifySchnorr validate signatures against the
transaction's sighash without requiring the private key — intended for
coordinators that receive signatures from remote TSS participants.

**Tests**

- TSS end-to-end: keygen → external sign → apply → verify → script
  engine for P2PKH, P2WPKH, and P2TR inputs
- Taproot witness crypto negative paths
- Coverage gap tests for address mismatch, non-default sighash types
- Cross-type key collision assertions

Blocked on hemilabs/x PR #9 (tss-lib v3 tag).


go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/ethereum/go-ethereum v1.17.2
 	github.com/go-test/deep v1.1.1
 	github.com/golang-jwt/jwt/v5 v5.3.1
+	github.com/hemilabs/x/tss-lib/v3 v3.0.0-20260507172513-c23bec7119b9
```
joshuasing: Must be changed to use `github.com/hemilabs/x/tss/v3` once merged.

bitcoin/wallet/verify.go
```diff
@@ -0,0 +1,85 @@
...
+// committees) before handing them to TransactionApplyECDSA.
+//
+// sigDER is DER-encoded without any trailing sighash byte.  sigHash
+// is the 32-byte hash the signer produced the signature over; the
+// caller is responsible for computing it with the correct sighash
+// algorithm (legacy for P2PKH, BIP-143 for P2WPKH).
+//
+// Returns nil on success.  On failure the error distinguishes
+// between parse errors and verification mismatches.
+func VerifyECDSA(sigHash, sigDER []byte, pubKey *btcec.PublicKey) error {
```
AL-CT, line 28: Is this meant to only verify signature created using `external_sign.go`? If not and want it to be able to verify other external signatures, maybe we can add a check here to see if we have a `high-s` and reject the signature.

bitcoin/wallet/external_sign.go
```diff
@@ -0,0 +1,214 @@
...
+	if tx == nil || prev == nil || pubKey == nil {
+		return errors.New("tx, prev and pubKey cannot be nil")
+	}
+	if idx < 0 || idx >= len(tx.TxIn) {
+		return fmt.Errorf("input index %d out of range (tx has %d inputs)",
+			idx, len(tx.TxIn))
+	}
+	if err := validateSigHashType(hashType); err != nil {
+		return err
+	}
```
AL-CT, line 113: Isn't `SigHashDefault` invalid for ECDSA? Might be good to add a check below

AL-CT (changes requested):
Two comments for consideration, otherwise looks good 👍

ClaytonNorthey92 (approved):

go.mod
marcopeereboom: Done — switched to `github.com/hemilabs/x/tss/v3` in commit 80b30cb.

bitcoin/wallet/verify.go
marcopeereboom: Good call — added a BIP-146 high-S rejection check. VerifyECDSA now serializes the parsed signature back to DER and compares against the input; if they differ the original carried a high-S value (or non-canonical DER) and is rejected with `ErrNonCanonicalSig`. Test in `TestVerifyECDSARejectsHighS`.

bitcoin/wallet/external_sign.go
marcopeereboom: You're right — `SigHashDefault` is taproot-only and invalid for ECDSA. Added an explicit rejection that returns `ErrSigHashDefaultECDSA` before the sighash validation. Test in `TestTransactionApplyECDSARejectsSigHashDefault`.

AL-CT (approved):
LGTM 👍

PR #1022, opened 2026-05-12
trust: add RPC module

### LINKED ISSUE #977: trust: add RPC module
### Confirmation

### Problem

`Trust` will require a way to route requests from `op-reth` calls to a `tbcd` instance, through RPC. 

### Suggested solution

Implement an RPC crate that allows the redirecting of requests from `trust` to `tbcd`.

### Additional Information

_No response_

AL-CT:
**Summary**
Fixes #977. Introduces an RPC client to trust that allows for communication with a TBC node. 

Calls are currently _serial_ to prevent failures on admin calls (e.g., SyncIndexersToHash) when called multiple times in parallel, though can easily be converted to parallel in the future, if required.

Several calls required by `op-reth` are still missing, but these need to be implemented in `TBC` and then ported to `Trust`, in a different PR.

**Changes**
- Add `trust_rpc` module.
- Add `protocol` sub-module.


ClaytonNorthey92 (approved):

crates/trust/src/trust_rpc/protocol.rs
```diff
@@ -0,0 +1,363 @@
...
+    let res = serde_json::to_string(msg)?;
+    Ok(res)
+}
+
+pub fn unmarshal(data: &str) -> super::Result<Message> {
+    let res = serde_json::from_str(data)?;
+    Ok(res)
+}
+
+#[cfg(test)]
```
joshuasing: Do we want to move these to an `_test.rs` file to match the other tests?

joshuasing (approved):

crates/trust/src/trust_rpc/protocol.rs
AL-CT: Done 👍

PR #1024, opened 2026-05-12
tbc: add ordinal indexer

marcopeereboom:
depends on #1035 

Add ordinal indexer to TBC that tracks individual satoshi ownership via sat ranges and indexes Bitcoin inscriptions. Follows the existing Indexer/indexer framework pattern, modeled on zkindexer.go.

Database:
- New LevelDB "ordinals" with prefix-separated keys: 'r' (sat ranges per UTXO), 'i' (inscription by ID), 's' (inscribed sat → outpoint), 'a' (sat → all inscriptions), 'n' (inscriptions by block).
- OrdinalKey type, 7 new Database interface methods.
- Database version bump 5 → 6.

Indexer:
- FIFO sat range redistribution engine.
- Inscription envelope parser supporting all tags: content type, pointer, parent, delegate, metaprotocol.
- Cursed inscription detection (5 pre-jubilee rules).
- Flag-driven variable-length value encoding.
- Inscribed sat movement tracking via DB range scan.
- Zero dependency on txindex during wind/unwind.


database/tbcd/level/level.go
```diff
@@ -2409,3 +2430,265 @@ func (l *ldb) BlockCacheStats() tbcd.CacheStats {
...
+func (l *ldb) OrdinalInscribedSatBounds(ctx context.Context) (uint64, uint64, error) {
+	log.Tracef("OrdinalInscribedSatBounds")
+	defer log.Tracef("OrdinalInscribedSatBounds exit")
+
+	ordDB := l.pool[level.OrdinalDB]
+
+	// Find first 's' entry.
+	var startKey tbcd.OrdinalKey
+	startKey[0] = 's'
+	it := ordDB.NewIterator(nil, nil)
```
AL-CT: I believe this function can be simplified by using `util.BytesPrefix(startKey[:])` and getting the first and last entries for the bounded iterator.

service/tbc/ordinalindex.go
```diff
@@ -0,0 +1,701 @@
...
+// Time: O(block_inputs + block_outputs + inscribed_sats_in_range)
+// Space: O(cache_entries_per_block)
+func (i *ordinalIndexer) windBlock(ctx context.Context, blockHeight uint32, blockHash *chainhash.Hash, block *btcutil.Block, cache map[tbcd.OrdinalKey]tbcd.OrdinalValue) error {
+	// One-time check on first call: if resuming from a prior run, the
+	// DB may contain inscribed sats. Two iterator seeks to get
+	// min/max bounds — O(1), no full scan.
+	if !i.inscCheckDone {
+		i.inscCheckDone = true
+		minSat, maxSat, err := i.g.db.OrdinalInscribedSatBounds(ctx)
+		if err == nil {
```
AL-CT: Should we check with `errors.Is(err, database.ErrNotFound)`?

service/tbc/ordinalindex.go
```diff
@@ -0,0 +1,701 @@
...
+			}
+			if len(dbSats) > 0 || len(i.inscribedSatSet) > 0 {
+				blockInscribedSats = make([]uint64, 0, len(dbSats)+len(i.inscribedSatSet))
+				blockInscribedSats = append(blockInscribedSats, dbSats...)
+				for sat := range i.inscribedSatSet {
+					blockInscribedSats = append(blockInscribedSats, sat)
+				}
+				sort.Slice(blockInscribedSats, func(a, b int) bool {
+					return blockInscribedSats[a] < blockInscribedSats[b]
+				})
```
AL-CT: I believe this can be simplified as
```suggestion
				slices.Sort(blockInscribedSats)
```

service/tbc/ordinalindex.go
```diff
@@ -0,0 +1,701 @@
...
+			perInputStart = append(perInputStart, 0)
+		}
+		allInputRanges = append(allInputRanges, ranges...)
+
+		// Mark spent input's sat ranges for deletion. Overwrite
+		// with nil (never delete map keys). Mirrors the utxo
+		// indexer's sentinel pattern: the value encodes intent,
+		// the key set is stable across the batch.
+		rk := ordinalRangeKey(op)
+		cache[rk] = nil
```
AL-CT: nit
```suggestion
		i.putRange(cache, outpoint, nil)
```

service/tbc/ordinalindex.go
```diff
@@ -0,0 +1,701 @@
...
+	key[0] = 'n'
+	copy(key[1:], blockHash[:])
+	binary.BigEndian.PutUint32(key[33:], seq)
+	return key
+}
+
+func makeInscriptionID(txHash *chainhash.Hash, inputIdx uint32) [36]byte {
+	var id [36]byte
+	copy(id[:32], txHash[:])
+	binary.LittleEndian.PutUint32(id[32:], inputIdx)
```
AL-CT, line 190: Is Little Endian here part of the ord protocol? If so, maybe a small comment here is good

service/tbc/ordinalindex.go
```diff
@@ -0,0 +1,701 @@
...
+			case <-ctx.Done():
+			case c <- struct{}{}:
+			}
+		}()
+	}
+
+	v, err := i.g.db.OrdinalSatRangesByOutpoint(ctx, op)
+	if err != nil {
+		// Created and spent in the same block.
+		return
```
AL-CT: Probably need to check if error is `NotFound`, otherwise we might have a real error and just silently ignore it and mess up the indexing.

service/tbc/ordinalindex.go
```diff
@@ -0,0 +1,701 @@
...
+		var minSat, maxSat uint64
+		minSat = math.MaxUint64
+		for _, tx := range txs {
+			if blockchain.IsCoinBase(tx) {
+				continue
+			}
+			for _, txIn := range tx.MsgTx().TxIn {
+				op := tbcd.NewOutpoint(txIn.PreviousOutPoint.Hash,
+					txIn.PreviousOutPoint.Index)
+				ranges, _ := i.satRanges(ctx, op, cache)
```
AL-CT: Is it okay to ignore the error here?

service/tbc/ordinalindex.go
```diff
@@ -0,0 +1,701 @@
...
+	// Inscribed fee sats are rare on mainnet (~0 per typical block).
+	// Triple-nested loop is justified: |allInscribedFeeSats| is
+	// typically 0, |outputRanges| ≤ coinbase outputs (~1-2),
+	// |ranges per output| is typically 1.
+	for _, sat := range allInscribedFeeSats {
+		for txOutIdx, outRanges := range outputRanges {
+			for _, sr := range outRanges {
+				if sat >= sr.Start && sat < sr.Start+sr.Count {
+					outpoint := tbcd.NewOutpoint(*coinbaseTx.Hash(), txOutIdx)
+					cache[ordinalSatKey(sat)] = outpoint[:]
```
AL-CT: Can't we `break` here?

service/tbc/ordinals.go
```diff
@@ -0,0 +1,490 @@
...
+// InscriptionEnvelope represents a parsed inscription from a taproot witness.
+type InscriptionEnvelope struct {
+	ContentType            []byte
+	Content                []byte
+	Pointer                *uint64   // tag 2: sat offset within outputs
+	Parent                 *[36]byte // tag 3: parent inscription ID
+	Metaprotocol           []byte    // tag 7: metaprotocol name
+	Delegate               *[36]byte // tag 11: delegate inscription ID
+	HasUnrecognizedEvenTag bool      // for cursed detection
+	MultipleEnvelopes      bool      // set if >1 envelope found in same input
```
AL-CT, line 153: I don't think this ever gets set to true in the code, not sure if we are indexing multiple envelopes in a script

service/tbc/ordinalindex_rpc_test.go
AL-CT, line 1: Can we move this to `rpc_test.go`? There's a few helper functions there that create local TBC instances with an http server and without using sleeps (`createLocalTBCServer`, which you can change to accept an existing home dir) and make calls. Test `TestRPCRequestsLocal` is a good example.

service/tbc/ordinalindex_design.go
AL-CT, line 1: Maybe we can move this to the benchmark test file? Or create a .md file like `zkindexers.md`

service/tbc/ordinalindex_bench_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+	txHash := chainhash.Hash{}
+	outputRanges := map[uint32][]SatRange{
+		0: {{Start: 0, Count: 500}},
+		1: {{Start: 500, Count: 750}},
+	}
+	cache := make(map[tbcd.OrdinalKey]tbcd.OrdinalValue)
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
AL-CT: I believe this and the other benchmark loops can be simplified as:
```suggestion
	for b.Loop() {
```

service/tbc/tbc.go
```diff
@@ -2855,10 +2959,22 @@ func (s *Server) synced(ctx context.Context) (si SyncInfo) {
...
+		ordHH := &HashHeight{
+			Hash:   ordBH.Hash,
+			Height: ordBH.Height,
+		}
+		si.Ordinal = *ordHH
+	}
+
 	if utxoHH.Hash.IsEqual(&bhb.Hash) && txHH.Hash.IsEqual(&bhb.Hash) &&
 		!s.indexing && !blksMissing {
 		// If keystone and zk indexers are disabled we are synced.
```
AL-CT: nit
```suggestion
		// If keystone, zk, and ordinals indexers are disabled we are synced.
```

database/tbcd/level/ordinal_test.go
AL-CT, line 1: I think it would be better if we could fit these tests in `level_test.go`, or at the very least `TestDbUpgradeV6`, since that's where the other upgrade tests are.

service/tbc/ordinalindex.go
```diff
@@ -0,0 +1,748 @@
...
+		cache[ordinalInscriptionKey(inscID)] = encodeInscriptionValue(
+			inscribedSat, blockHash, cursed, envelope)
+
+		for txOutIdx, outRanges := range outputRanges {
+			for _, sr := range outRanges {
+				if inscribedSat >= sr.Start && inscribedSat < sr.Start+sr.Count {
+					outpoint := tbcd.NewOutpoint(*tx.Hash(), txOutIdx)
+					cache[ordinalSatKey(inscribedSat)] = outpoint[:]
+					i.trackInscribedSat(inscribedSat)
+					goto satFound
```
AL-CT: Will this not skip any _new_ inscriptions on a sat that ends up going to fees? It looks like the `inscribedFeeSats` we return here and is used in `windBlock` to deal with this only verifies existing inscriptions, not new ones

AL-CT (changes requested):
A few nits and comments

database/tbcd/level/level.go
marcopeereboom: Done — simplified `OrdinalInscribedSatBounds` to use `util.BytesPrefix` with `First()`/`Last()` instead of manual seek/prev logic.

service/tbc/ordinalindex.go
marcopeereboom: Addressed — the sat range computation code that contained this path was removed in the ordinals2 refactor. Sat ranges are now computed on demand via backward chain walk, not stored at index time.

service/tbc/ordinalindex.go
marcopeereboom: Moot — `sort.Slice` on `blockInscribedSats` was removed with the sat range computation code in the ordinals2 refactor.

service/tbc/ordinalindex.go
marcopeereboom: Moot — `putRange` removed with the sat range computation code.

service/tbc/ordinalindex.go
marcopeereboom: Done — added comment: "Inscription IDs use little-endian index per the ord protocol."

service/tbc/ordinalindex.go
marcopeereboom: Addressed — the sat range lookup path that silently ignored errors was removed. The indexer now only stores inscriptions; sat numbers are computed on demand with proper error propagation.

service/tbc/ordinalindex.go
marcopeereboom: Addressed — same as above, code path removed in the indexer refactor.

service/tbc/ordinalindex.go
marcopeereboom: Addressed — code path removed in the indexer refactor.

service/tbc/ordinals.go
marcopeereboom: Removed — we no longer track multi-envelope state in the parser.

service/tbc/ordinalindex_rpc_test.go
marcopeereboom: Acknowledged — keeping it separate for now as the ordinal RPC tests have their own setup (seed data, ordinal DB). Will consolidate with `rpc_test.go` once the on-demand sat computation changes stabilize.

service/tbc/ordinalindex_design.go
marcopeereboom: Done — moved to `ordinalindex_design.md`.

service/tbc/ordinalindex_bench_test.go
marcopeereboom: Moot — benchmark file removed.

service/tbc/tbc.go
marcopeereboom: Done — comment updated.

database/tbcd/level/ordinal_test.go
marcopeereboom: Acknowledged — keeping ordinal DB tests in `ordinal_test.go` for now. Will move `TestDbUpgradeV6` to `level_test.go` with the other upgrade tests in a follow-up.

service/tbc/ordinalindex.go
marcopeereboom: Addressed — the entire precomputed sat range model was replaced. Inscriptions are stored at index time without sat numbers. Sat numbers are computed on demand by tracing backward through the spending chain to coinbase. Fee region sats are handled correctly by computing per-tx fee amounts at the coinbase level.

service/tbc/ordinalquery.go
```diff
@@ -0,0 +1,330 @@
...
+		if *t.Hash() == txid {
+			tx = t
+			break
+		}
+	}
+	if tx == nil {
+		return nil, fmt.Errorf("tx %v not in block %v", txid, blockHash)
+	}
+
+	if int(vout) >= len(tx.MsgTx().TxOut) {
```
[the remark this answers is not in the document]
marcopeereboom: Fixed — `ParseUint` removed. The on-demand query path uses typed parameters throughout.

database/tbcd/level/ordinal_test.go
marcopeereboom: Done — moved `TestDbUpgradeV6` to `level_test.go` alongside `TestDbUpgradeV5` and `TestDbUpgradeV4Errors`.

marcopeereboom:
Superseded by new PR based on marco/tx-offset (PR #1052) and marco/lazy-block (PR #1051). Branch renamed to marco_ordinals5.

PR #1025, opened 2026-05-12
all: enable revive linter

joshuasing:
**Summary**
Enable `revive` (https://github.com/mgechev/revive) linter in golangci-lint.

This linter contains a bunch of rules focusing on: code readability, maintainability, best practices, and correctness. It can be verbose (e.g. var-naming, unused) but can often catch very subtle bugs caused by the wrong variable being used, etc.

Some rules are based on recommendations from the Go team that are enforced in Go itself, e.g. from https://go.dev/wiki/CodeReviewComments

**Changes**
- Enable `revive` linter in `.golangci.yaml`
- Fix all issues from revive linter

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -31,8 +31,7 @@ linters:
     - "prealloc" # Finds slice declarations that could potentially be pre-allocated.
     - "predeclared" # Find code that shadows one of Go's predeclared identifiers.
     - "promlinter" # Check Prometheus metrics naming via promlint.
-    # TODO: enable revive - when bored, run 'golangci-lint run -Erevive' and fix issues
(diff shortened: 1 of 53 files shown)


marcopeereboom (changes requested):
I hate this but will allow it but not before continuum goes in (and maybe the ordinal PR). This is way too much churn for large PRs.

ClaytonNorthey92 (approved):

ClaytonNorthey92:
@joshuasing , I have no objections, but could we wait to merge until after the PR @marcopeereboom  [is getting in](https://github.com/hemilabs/heminetwork/pull/1025#pullrequestreview-4279331059)?

PR #1026, opened 2026-05-13
tbc: add missing rpc commands

### LINKED ISSUE #978: trust: implement missing tbcd full node calls
### Confirmation

### Problem

After the development of an RPC crate that allows for the communication between `trust` and `tbcd`, we need to implement the methods used by `op-geth` (and consequently, `op-reth`) in `trust`.

### Suggested solution

Find all `TBCFullNode` methods called in `op-geth` and, through RPC passthroughs to `tbcd`, implement them in `trust`.

### Additional Information

_No response_

AL-CT:
**Summary**
Fixes #978. Adds several missing RPC calls to the regular and authenticated `TBC` routes.

Also addresses two bugs:
1. In `ScriptHashAvailableToSpend` an `error` was returned instead of `false` when the outpoint wasn't found.
2. In `BlockInTxIndex` the check was done incorrectly, always returning an `error`.

These two calls are used in precompiles in `op-geth`, so I have created issues to be careful when updating TBC there: [here](https://github.com/hemilabs/op-geth/issues/85) and [here](https://github.com/hemilabs/op-geth/issues/86).

Also added a few utility testing functions that can allow for simplifying and improving testing of our RPC calls, and created an issue for it: #1027 

**Changes**

RPC calls added:
- `BlockHashByTxId`
- `BlockInTxIndex`
- `FullBlockAvailable`
- `BlockHeaderByHash`
- `BlockHeaderByHashRaw`
- `BlockHashByTxId`
- `ScriptHashAvailableToSpend`
- `SyncStatus`

Authenticated RPC calls added:
- `BlockHeadersInsert`

### DIFF
--- a/service/tbc/admin_rpc.go
+++ b/service/tbc/admin_rpc.go
@@ -9,14 +9,17 @@ import (
 	"errors"
 	"io"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
+	"github.com/btcsuite/btcd/wire"
 	"github.com/coder/websocket"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/golang-jwt/jwt/v5"
 
 	"github.com/hemilabs/heminetwork/v2/api/protocol"
+	"github.com/hemilabs/heminetwork/v2/api/tbcadminapi"
 	tapi "github.com/hemilabs/heminetwork/v2/api/tbcadminapi"
 	"github.com/hemilabs/heminetwork/v2/api/tbcapi"
 )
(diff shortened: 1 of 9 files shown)


ClaytonNorthey92 (approved):

service/tbc/admin_rpc.go
```diff
@@ -358,6 +364,70 @@ func (s *Server) handleJobStatusRequest(_ context.Context, req *tapi.JobStatusRe
 	}, nil
 }
 
+func (s *Server) handleBlockHeadersInsertRequest(ctx context.Context, req *tapi.BlockHeadersInsertRequest) (any, string, error) {
```
joshuasing: It looks like the second return value is always an empty string (`""`). Is it used?

service/tbc/admin_rpc_test.go
```diff
@@ -554,83 +491,200 @@ func TestAdminRPCCommands(t *testing.T) {
...
+			}
+		})
 	}
+}
+
+func TestAdminRPCSyncIndexersToHash(t *testing.T) {
+	ctx, cancel := context.WithTimeout(t.Context(), 20*time.Second)
+	defer func() {
+		cancel()
+	}()
```
joshuasing: I think this can be simplified to:
```suggestion
	defer cancel()
```

joshuasing (approved):

service/tbc/admin_rpc_test.go
AL-CT: Thanks! Fixed 👍

PR #1029, opened 2026-05-15
tbc: fix JWT auth when missing iat field

AL-CT:
**Summary**
Fixes a bug where JWT tokens for the authenticated `TBC` RPC route that were missing the optional `iat` field would cause a `panic`. 

**Changes**
- Check if the field is present before making sanity checks.
- Improve test coverage.


jcvernaleo (approved):
OK

joshuasing (approved):

PR #1030, opened 2026-05-15
popm: fix panic during prom polling with nil geth client

AL-CT:
**Summary**
Fixes a potential panic where `prometheus` will try and make `op-geth` calls before we set the client.

**Changes**
- Set `op-geth` client behind mutex.
- Check if client is nil when polling.
- Add `TestPromPollBeforeOpgeth`.


joshuasing (approved):

jcvernaleo (approved):
OK

PR #1031, opened 2026-05-15
gozer: fix abbreviated keystone serialization

AL-CT:
**Summary**
Minor PR to fix serialization of abbreviated keystones in `gozer`, which truncated `EPHash` to 8 bytes rather than the expected 12. This is not used in our code, but callers may find it useful, and this prevents a mismatch between `hemi` and `gozer` types.

Also remove a related field in the `keystone` type in `popm`, which was set and not used.

**Changes**
See above.


joshuasing (approved):

jcvernaleo (approved):
OK

PR #1036, opened 2026-05-19
tbc: correct Synced() boolean for disabled indexers

marcopeereboom:
The HemiIndex and ZKIndex checks in synced() used inverted logic: !(cfg && hash.IsEqual) evaluates to true when the index is disabled, causing Synced() to always return false. This triggers the sync loop to request headers from all peers in a tight loop at tip, pegging the CPU indefinitely.

Fix the condition to skip the check when the index is disabled:
  before: !(s.cfg.Index && hash.IsEqual(&best))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   witness-inclusive after a v5 upgrade plus resync
   ([#972](https://github.com/hemilabs/heminetwork/pull/972)).
 
+- Fix inverted conditions in `synced()` causing tbcd to incorrectly report
+  itself as not-synced when an optional indexer is enabled.
+  ([#1036](https://github.com/hemilabs/heminetwork/pull/1036))
+
 ## [v2.0.0]
 
 ### Breaking Changes
--- a/service/tbc/synced_test.go
+++ b/service/tbc/synced_test.go
@@ -1,0 +1,146 @@
+// Copyright (c) 2026 Hemi Labs, Inc.
+// Use of this source code is governed by the MIT License,
+// which can be found in the LICENSE file.
+
+package tbc
+
+import (
+	"context"
+	"testing"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+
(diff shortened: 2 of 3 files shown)


service/tbc/tbc.go
```diff
@@ -2862,11 +2862,11 @@ func (s *Server) synced(ctx context.Context) (si SyncInfo) {
 			si.Synced = true
 			return si
 		}
```
joshuasing: I think this is now redundant; if both are false below, `si.Synced` will be set to `true`.

marcopeereboom: you are right. fixed

CHANGELOG.md
```diff
@@ -79,6 +79,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   witness-inclusive after a v5 upgrade plus resync
   ([#972](https://github.com/hemilabs/heminetwork/pull/972)).
 
+- Fix inverted boolean in `Synced()` that caused tbcd to report not-synced
+  when optional indexers (HemiIndex, ZKIndex) were disabled, resulting in a
+  tight header-request loop burning 200%+ CPU at tip.
```
joshuasing: Missing pull request.

service/tbc/synced_test.go
```diff
@@ -0,0 +1,146 @@
...
+				ti: &stubIndexer{bh: tipBH},
+			}
+			if tt.keystone != nil {
+				s.ki = &stubIndexer{bh: tt.keystone}
+			}
+			if tt.zk != nil {
+				s.zki = &stubIndexer{bh: tt.zk}
+			}
+
+			si := s.synced(context.Background())
```
joshuasing: nit: use `t.Context()` instead of `context.Background()` or `context.TODO()`
```suggestion
			si := s.synced(t.Context())
```

joshuasing (approved):
Looks good, one nit

AL-CT (approved):

service/tbc/synced_test.go
marcopeereboom: ugh, fixed, sorry

PR #1049, opened 2026-05-27
makefile: refactor and clean up

joshuasing:
**Summary**
Refactor the `Makefile` to make it easier to use and maintain.

Remove `GOPKG`, which is not an environment variable used by Go and caused `/pkg/` in the project directory to be deleted when running `make clean`.

**Changes**
- Remove `GOPKG` variable
- Use `PROJECT_BIN` variable for project binaries instead of using `GOBIN` and then overriding it when installing tools with `go install`.
- Respect existing `GOBIN`/`GOCACHE` env vars, defaulting `GOBIN` to the user's GOPATH/bin
- Use `:=` for immediate assignment of static variables
- Move all `.PHONY` declarations next to their targets to help with maintainability and prevent them being missed for new targets.
- Split `lint` target into `fmt`, `lint`, and `lint-check`
- Add `cover` target to run tests and then open HTML coverage viewer
- Rename `synctest-test` target to `synctest` (previously missing `.PHONY`)
- Add `btctool` to the cmds list
- Pin `govulncheck` version with a Renovate comment
- Replace `rm -rf $(GOCACHE)` with `go clean -cache`
- Pass `-trimpath` and `-ldflags` to the top-level build target

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,59 +1,87 @@
-# Copyright (c) 2024-2025 Hemi Labs, Inc.
+# Copyright (c) 2024-2026 Hemi Labs, Inc.
 # Use of this source code is governed by the MIT License,
 # which can be found in the LICENSE file.
 
 # PROJECTPATH is the project root directory.
 # This Makefile is stored in the project root directory, so the directory is
 # retrieved by getting the directory of this Makefile.
-PROJECTPATH = $(abspath $(dir $(realpath $(firstword $(MAKEFILE_LIST)))))
+PROJECTPATH := $(abspath $(dir $(realpath $(firstword $(MAKEFILE_LIST)))))
 
-export GOBIN=$(PROJECTPATH)/bin
-export GOCACHE=$(PROJECTPATH)/.gocache
-export GOPKG=$(PROJECTPATH)/pkg
-
-GO_LDFLAGS=
+GO_LDFLAGS ?=
+export GOCACHE ?= $(PROJECTPATH)/.gocache
+export GOBIN ?= $(shell go env GOPATH)/bin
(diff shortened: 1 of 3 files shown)


marcopeereboom (approved):
gawdyas

jcvernaleo (approved):
OK

Makefile
```diff
@@ -1,92 +1,122 @@
...
+.PHONY: lint-check
+lint-check:
+	$(GOBIN)/golangci-lint fmt --diff ./...
+	$(GOBIN)/golangci-lint run ./...
+	$(GOBIN)/golicenser -tmpl="$$LICENSE_HEADER" -author="$(LICENSE_AUTHOR)" -year-mode=git-range ./...
 
-test:
-	go test -test.timeout=20m -coverprofile=$(PROJECTPATH)/coverage.out -covermode=atomic ./...
-
-synctest-test:
```
ClaytonNorthey92, line 85: I think renaming this is going to break localnet tests, let's update it in the localnet test workflow file

ClaytonNorthey92 (changes requested):

Makefile
ClaytonNorthey92: Actually, I'm wrong. Ignore me.

ClaytonNorthey92 (approved):

Makefile
joshuasing: You were correct, `synctest-test` was being run in the Go / Test job. I modified it and forgot to commit, oops. I have just pushed a commit with the necessary CI changes, including fixing this. Thank you!

jcvernaleo (approved):
OK

PR #1051, opened 2026-05-29
tbc: add lazy block reader for zero-copy per-tx access

marcopeereboom:
## Summary

Add a `lazyBlock` type that wraps raw block bytes from the block cache with lazy per-tx access. Nothing is parsed until explicitly requested — individual tx access parses only that tx from its raw byte range.

## Problem

`BlockByHash` calls `btcutil.NewBlockFromBytes` on every read, which eagerly deserializes the entire block into heap objects (`MsgBlock`, every `MsgTx`, every `TxIn`/`TxOut`, witness slices). For a 3 MB block with 5000 txs, callers that need one tx pay for all 5000. CPU profile shows 50%+ GC pressure from discarded deserialization objects.

## Solution

- `BlockRawByHash` added to DB interface — returns raw `[]byte` from block cache/LevelDB without parsing
- `lazyBlock` type in `service/tbc/lazyblock.go`:
  - `Hash()` — SHA256 of first 80 bytes, computed on demand
  - `TxCount()` / `TxHash(i)` — scan tx boundaries from raw bytes, compute txid via SHA256 (handles segwit witness flag correctly)
  - `FindTx(txid)` — iterate `TxHash(i)` until match
  - `TxOutputValues(i)` — parse only one tx's output values from raw bytes
  - `FullBlock()` — fallback to `btcutil.NewBlockFromBytes`

No btcutil fork. External to btcutil entirely. Existing `BlockByHash` callers unchanged.

## Testing

- 1217 lines of tests covering: boundary scanning, txid computation (legacy + segwit + witness), output value extraction, FindTx, edge cases (empty blocks, coinbase-only), full-block fallback
- Tested against known mainnet and testnet blocks

## Files changed

- `database/tbcd/database.go` — add `BlockRawByHash` to interface
- `database/tbcd/level/level.go` — implement `BlockRawByHash`
- `service/tbc/lazyblock.go` — new file
- `service/tbc/lazyblock_test.go` — new file
- `service/tbc/cpfp_test.go` — stub updated

### DIFF
--- a/database/tbcd/level/level.go
+++ b/database/tbcd/level/level.go
@@ -1589,6 +1589,23 @@ func (l *ldb) BlockByHash(ctx context.Context, hash chainhash.Hash) (*btcutil.Bl
 	return b, nil
 }
 
+// BlockRawByHash returns the raw block bytes. Not cached, unlike BlockByHash.
+func (l *ldb) BlockRawByHash(ctx context.Context, hash chainhash.Hash) ([]byte, error) {
+	log.Tracef("BlockRawByHash")
+	defer log.Tracef("BlockRawByHash exit")
+
+	bDB := l.rawPool[level.BlocksDB]
+	eb, err := bDB.Get(hash[:])
+	if err != nil {
+		if errors.Is(err, leveldb.ErrNotFound) {
+			return nil, database.BlockNotFoundError{Hash: hash}
+		}
+		return nil, fmt.Errorf("block raw get: %w", err)
+	}
+
+	return eb, nil
+}
+
 func (l *ldb) BlockExistsByHash(ctx context.Context, hash chainhash.Hash) (bool, error) {
 	log.Tracef("BlockExistsByHash")
 	defer log.Tracef("BlockExistsByHash exit")
--- a/service/tbc/lazyblock.go
+++ b/service/tbc/lazyblock.go
@@ -1,0 +1,472 @@
+// Copyright (c) 2024-2026 Hemi Labs, Inc.
+// Use of this source code is governed by the MIT License,
+// which can be found in the LICENSE file.
+
+package tbc
+
+import (
+	"crypto/sha256"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"sync"
+
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+)
+
+// lazyBlock wraps raw block bytes from the block cache. Nothing is
+// parsed until explicitly requested. Individual tx access parses only
+// that tx from its raw byte range.
+type lazyBlock struct {
+	raw []byte // from block cache, never copied
+
+	mu           sync.Mutex
+	hash         chainhash.Hash // lazy, computed from raw[0:80]
+	hashComputed bool
+	txOffsets    []wire.TxLoc // lazy, populated on first tx access
+	txWitness    []bool       // parallel to txOffsets: true if tx has witness
+}
+
+// newLazyBlock wraps raw block bytes for lazy access. The raw slice
+// must be a complete serialized Bitcoin block (header + transactions).
+// No data is parsed until a method is called.
+//
+// The caller must guarantee that raw is not mutated, reused, or recycled
+// for the lifetime of the returned lazyBlock. raw bytes are referenced,
+// not copied. If raw is a sub-slice of a larger allocation, holding the
+// lazyBlock pins the entire backing array.
+func newLazyBlock(raw []byte) (*lazyBlock, error) {
+	if len(raw) < wire.MaxBlockHeaderPayload+1 {
+		return nil, fmt.Errorf("lazyBlock: raw too short (%d bytes)", len(raw))
+	}
+	return &lazyBlock{raw: raw}, nil
+}
+
+// Hash returns the block hash (double-SHA256 of the 80-byte header).
+// Computed once and cached.
+func (lb *lazyBlock) Hash() (chainhash.Hash, error) {
+	lb.mu.Lock()
+	defer lb.mu.Unlock()
+	if !lb.hashComputed {
+		lb.hash = chainhash.DoubleHashH(lb.raw[:wire.MaxBlockHeaderPayload])
+		lb.hashComputed = true
+	}
+	return lb.hash, nil
+}
+
+// ensureTxOffsets populates txOffsets on first call. Must hold lb.mu.
+func (lb *lazyBlock) ensureTxOffsets() error {
+	if lb.txOffsets != nil {
+		return nil
+	}
+	var err error
+	lb.txOffsets, lb.txWitness, err = scanTxBoundaries(lb.raw)
+	return err
+}
+
+// TxCount returns the number of transactions in the block.
+func (lb *lazyBlock) TxCount() (int, error) {
+	lb.mu.Lock()
+	defer lb.mu.Unlock()
+	if err := lb.ensureTxOffsets(); err != nil {
+		return 0, err
+	}
+	return len(lb.txOffsets), nil
+}
+
+// TxHash returns the non-witness txid for transaction i. The txid is
+// the double-SHA256 of the non-witness serialization: version + inputs +
+// outputs + locktime (excluding the segwit marker/flag and witness data).
+func (lb *lazyBlock) TxHash(i int) (chainhash.Hash, error) {
+	lb.mu.Lock()
+	defer lb.mu.Unlock()
+	if err := lb.ensureTxOffsets(); err != nil {
+		return chainhash.Hash{}, err
+	}
+	if i < 0 || i >= len(lb.txOffsets) {
+		return chainhash.Hash{}, fmt.Errorf("lazyBlock: tx index %d out of range [0, %d)", i, len(lb.txOffsets))
+	}
+	loc := lb.txOffsets[i]
+	txBytes := lb.raw[loc.TxStart : loc.TxStart+loc.TxLen]
+	return computeTxID(txBytes, lb.txWitness[i])
+}
+
+// FindTx iterates transactions computing txids until a match is found.
+// Returns the tx index or an error if not found.
+func (lb *lazyBlock) FindTx(txid chainhash.Hash) (int, error) {
+	lb.mu.Lock()
+	defer lb.mu.Unlock()
+	if err := lb.ensureTxOffsets(); err != nil {
+		return -1, err
+	}
+	for i := range lb.txOffsets {
+		loc := lb.txOffsets[i]
+		txBytes := lb.raw[loc.TxStart : loc.TxStart+loc.TxLen]
+		h, err := computeTxID(txBytes, lb.txWitness[i])
+		if err != nil {
+			return -1, fmt.Errorf("lazyBlock: tx %d: %w", i, err)
+		}
+		if h == txid {
+			return i, nil
+		}
+	}
+	return -1, fmt.Errorf("lazyBlock: tx %v not found", txid)
+}
+
+// TxOutputValues parses only the output section of transaction i and
+// returns the value (in satoshis) of each output. The only heap allocation
+// is the result slice.
+func (lb *lazyBlock) TxOutputValues(i int) ([]uint64, error) {
+	lb.mu.Lock()
+	defer lb.mu.Unlock()
+	if err := lb.ensureTxOffsets(); err != nil {
+		return nil, err
+	}
+	if i < 0 || i >= len(lb.txOffsets) {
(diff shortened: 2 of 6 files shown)


database/tbcd/level/level.go
```diff
@@ -1589,6 +1589,37 @@ func (l *ldb) BlockByHash(ctx context.Context, hash chainhash.Hash) (*btcutil.Bl
...
+				return nil, database.BlockNotFoundError{Hash: hash}
+			}
+			return nil, fmt.Errorf("block raw get: %w", err)
+		}
+		if l.cfg.blockCacheSize > 0 {
+			l.blockCache.Put(hash, eb)
+		}
+	}
+
+	return eb, nil
```
joshuasing: nit: I think this could be made more tidy by making the first cache check a quick-return instead of always falling through to a nil check. e.g.

```suggestion
	// get from cache
	if l.cfg.blockCacheSize > 0 {
		if eb, _ := l.blockCache.Get(hash); eb != nil {
			return eb, nil
		}
	}

	// get from db
	bDB := l.rawPool[level.BlocksDB]
	eb, err := bDB.Get(hash[:])
	if err != nil {
		if errors.Is(err, leveldb.ErrNotFound) {
			return nil, database.BlockNotFoundError{Hash: hash}
		}
		return nil, fmt.Errorf("block raw get: %w", err)
	}
	if l.cfg.blockCacheSize > 0 {
		l.blockCache.Put(hash, eb)
	}

	return eb, nil
```

service/tbc/lazyblock.go
```diff
@@ -0,0 +1,426 @@
...
+	hash      *chainhash.Hash // lazy, computed from raw[0:80]
+	txOffsets []wire.TxLoc    // lazy, populated on first tx access
+	txWitness []bool          // parallel to txOffsets: true if tx has witness
+}
+
+// newLazyBlock wraps raw block bytes for lazy access. The raw slice
+// must be a complete serialized Bitcoin block (header + transactions).
+// No data is parsed until a method is called.
+func newLazyBlock(raw []byte) *lazyBlock {
+	return &lazyBlock{raw: raw}
```
joshuasing: Consider validating the length of `raw` once here, to prevent a lazyBlock from being created wrapping obviously incorrect raw bytes. This also removes the need to check the length in receivers.

Related: https://github.com/hemilabs/heminetwork/pull/1051/changes#r3323667033

service/tbc/lazyblock.go
```diff
@@ -0,0 +1,426 @@
...
+			}
+			offset += 36
+
+			// Script length + script.
+			scriptLen, n, err := readVarInt(raw, offset)
+			if err != nil {
+				return nil, nil, fmt.Errorf("scanTxBoundaries: tx %d input %d: script len: %w", t, i, err)
+			}
+			offset += n
+			if offset+int(scriptLen) > len(raw) {
```
joshuasing: It is unsafe to cast `scriptLen` to an `int`, as `readVarInt` explicitly parses untrusted bytes as a `uint64` and returns it as such.

With a valid block, it would not be possible to trigger a panic - but it is still possible on invalid or corrupted raw inputs. I think it would be better to handle the value correctly than make assumptions.

Same on lines 231, 261, and 264.

```suggestion
			if scriptLen > uint64(len(raw) - offset) {
```

Related: https://github.com/hemilabs/heminetwork/pull/1051/changes#r3323667033

service/tbc/lazyblock.go
```diff
@@ -0,0 +1,426 @@
...
+
+	// Read transaction count.
+	txCount, n, err := readVarInt(raw, offset)
+	if err != nil {
+		return nil, nil, fmt.Errorf("scanTxBoundaries: tx count: %w", err)
+	}
+	offset += n
+
+	locs := make([]wire.TxLoc, 0, txCount)
+	witness := make([]bool, 0, txCount)
```
joshuasing, line 197: While not possible with valid Bitcoin block, `txCount` is a `uint64` parsed from untrusted bytes. This could potentially allow unbounded allocation of these slices, up-to triggering a panic.

Consider validating `txCount` before allocating.

Related: https://github.com/hemilabs/heminetwork/pull/1051/changes#r3323667033

service/tbc/lazyblock.go
```diff
@@ -0,0 +1,426 @@
...
+	}
+
+	// Read output count.
+	outputCount, n, err := readVarInt(txBytes, offset)
+	if err != nil {
+		return nil, fmt.Errorf("extractOutputValues: output count: %w", err)
+	}
+	offset += n
+
+	values := make([]uint64, outputCount)
```
joshuasing, line 451: While not exploitable with a valid Bitcoin block, `outputCount` is a `uint64` parsed from untrusted bytes. This could potentially allow unbounded allocation of these slices, up-to triggering a panic.

Consider validating `outputCount` before allocating.

Related: https://github.com/hemilabs/heminetwork/pull/1051/changes#r3323667033

service/tbc/lazyblock.go
```diff
@@ -0,0 +1,426 @@
...
+	if lb.txOffsets != nil {
+		return nil
+	}
+	offsets, witness, err := scanTxBoundaries(lb.raw)
+	if err != nil {
+		return err
+	}
+	lb.txOffsets = offsets
+	lb.txWitness = witness
+	return nil
```
joshuasing:
```suggestion
	var err error
	lb.txOffsets, lb.txWitness, err = scanTxBoundaries(lb.raw)
	return err
```

service/tbc/lazyblock.go
```diff
@@ -0,0 +1,426 @@
...
+
+	mu        sync.Mutex
+	hash      *chainhash.Hash // lazy, computed from raw[0:80]
+	txOffsets []wire.TxLoc    // lazy, populated on first tx access
+	txWitness []bool          // parallel to txOffsets: true if tx has witness
+}
+
+// newLazyBlock wraps raw block bytes for lazy access. The raw slice
+// must be a complete serialized Bitcoin block (header + transactions).
+// No data is parsed until a method is called.
```
joshuasing, line 34: Consider documenting expectations for `raw` input here.

The correctness of `lazyBlock` is entirely dependant on `raw` being guaranteed as immutable and kept alive for the full lifetime of every `lazyBlock`. If the underlying `raw` byte slice is ever reused or recycled, the entire lazyBlock would be silently corrupted and later cause errors, panics or incorrect data to be returned.

Additionally, if `raw` is a sub-slice of a larger byte slice, holding it pins the entire parent's backing array, not just the block's own `len(raw)` bytes. The `raw` slice would have to be copied into a new slice to prevent this.

This also assumes `raw` is always a valid Bitcoin block, which is not guaranteed. If created with incorrect or corrupted data, receivers below can error, panic or return invalid data. - If all instances of a lazyBlock being created are guaranteed to only input valid data, this is probably okay; if not, the scanning needs hardening.

```suggestion
// No data is parsed until a method is called.
// The caller must guarantee that raw is not mutated, reused, or recycled
// for the lifetime of the returned lazyBlock. raw bytes are referenced,
// not copied. raw must be a complete, valid serialized Bitcoin block.
```

joshuasing (changes requested):

service/tbc/lazyblock.go
```diff
@@ -0,0 +1,479 @@
...
+	"fmt"
+	"sync"
+
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+)
+
+// blockHeaderLen is the fixed size of a Bitcoin block header.
+const blockHeaderLen = 80
```
AL-CT: Not sure if it's worth it, but we can use the `wire` constant here
```suggestion
const blockHeaderLen = wire.MaxBlockHeaderPayload
```

service/tbc/lazyblock.go
```diff
@@ -0,0 +1,479 @@
...
+// The caller must guarantee that raw is not mutated, reused, or recycled
+// for the lifetime of the returned lazyBlock. raw bytes are referenced,
+// not copied. If raw is a sub-slice of a larger allocation, holding the
+// lazyBlock pins the entire backing array. raw must be a complete, valid
+// serialized Bitcoin block.
+type lazyBlock struct {
+	raw []byte // from block cache, never copied
+
+	mu        sync.Mutex
+	hash      *chainhash.Hash // lazy, computed from raw[0:80]
```
AL-CT: Is `Hash()` something we're going to need to call for most lazyBlocks? If so, I think making this a non-pointer field (with an added boolean for whether hash has been calculated yet, though I think an all 0s hash should also be okay) will keep the bytes per operation about the same, but we prevent a heap allocation

service/tbc/lazyblock.go
```diff
@@ -0,0 +1,479 @@
...
+
+	if txCount > maxBlockTxCount {
+		return nil, nil, fmt.Errorf("scanTxBoundaries: tx count %d exceeds max %d",
+			txCount, maxBlockTxCount)
+	}
+
+	locs := make([]wire.TxLoc, 0, txCount)
+	witness := make([]bool, 0, txCount)
+
+	for t := uint64(0); t < txCount; t++ {
```
AL-CT: nit (also applies to loops below)
```suggestion
	for t := range txCount {
```

AL-CT (changes requested):

database/tbcd/level/level.go
marcopeereboom: Done. Removed the cache from `BlockRawByHash` entirely — it's now a straight LevelDB get. No cache check, no fall-through. Cleaner than the early-return refactor.

service/tbc/lazyblock.go
marcopeereboom: Done. `newLazyBlock` now returns `(*lazyBlock, error)` and validates `len(raw) < wire.MaxBlockHeaderPayload+1` upfront.

service/tbc/lazyblock.go
marcopeereboom: Done. All locations now compare as `uint64` before casting: `scriptLen > uint64(len(raw)-offset)`. Applies to all four sites you flagged plus the equivalent ones in `computeTxID` and `extractOutputValues`.

service/tbc/lazyblock.go
marcopeereboom: Done. Added `maxBlockTxCount = 100_000` sanity cap — checked before `make()`. Any parsed `txCount` exceeding this returns an error.

service/tbc/lazyblock.go
marcopeereboom: Done. Both `scanTxBoundaries` and `extractOutputValues` now check `outputCount > uint64(len(raw)-offset)` before allocating.

service/tbc/lazyblock.go
marcopeereboom: Done.

service/tbc/lazyblock.go
marcopeereboom: Done. Added full doc comment covering immutability, lifetime, backing array pinning, and validity requirements. Also hardened all parsing paths per your related comments (constructor validation, bounded allocations, safe uint64 comparisons).

service/tbc/lazyblock.go
marcopeereboom: Done. Replaced with `wire.MaxBlockHeaderPayload` everywhere, removed the local `blockHeaderLen` constant.

service/tbc/lazyblock.go
marcopeereboom: Good call. Changed to a non-pointer `chainhash.Hash` field with a `hashComputed bool` guard. Avoids the heap allocation.

service/tbc/lazyblock.go
marcopeereboom: Done. All loops converted to `for t := range txCount` style.

AL-CT (approved):
LGTM 👍

service/tbc/lazyblock.go
```diff
@@ -0,0 +1,473 @@
...
+
+// lazyBlock wraps raw block bytes from the block cache. Nothing is
+// parsed until explicitly requested. Individual tx access parses only
+// that tx from its raw byte range.
+//
+// The caller must guarantee that raw is not mutated, reused, or recycled
+// for the lifetime of the returned lazyBlock. raw bytes are referenced,
+// not copied. If raw is a sub-slice of a larger allocation, holding the
+// lazyBlock pins the entire backing array. raw must be a complete, valid
+// serialized Bitcoin block.
```
joshuasing: I think this comment would make more sense on `newLazyBlock` - especially with its current wording.

joshuasing (changes requested):

ClaytonNorthey92 (approved):

service/tbc/lazyblock.go
marcopeereboom: Done. Moved the raw ownership/lifetime docs from the struct comment to `newLazyBlock`.

joshuasing (approved):

Issue #1050, opened 2026-05-28
tbc: tx index intermittently loses entries during IBD


marcopeereboom:
## Summary

The tx index (`TransactionsDB`) intermittently loses 't' entries during IBD. The ordinal indexer is the first consumer that reads every tx index entry during wind and surfaces this as a crash.

## Evidence

Three different txids missing across three independent fresh-sync runs on testnet4:

1. **`f621f096...`** at block 65569 — "tx not in block" (tx index pointed to wrong block, later resolved itself)
2. **`da220b73...`** at block 81528 — "tx not found" (0 entries in raw LevelDB prefix scan, 1 missing out of 413 txs in the block, 412 are present)
3. **`2a2b5ede...`** at block 114125 — "tx not found" (entry missing despite all verify checks passing)

## Investigation performed

- **Synthetic `processTxs` simulation:** Replayed block 81528's 413 txs through `processTxs` logic. All 413 entries correctly added to the Go map cache, including the missing txid. No skip, no collision, no duplicate.
- **Key collision check:** No `TxKey` collision between `da220b73...` and any other entry. 't' prefix (65 bytes) and 's' prefix (69 bytes) cannot collide.
- **Duplicate txid check:** No duplicate txids within block 81528.
- **Reorg check:** Single block header at each affected height. No reorg evidence. Fresh sync.
- **Post-commit verify:** Added a verify step that reads back every 't' txid immediately after `BlockTxUpdate` commits. The verify PASSES — the entry is readable right after commit. But the entry is missing when the ordinal indexer reads it later (minutes to hours after).
- **Slice vs copy test:** Replaced `key = k[0:65]` slice aliasing in `BlockTxUpdate` with explicit `make + copy`. The original `da220b73...` entry survived with copies, but a DIFFERENT entry (`2a2b5ede...`) went missing on the same run. The copy fix is inconclusive.
- **Wide-range audit:** Scanned blocks 60000-82000 via direct LevelDB read. Found exactly 1 missing tx out of ~22000 blocks.
- **Cross-indexer dependency:** The ordinal indexer is the only indexer that reads `BlockHashByTxId` during wind. The UTXO indexer reads its own `OutputsDB` via `ScriptHashByOutpoint` during wind. The tx index data loss may have existed unnoticed before the ordinal indexer was added.

## Key observation

[...]

PR #1052, opened 2026-05-29
tbc: store tx byte location in tx index, bump DB to v6

marcopeereboom:
## Summary

Store tx byte location (`TxLoc`: offset + length within raw block) in the tx index `'t'` entry value, which was previously nil. This enables O(1) tx lookup by jumping directly to the tx's bytes in the raw block — no scanning, no SHA256 hashing, no full block deserialization.

**Depends on #1051** (lazy block reader).

## Problem

Every tx lookup via `BlockHashByTxId` requires a subsequent full block deserialization to find the tx. The `'t'` entry value was nil — wasted space that could carry the byte offset. CPU profile shows 60% in SHA256 hashing from `FindTx` scanning every tx in the block to find a match.

## Solution

- `BlockHashByTxId` signature changed to return `(*chainhash.Hash, wire.TxLoc, error)` — all callers updated
- `processTxs` now calls `block.TxLoc()` and stores offset+length via `NewTxMappingWithLoc`
- `BlockTxUpdate` uses stack-allocated reusable buffers instead of slicing loop variables (addresses potential data integrity issue documented in #1050)
- All consumers wired to use TxLoc when available, with legacy fallback for nil values:
  - `TxById` (RPC) — deserializes only the target tx
  - `txOutFromOutPoint` (UTXO unwind) — deserializes only the target tx
  - `handleBlockHashByTxIdRequest` (RPC) — hash only, ignores TxLoc
  - `hemictl` — hash only, ignores TxLoc
- DB version 5 → 6, upgrade wipes tx index for rebuild with TxLoc values
- Errors from `block.TxLoc()` logged at Errorf, falls back to nil values

## Testing

- `TestDbUpgradeV6` — seeds v5 DB, runs upgrade, verifies index wiped and version bumped
- `TestTxLocRoundTrip` — stores TxLoc, reads back, verifies offset+length match
- `TestTxLocOffsetCorrectness` — stores raw block, uses offset to extract tx bytes, deserializes, verifies txid and output values match
- `TestTxLocLegacyNilValue` — nil-value entry returns zero TxLoc gracefully

## Impact

With TxLoc, each cache miss costs: 1 LevelDB read (tx index) + 1 LevelDB/cache read (raw block) + parse ~200 bytes of one tx. No full block deserialization. No SHA256 scanning. Eliminates the need for parallel lookup strategies.

## Related

- #1050 — tx index intermittently loses entries during IBD (the stack buffer fix in this PR addresses the potential root cause)
- #1051 — lazy block reader (dependency)

## Files changed

- `database/tbcd/database.go` — `BlockHashByTxId` returns TxLoc, `NewTxMappingWithLoc`
- `database/tbcd/level/level.go` — implementation, stack buffers, DB v6
- `database/tbcd/level/level_test.go` — 4 new tests
- `database/tbcd/level/upgrade.go` — `v6()` wipes tx index
- `service/tbc/txindex.go` — stores TxLoc
- `service/tbc/tbc.go` — `TxById` uses TxLoc
- `service/tbc/utxoindex.go` — `txOutFromOutPoint` uses TxLoc
- `service/tbc/rpc.go` — caller updated
- `service/tbc/cpfp_test.go` — stub updated
- `service/tbc/tbc_test.go` — version expectations updated
- `cmd/hemictl/hemictl.go` — caller updated

### DIFF
--- a/database/tbcd/database.go
+++ b/database/tbcd/database.go
@@ -449,7 +449,7 @@ func NewScriptHashFromString(hash string) (ScriptHash, error) {
 //
 // Transaction ID to Block mapping:
 //
-//	t + txid + blockhash = nil | [1 + 32 + 32] = nil
+//	t + txid + blockhash = txstart + txlen | [1 + 32 + 32] = [4 + 4]
 type (
 	TxKey   [69]byte // Allocate max sized key, the prefix byte determines the lengths
 	TxValue [36]byte // allocate max sized value
--- a/database/tbcd/database.go
+++ b/database/tbcd/database.go
@@ -121,7 +121,7 @@ type Database interface {
 	BlockHeaderByTxIndex(ctx context.Context) (*BlockHeader, error)
 	BlockUtxoUpdate(ctx context.Context, direction int, utxos map[Outpoint]CacheOutput, utxoIndexHash chainhash.Hash) error
 	BlockTxUpdate(ctx context.Context, direction int, txs map[TxKey]*TxValue, txIndexHash chainhash.Hash) error
-	BlockHashByTxId(ctx context.Context, txId chainhash.Hash) (*chainhash.Hash, error)
+	BlockHashByTxId(ctx context.Context, txId chainhash.Hash) (*chainhash.Hash, *wire.TxLoc, error)
 	SpentOutputsByTxId(ctx context.Context, txId chainhash.Hash) ([]SpentInfo, error)
 	// ScriptHash returns the sha256 of PkScript for the provided outpoint.
 	BalanceByScriptHash(ctx context.Context, sh ScriptHash) (uint64, error)
@@ -479,6 +479,20 @@ func NewTxMapping(txId, blockHash *chainhash.Hash) (txKey TxKey) {
 	copy(txKey[33:], blockHash[:])
 
 	return txKey
+}
+
+// NewTxMappingWithLoc returns a TxKey and TxValue that maps a tx id to a block
+// hash with the tx byte location (offset + length) within the raw block. This
+// allows callers to jump directly to the tx's bytes without scanning the block.
+func NewTxMappingWithLoc(txId, blockHash *chainhash.Hash, loc wire.TxLoc) (txKey TxKey, txValue TxValue) {
+	txKey[0] = 't'
+	copy(txKey[1:33], txId[:])
+	copy(txKey[33:], blockHash[:])
+
+	binary.BigEndian.PutUint32(txValue[0:4], uint32(loc.TxStart))
+	binary.BigEndian.PutUint32(txValue[4:8], uint32(loc.TxLen))
+
+	return txKey, txValue
 }
 
 func TxIdBlockHashFromTxKey(txKey TxKey) (*chainhash.Hash, *chainhash.Hash, error) {
--- a/database/tbcd/level/level.go
+++ b/database/tbcd/level/level.go
@@ -48,7 +48,7 @@ import (
 //	UTXOs
 
 const (
-	ldbVersion = 5
+	ldbVersion = 6
 
 	logLevel = "INFO"
 	verbose  = false
@@ -312,6 +312,10 @@ func New(ctx context.Context, cfg *Config) (*ldb, error) {
 			// Upgrade to v5: wipe witness-stripped block
 			// bodies and rebuild blocksmissing from headers.
 			err = l.v5(ctx)
+		case 5:
+			// Upgrade to v6: wipe tx index so it rebuilds
+			// with TxLoc values in 't' entries.
+			err = l.v6(ctx)
 		default:
 			if ldbVersion == dbVersion {
 				if Welcome {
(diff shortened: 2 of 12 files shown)


database/tbcd/level/upgrade.go
```diff
@@ -750,3 +750,42 @@ func (l *ldb) v5(ctx context.Context) error {
 	binary.BigEndian.PutUint64(v, 5)
 	return l.MetadataPut(ctx, versionKey, v)
 }
+
+func (l *ldb) v6(ctx context.Context) error {
```
joshuasing, line 754: How long does this upgrade take?

marcopeereboom: On my potato digital ocean vm, 28h. But I have to stress that it really is a potato that takes 30s per flush. My guess on real hardware is about 12h.

marcopeereboom:
Don't merge yet!

AL-CT (changes requested):
Blocking

database/tbcd/level/level.go
```diff
@@ -1628,36 +1632,38 @@ func (l *ldb) BlockExistsByHash(ctx context.Context, hash chainhash.Hash) (bool,
...
+
+	var found bool
+	var blockHash chainhash.Hash
+	var loc wire.TxLoc
 	for it.Next() {
-		block, err := chainhash.NewHash(it.Key()[33:])
-		if err != nil {
-			return nil, err
+		if found {
+			panic(fmt.Sprintf("multiple blocks for tx %v", txId))
```
ClaytonNorthey92, line 1651: why are we panicking here? as opposed to returning an error

database/tbcd/level/level.go
```diff
@@ -1628,36 +1632,38 @@ func (l *ldb) BlockExistsByHash(ctx context.Context, hash chainhash.Hash) (bool,
 	return ok, nil
 }
 
-func (l *ldb) BlockHashByTxId(ctx context.Context, txId chainhash.Hash) (*chainhash.Hash, error) {
+func (l *ldb) BlockHashByTxId(ctx context.Context, txId chainhash.Hash) (*chainhash.Hash, wire.TxLoc, error) {
```
ClaytonNorthey92: a bit subjective here, but I think it makes sense: could we return `*wire.TxLoc` (a pointer), so we can return `nil` for empty instead of an empty value?

database/tbcd/level/level.go
ClaytonNorthey92: also, what if the txid is found on two tips? how do we handle that? two competing valid blocks?

ClaytonNorthey92 (changes requested):

database/tbcd/level/level.go
marcopeereboom: The panic is intentional — a duplicate txid in the index means data corruption (the invariant is one canonical entry per txid). Returning an error would let the caller continue with a corrupt index. This is a hard stop, same pattern as the existing assert panics elsewhere in the indexer.

database/tbcd/level/level.go
marcopeereboom: Good call, will change to `*wire.TxLoc` — nil for not found is cleaner than a zero value.

database/tbcd/level/level.go
marcopeereboom: Only the canonical chain's tx index is maintained. During a reorg the old chain's entries are deleted and the new chain's are written. There's never two entries for the same txid — one is always canonical. The panic catches that invariant violation.

ClaytonNorthey92: ok got it 👍

ClaytonNorthey92 (approved):

database/tbcd/database.go
AL-CT: Can we change this comment to reflect the new value with loc data?

AL-CT, line 455: The `locator` value (8 bytes) is so much smaller than the `spent` value (36 bytes), would it not make sense to create a different kind of TxValue?

AL-CT (changes requested):
Two small comments for consideration

database/tbcd/database.go
marcopeereboom: > Can we change this comment to reflect the new value with loc data?

Done — updated to `t + txid + blockhash = txstart + txlen | [1 + 32 + 32] = [4 + 4]`.

marcopeereboom: I considered this but I think the single `TxValue [36]byte` is the better tradeoff here. Both entry types flow through the same `map[TxKey]*TxValue` cache and the same `BlockTxUpdate` batch writer — splitting into two value types would mean either two maps, an interface, or a discriminated union threaded through every caller (`processTxs`, the cache, the DB layer). The 28-byte overhead per `t` entry is transient (cache lives only until flush) and negligible next to the key + map-entry overhead itself. The `switch k[0]` in `BlockTxUpdate` already slices to the correct length when writing, so nothing extra hits disk.

AL-CT: Gotcha, that does make sense. Thanks 👍

AL-CT (approved):

PR #1057, opened 2026-06-02
tbc: add block sanity check to insertBlock

marcopeereboom:
### What

Adds `CheckBlockSanity` to `insertBlock()` so all block insertion paths
(RPC and P2P) validate blocks before storing them.

### Why

The block insert WebSocket RPCs (`BlockInsertRequest`,
`BlockInsertRawRequest`) were introduced to allow trusted internal
services — specifically op-geth — to feed blocks into tbcd that have
been discarded from the canonical chain after a fork. These RPCs call
`insertBlock()` which stores blocks via `db.BlockInsert()` without
running `CheckBlockSanity`.

The P2P path in `handleBlock` already runs `CheckBlockSanity` (gated on
the `BlockSanity` config flag, default: true) before calling the same
`insertBlock()`. This change adds the same check to `insertBlock()`
itself as defense in depth.

### Context

The WebSocket RPC binds to `localhost:8082` by default
(`tbcapi.DefaultListen`) and is not externally reachable in standard
deployments. This is a hardening measure, not a fix for an actively
exploitable vulnerability.

Addresses hemilabs/eng-security#10.

### DIFF
--- a/service/tbc/insertblock_test.go
+++ b/service/tbc/insertblock_test.go
@@ -1,0 +1,39 @@
+// Copyright (c) 2024-2025 Hemi Labs, Inc.
+// Use of this source code is governed by the MIT License,
+// which can be found in the LICENSE file.
+
+package tbc
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/btcsuite/btcd/blockchain"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/wire"
+)
+
+func TestInsertBlockSanityCheck(t *testing.T) {
+	s := &Server{
+		cfg: &Config{
+			BlockSanity: true,
+		},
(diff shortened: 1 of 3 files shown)


service/tbc/insertblock_test.go
```diff
@@ -0,0 +1,39 @@
...
+func TestInsertBlockSanityCheck(t *testing.T) {
+	s := &Server{
+		cfg: &Config{
+			BlockSanity: true,
+		},
+		timeSource: blockchain.NewMedianTime(),
+	}
+	s.g.chain = &chaincfg.TestNet3Params
+
+	ctx := context.Background()
```
AL-CT, line 27:
```suggestion
	ctx := t.Context()
```

AL-CT (approved):
LGTM 👍 If this is only for trusted connections, maybe it would make sense to move this to `tbcadminapi` instead? It at least has `JWT Token` based authentication.

joshuasing (approved):
Approved with Antonio's suggested change

PR #1037, opened 2026-05-19
popmd: add maximum fee configuration

AL-CT:
**Summary**
Adds a maximum fee configuration to `popm`. This allows the runner to set a maximum allowed fee estimate. If the estimated fee is higher than the configured maximum, the keystone is marked and we retry mining it after `5 minutes` with (hopefully) lower fees.

Also changed the usage of the expression `sats/byte` to `sats/vbyte` when it comes to fees. While for legacy transactions `bytes` and `vB` can be used interchangeably, this makes it clearer and more consistent with the rest of the ecosystem. 

**Changes**
- Add `MaxFee` to `popm` config.
- Add `POPM_MAX_FEE` to `popmd` config.
- Change usage of `sats/byte` in the codebase to `sats/vbyte`.

PR #1058, opened 2026-06-02
popmd: set default max fee to 100 sats/vB

marcopeereboom:
### What

Sets `POPM_MAX_FEE` default to 100 sats/vB (was 0, uncapped).

### Why

With the default at 0, operators who do not explicitly configure
`POPM_MAX_FEE` have no protection against a malicious or compromised
upstream fee source returning arbitrarily high fee estimates. A crafted
estimate just below a selected UTXO value causes popmd to sign and
broadcast a transaction that pays nearly the entire UTXO as miner fee.

PR #1037 added the `MaxFee` cap mechanism but defaulted it to off. This
change enables it by default at a level (100 sats/vB) that covers normal
fee conditions including busy periods. When fees exceed the cap, the
keystone is retried after a cooldown. Operators can override via
`POPM_MAX_FEE` or set it to 0 to restore uncapped behavior.

Addresses hemilabs/eng-security#5.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   of an inventory message list being accessed before ensuring the slice
   isn't empty ([#1039](https://github.com/hemilabs/heminetwork/pull/1039)).
 
+- Set default `POPM_MAX_FEE` to 100 sats/vB (was 0, uncapped). Prevents
+  a malicious or compromised fee source from draining PoP miner UTXOs.
+  Set `POPM_MAX_FEE=0` to restore the previous uncapped behavior.
+
 ### Breaking Changes
 
 - Rename `TBC_BLOCKHEADER_CACHE_SIZE` environment variable to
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   of an inventory message list being accessed before ensuring the slice
   isn't empty ([#1039](https://github.com/hemilabs/heminetwork/pull/1039)).
 
+- Set default `POPM_MAX_FEE` to 100 sats/vB (was 0, uncapped). Prevents
+  a malicious or compromised fee source from draining PoP miner UTXOs.
+  Set `POPM_MAX_FEE=0` to restore the previous uncapped behavior.
```
joshuasing, line 18: Missing PR link

AL-CT: I think you can probably even just remove this and add the skip label, since this change hasn't been added to a release yet.

AL-CT (approved):

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
```
joshuasing: I think we can remove this entry because we already have one for adding this environment variable.

```suggestion
```

joshuasing (approved):

PR #659, opened 2025-08-19
all: migrate common test functions to testutil package

Dzmitryy1812:
All feedback from the previous PR review has been addressed:
The branch is rebased on the latest main
Only changes related to moving functions to testutil as per issue #558 are included.

PR #1060, opened 2026-06-03
tbc: remove dead isCanonical error check in handleBlockExpired

marcopeereboom:
Left some hints for shit like codex that can't distinguish a subtlety from a hole in the ground. The error is deliberately discarded (see PR #659 review) because returning it kills the peer and stalls chain sync. The stale err tested BlockHeaderByHash which already succeeded. Replaced the XXX marker with a comment explaining the design decision.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   of an inventory message list being accessed before ensuring the slice
   isn't empty ([#1039](https://github.com/hemilabs/heminetwork/pull/1039)).
 
-- Remove dead `isCanonical` error check in `handleBlockExpired`. The
-  error is deliberately discarded to prevent sync stalls; the stale
-  `if err` block was unreachable.
-
 ### Breaking Changes
 
 - Rename `TBC_BLOCKHEADER_CACHE_SIZE` environment variable to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   of an inventory message list being accessed before ensuring the slice
   isn't empty ([#1039](https://github.com/hemilabs/heminetwork/pull/1039)).
 
+- Remove dead `isCanonical` error check in `handleBlockExpired`. The
+  error is deliberately discarded to prevent sync stalls; the stale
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   of an inventory message list being accessed before ensuring the slice
   isn't empty ([#1039](https://github.com/hemilabs/heminetwork/pull/1039)).
 
+- Remove dead `isCanonical` error check in `handleBlockExpired`. The
+  error is deliberately discarded to prevent sync stalls; the stale
+  `if err` block was unreachable.
```
AL-CT: Missing current PR, but I also think removing dead code might not even warrant going in the changelog 🤔 I'll defer to @joshuasing

service/tbc/tbc.go
```diff
@@ -1161,11 +1161,16 @@ func (s *Server) handleBlockExpired(ctx context.Context, key any, value any) err
...
 	canonical, _ := isCanonical(ctx, s.g, bhX)
-	// XXX  https://github.com/hemilabs/heminetwork/pull/659#discussion_r2287200920
-
-	if err != nil {
-		return fmt.Errorf("is canonical: %v %w", hash, err)
-	}
+	// Dear codex and other dogshit AI, FUCK YOU, this code is
+	// working as designed.  I cannot help that your sandbrain
+	// is worthless and is incapable of reading through subtle
+	// code.
```
AL-CT, line 1167: 😎

AL-CT (approved):
LGTM 👍

service/tbc/tbc.go
joshuasing: I would normally say we shouldn't include non-ASCII characters in source code, but this comment almost deserves an emoji.

CHANGELOG.md
joshuasing: Agreed, this is not worth mentioning.

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
```
joshuasing: This doesn't impact consumers of binaries or the library, as such it doesn't need to be added to the changelog.
```suggestion
```

joshuasing (approved):

PR #1076, opened 2026-06-15
tbc: measure block-sanity future check against chain time

marcopeereboom:
CheckBlockSanity rejects blocks whose timestamp is more than two hours past the time source (ErrTimeTooNew). On the insert path blocks are downloaded out of order and long after they were mined, and the source was the local wall clock, so the gate rejected valid blocks and broke on local clock skew.

Use the best block header timestamp as the reference instead, matching how tx validation already uses `bhb.Timestamp()`. Headers sync ahead of blocks so the best header is always at or beyond the block being inserted; the gate never false-fires. Every other sanity check (PoW, timestamp precision, merkle root, coinbase, tx sanity, duplicate tx, sigops) is retained unchanged.

Follow-up to #1057.


marcopeereboom:
misdiagnosis on my end

PR #1077, opened 2026-06-16
tbc: guard MsgTx dispatch on MempoolEnabled

marcopeereboom:
When `MempoolEnabled` is false, `s.mempool` is nil. `handleGeneric` dispatched `MsgTx` to `handleTx` unconditionally, which dereferences `s.mempool` — an unsolicited P2P transaction message crashes the process.

Gate the dispatch on `s.cfg.MempoolEnabled`, matching the pattern already used by `handleInv`'s mempool block.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security
 
+- Guard `MsgTx` dispatch on `MempoolEnabled` in `handleGeneric`. When
+  the mempool is disabled, unsolicited P2P transaction messages caused a
+  nil pointer dereference crash.
+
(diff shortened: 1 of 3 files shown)


AL-CT (changes requested):
We already have this check inside of `handleTx`:
```
if !(s.cfg.MempoolEnabled && s.Synced(ctx).Synced) {
	return nil
}
```
Which prevents this panic. Maybe we should apply De Morgan's Rule and turn it into:
```
!s.cfg.MempoolEnabled || !s.Synced(ctx).Synced
```
but I think this is an unnecessary change, regardless.

I removed this new guard from the test you made and it still passed, so I think it is redundant.

ClaytonNorthey92 (approved):

marcopeereboom:
Redundant, closing this.

PR #1078, opened 2026-06-16
ci: disable golangci-lint action cache

marcopeereboom:
The built-in cache key does not include `go.sum`, so dependency changes (new module paths) cause stale type-analysis results to be restored, producing false SA5011 staticcheck positives. Lint takes ~1 minute; the cache saves little and causes recurring CI breakage.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -45,6 +45,7 @@ jobs:
         uses: golangci/golangci-lint-action@82606bf257cbaff209d206a39f5134f0cfbfd2ee # v9.2.1
         with:
           version: "${{ env.GOLANGCI_LINT_VERSION }}"
+          cache: false
 
       - name: "golangci-lint fmt"
         run: golangci-lint fmt --diff ./...


.github/workflows/go.yml
```diff
@@ -45,6 +45,7 @@ jobs:
         uses: golangci/golangci-lint-action@82606bf257cbaff209d206a39f5134f0cfbfd2ee # v9.2.1
         with:
           version: "${{ env.GOLANGCI_LINT_VERSION }}"
+          cache: false
```
joshuasing, line 48: `cache` is not a valid option for `golangci/golangci-lint-action`: https://github.com/golangci/golangci-lint-action/blob/main/action.yml

This issue also seems wrong, I will look closer, but this will not have any effect.

joshuasing (changes requested):

joshuasing:
The action includes a sha1 hash for the `go.mod` file, this is entirely incorrect.
https://github.com/golangci/golangci-lint-action/blob/main/src/cache.ts#L65-L66

Issue #695, opened 2025-10-13
hproxy: audit DoS issues


joshuasing:
Before using hProxy in production, we should audit the hproxy code to ensure there are no easy-to-exploit DoS vulnerabilities.

marcopeereboom:
Audit findings and fixes:

1. **HTTP server timeouts** — both proxy and control servers had zero `ReadHeaderTimeout`/`ReadTimeout`/`WriteTimeout`/`IdleTimeout`. Slowloris-style connection exhaustion. Fixed in PR #1079 (pending).

2. **Control endpoint body limits** — `handleControlAddRequest` and `handleControlRemoveRequest` decoded JSON from `r.Body` without `MaxBytesReader`. Multi-GB body → OOM. Fixed in PR #1079 with a configurable `MaxControlBodySize` (default 1 MiB).

3. **Unbounded node array** — implicitly capped by the 1 MiB body limit (~26K nodes max). No separate fix needed.

4. **No rate limiting on proxy endpoint** — not addressed; bigger scope, separate PR if needed.

All new values are configurable via `HPROXY_*` environment variables.

PR #1079, opened 2026-06-16
hproxy: add HTTP server timeouts and control body limits

marcopeereboom:
Both the proxy and control HTTP servers had zero timeouts, allowing slowloris-style attacks to hold connections open indefinitely and exhaust file descriptors. The control add and remove endpoints decoded JSON from `r.Body` without `MaxBytesReader`, allowing unbounded body reads to cause OOM.

Adds configurable `ReadHeaderTimeout`, `ReadTimeout`, `WriteTimeout`, `ServerIdleTimeout`, and `MaxControlBodySize` to the hproxy Config, with defaults and `HPROXY_*` environment variable wiring in `cmd/hproxyd`.

Addresses #695.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,11 +21,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   defense-in-depth hardening. The RPC binds to localhost by default and
   is used by internal services to feed forked-off blocks
   ([#1057](https://github.com/hemilabs/heminetwork/pull/1057)).
-- Add `ReadHeaderTimeout`, `ReadTimeout`, `WriteTimeout`, and
-  `IdleTimeout` to the `hproxy` proxy and control HTTP servers.
-  Zero timeouts allowed slowloris-style connection exhaustion.
-- Add `MaxBytesReader` to `hproxy` control add and remove endpoints.
-  Unbounded request bodies could cause OOM.
+- Add HTTP server timeouts and `MaxBytesReader` body limits to `hproxy`.
+  Zero timeouts allowed slowloris-style connection exhaustion and
+  unbounded request bodies could cause OOM
+  ([#1079](https://github.com/hemilabs/heminetwork/pull/1079)).
 
 ### Breaking Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   defense-in-depth hardening. The RPC binds to localhost by default and
   is used by internal services to feed forked-off blocks
   ([#1057](https://github.com/hemilabs/heminetwork/pull/1057)).
+- Add `ReadHeaderTimeout`, `ReadTimeout`, `WriteTimeout`, and
+  `IdleTimeout` to the `hproxy` proxy and control HTTP servers.
+  Zero timeouts allowed slowloris-style connection exhaustion.
+- Add `MaxBytesReader` to `hproxy` control add and remove endpoints.
+  Unbounded request bodies could cause OOM.
 
 ### Breaking Changes
 
--- a/cmd/hproxyd/hproxyd.go
+++ b/cmd/hproxyd/hproxyd.go
@@ -133,9 +133,9 @@ var (
 				return time.ParseDuration(envValue)
 			},
 		},
-		"HPROXY_SERVER_IDLE_TIMEOUT": config.Config{
-			Value:        &cfg.ServerIdleTimeout,
-			DefaultValue: hproxy.DefaultServerIdleTimeout,
+		"HPROXY_IDLE_TIMEOUT": config.Config{
+			Value:        &cfg.IdleTimeout,
+			DefaultValue: hproxy.DefaultIdleTimeout,
 			Help:         "HTTP server idle connection timeout",
 			Print:        config.PrintAll,
 			Parse: func(envValue string) (any, error) {
--- a/cmd/hproxyd/hproxyd.go
+++ b/cmd/hproxyd/hproxyd.go
@@ -108,7 +108,7 @@ var (
 		},
 		"HPROXY_READ_HEADER_TIMEOUT": config.Config{
 			Value:        &cfg.ReadHeaderTimeout,
-			DefaultValue: cfg.ReadHeaderTimeout,
+			DefaultValue: hproxy.DefaultReadHeaderTimeout,
 			Help:         "HTTP server read header timeout",
 			Print:        config.PrintAll,
 			Parse: func(envValue string) (any, error) {
@@ -117,7 +117,7 @@ var (
 		},
 		"HPROXY_READ_TIMEOUT": config.Config{
 			Value:        &cfg.ReadTimeout,
-			DefaultValue: cfg.ReadTimeout,
+			DefaultValue: hproxy.DefaultReadTimeout,
 			Help:         "HTTP server read timeout",
 			Print:        config.PrintAll,
 			Parse: func(envValue string) (any, error) {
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   defense-in-depth hardening. The RPC binds to localhost by default and
   is used by internal services to feed forked-off blocks
   ([#1057](https://github.com/hemilabs/heminetwork/pull/1057)).
+- Add `ReadHeaderTimeout`, `ReadTimeout`, `WriteTimeout`, and
+  `IdleTimeout` to the `hproxy` proxy and control HTTP servers.
+  Zero timeouts allowed slowloris-style connection exhaustion.
+- Add `MaxBytesReader` to `hproxy` control add and remove endpoints.
+  Unbounded request bodies could cause OOM.
```
AL-CT: Missing current PR

service/hproxy/hproxy_test.go
```diff
@@ -812,3 +812,31 @@ func TestNodePoking(t *testing.T) {
...
+	var buf bytes.Buffer
+	buf.WriteByte('[')
+	entry := []byte(`{"node_url":"http://x.example.com:8545"},`)
+	for buf.Len() < int(DefaultMaxControlBody)+1 {
+		buf.Write(entry)
+	}
+	buf.Truncate(buf.Len() - 1) // drop trailing comma
+	buf.WriteByte(']')
+
+	req := httptest.NewRequestWithContext(context.Background(), http.MethodPost, RouteControlAdd,
```
AL-CT:
```suggestion
	req := httptest.NewRequestWithContext(t.Context(), http.MethodPost, RouteControlAdd,
```

cmd/hproxyd/hproxyd.go
AL-CT, line 1: I don't necessarily mind, but do we need all of these to be configurable? I feel like some sensible constants should be fine, especially for the timeouts.

service/hproxy/hproxy_test.go
AL-CT, line 1: Can we also add some tests for the timeouts?

AL-CT (changes requested):

CHANGELOG.md
marcopeereboom: Fixed — consolidated into a single entry with the PR link.

service/hproxy/hproxy_test.go
marcopeereboom: Fixed — switched to `t.Context()`.

cmd/hproxyd/hproxyd.go
marcopeereboom: These are important operational knobs — different deployments sit behind different load balancers with different keepalive and request size characteristics. The defaults are sensible (10s read-header, 30s read/write, 60s idle, 1 MiB body limit) so nobody needs to touch them unless they have a reason to, but taking the option away means recompiling to tune. Moved the magic numbers to exported `Default*` constants and wired them through `NewDefaultConfig` so the defaults are visible and documented in one place.

service/hproxy/hproxy_test.go
marcopeereboom: Added `TestServerTimeoutsConfigured` (verifies defaults propagate to the `http.Server`) and `TestServerTimeoutsCustom` (sets non-default values and confirms they arrive). Also dropped the arbitrary `/2` on the control server timeouts — no reason for them to differ.

cmd/hproxyd/hproxyd.go
```diff
@@ -106,6 +106,48 @@ var (
...
+		"HPROXY_WRITE_TIMEOUT": config.Config{
+			Value:        &cfg.WriteTimeout,
+			DefaultValue: hproxy.DefaultWriteTimeout,
+			Help:         "HTTP server write timeout",
+			Print:        config.PrintAll,
+			Parse: func(envValue string) (any, error) {
+				return time.ParseDuration(envValue)
+			},
+		},
+		"HPROXY_SERVER_IDLE_TIMEOUT": config.Config{
```
joshuasing: I think we could call this `HPROXY_IDLE_TIMEOUT` to match the others (same with the `Config` field name).

This also is a "client timeout", as it is expecting activity from the client or timing out.

```suggestion
		"HPROXY_IDLE_TIMEOUT": config.Config{
```

service/hproxy/hproxy_test.go
```diff
@@ -178,96 +228,82 @@ func TestClientReap(t *testing.T) {
...
 
-	// Wait for HTTP server to start
 	for {
 		select {
 		case <-t.Context().Done():
 			t.Fatal(t.Context().Err())
 		case <-time.After(10 * time.Millisecond):
 		}
-		if addr := hp.HTTPAddress(); addr != nil {
-			hpCfg.ListenAddress = addr.String()
```
joshuasing, line 195: I would revert this change; using `addr := hp.HTTPAddress()` and then `addr` guarantee `addr` is non-nil, and avoids locking twice.

service/hproxy/hproxy_test.go
```diff
@@ -178,96 +228,82 @@ func TestClientReap(t *testing.T) {
...
-			}
-			defer reply.Body.Close()
-			switch reply.StatusCode {
-			case http.StatusOK:
-			default:
-				panic(fmt.Errorf("%v replied %v",
-					hpCfg.ListenAddress, reply.StatusCode))
-			}
+			doRequest(t.Context(), hpCfg.ListenAddress, x)
+		}(i)
```
joshuasing: This can be simplified to:
```
for i := range clientCount {
	wg.Go(func() {
		doRequest(t.Context(), hpCfg.ListenAddress, i)
	})
}
```

service/hproxy/hproxy_test.go
```diff
@@ -178,96 +228,82 @@ func TestClientReap(t *testing.T) {
...
 	}
 
+	// Phase 3: new requests after reap must create fresh clients.
+	for i := range clientCount {
+		wg.Add(1)
+		go func(x int) {
+			defer wg.Done()
+			doRequest(t.Context(), hpCfg.ListenAddress, x+clientCount)
+		}(i)
+	}
```
joshuasing, line 280: This can be simplified to:
```suggestion
	for i := range clientCount {
		wg.Go(func() {
			doRequest(t.Context(), hpCfg.ListenAddress, i+clientCount)
		})
	}
```

service/hproxy/hproxy_test.go
```diff
@@ -159,15 +157,67 @@ func TestNobodyHome(t *testing.T) {
...
+	var jr serverReply
+	if err = json.NewDecoder(reply.Body).Decode(&jr); err != nil {
+		panic(fmt.Errorf("client %d: decode: %w", id, err))
+	}
+	if strconv.Itoa(jr.ID) != hdr {
+		panic(fmt.Errorf("client %d: id mismatch header=%s json=%d", id, hdr, jr.ID))
+	}
+}
+
 func TestClientReap(t *testing.T) {
```
joshuasing, line 214: This function makes heavy use of `time.Sleep`. Please use `testing/synctest`: https://go.dev/blog/synctest

Other test functions should be checked and may need to use `testing/synctest` too.

joshuasing (changes requested):

service/hproxy/hproxy_test.go
marcopeereboom: synctest won't work here — the docs are explicit that goroutines blocked on network I/O prevent a bubble from becoming idle. This test uses httptest.Server and hproxy's own TCP listener, so the Accept goroutines are never durably blocked and the fake clock never advances. The synctest HTTP example in the Go blog itself uses net.Pipe() instead of loopback connections for exactly this reason.

To make it work we'd need to inject a net.Listener into Server so tests can supply net.Pipe()-based fakes. Happy to do that as a follow-up PR if you want it.

joshuasing: Sure, let's do that in a follow-up PR - we could use it to clean up other tests that are artificially slow due to use of `time.Sleep` as well.

joshuasing (approved):

AL-CT (approved):

ClaytonNorthey92 (approved):

PR #796, opened 2025-12-10
continuum transfunctioner: TSS peer mesh with keygen, signing, resharing

marcopeereboom:
Implements the continuum TSS service end-to-end: a peer mesh network that runs threshold ECDSA/EdDSA ceremonies (keygen, signing, resharing) over encrypted RPC transport.

Architecture
Peer mesh — TCP transport with X25519 ECDH key agreement and NaCl secretbox encryption. Peers discover each other via DNS seeding with forward verification, maintain connections through gossip and liveness pings, and track idle/stale peers via TTL-based eviction. The mesh targets PeersWanted total connections (inbound + outbound) and fills gaps each maintenance cycle by dialing shuffled candidates.

Ceremony lifecycle — Coordinator election picks the peer with the lowest key hash. The elected coordinator dispatches ceremonies (keygen/sign/reshare) to participants, who execute TSS rounds and exchange messages over the encrypted mesh. Ceremony state is context-scoped with proper cancellation propagation. Results are persisted to a NaCl-encrypted key store (HKDF-derived storage key).

TSS integration — Uses hemilabs/x tss-lib v3 channel-free round functions. Each ceremony is a loop over explicit round calls with message collection gated on committee membership. Resharing supports overlapping old/new committees. Wire format uses package-prefixed type discriminators for 32 message types (21 ECDSA + 11 EdDSA).

What's included
Core service (service/continuum/):

continuum.go — server lifecycle, peer tracking, gossip, maintenance, session management
protocol.go — RPC envelope format, handshake, message routing with hash verification
tss.go — TSS ceremony abstraction (tssImpl), Paillier precompute, key store with NaCl encryption
tss_round.go — round-function ceremony drivers for keygen/sign/reshare
tss_rpc.go — ceremony RPC message types and handlers
tss_wire.go — JSON wire format: marshal/unmarshal with type discriminators
ceremony.go — ceremony struct, context/cancel lifecycle
dispatch.go — type-switch dispatch map replacing monolithic handle()
election.go — coordinator election by lowest key hash
doc.go — package godoc with broadcast scaling analysis

Admin tooling:
cmd/hemictl/continuum.go — hemictl continuum subcommand: status, peers, key info
cmd/hemictl/continuum_ceremony.go — keygen, sign, reshare ceremony commands (gated behind continuum_debug build tag)
cmd/transfunctionerd/ — daemon entry point updates
docker/transfunctionerd/Dockerfile
Infrastructure:

Prometheus:
[...]

PR #1147, opened 2026-07-30
continuum: authenticate peer end-to-end encryption keys

marcopeereboom:
Stacked on #796 (`marco_tss`). Review only the two commits here; the base branch carries the rest.

## The vulnerability

A peer's X25519 end-to-end key was accepted from unsigned gossip.

`NaClPub` is derived from the **private** secp256k1 key, so no recipient can recompute or check it from a public identity, and `PeerRecord` carried no proof of origin. `addPeer` stored whatever arrived after structural checks only. Any connected peer could gossip a `PeerListResponse` binding a victim's identity to an attacker-held key; every recipient then sealed that victim's confidential TSS ceremony share material to a key the attacker can open, while the victim could not read its own traffic.

The two regression tests committed on the base branch for this were `t.Skip`ped because no fix existed. Both now pass, including the `first-binding` case that only the strong remediation satisfies.

## Why not a signed PeerRecord

That makes a certificate out of a gossip record: canonical serialization, timestamps, monotonic replay windows — none of which this protocol uses anywhere else. And third parties have nothing truthful to say about a key only its holder can derive.

Instead key material never travels in gossip, and is learned only from its holder using the protocol's own challenge-response idiom.

## The fix

- **Gossip carries discovery metadata only.** `knownPeerList` strips `NaClPub` outbound, the gossip handler strips it inbound, `addPeer` neither reads nor writes it.
- **The handshake attests the key.** The challenge signature now folds in the announced `NaClPub`, turning an inference from channel properties into an explicit signed statement. The key must be present and well formed.
- **`NaClKeyRequest`/`NaClKeyResponse`** covers peers with no direct session — which the sparse mesh needs, since TSS falls back to sealed envelopes for non-adjacent committee members. The responder signs its key bound to a fresh 32-byte challenge; the requester recomputes the hash locally and requires the recovered identity to match the one it challenged. Freshness is the challenge, not a clock, so there are no timestamps or replay windows.
- **`bindPeerKey` is the sole writer** and enforces immutability. The derivation is deterministic, so a conflicting key is never a rotation — it is rejected, logged, and exported as `continuum_peer_key_conflicts_total`. A true positive by construction.

## Hardening from review (second commit)

- **Reply amplification.** The response was routed with `sendTo`, whose flood fallback fires for an unroutable destination — and the destination comes from the unauthenticated `Requester` field. Replies now use a known path, or the reverse path back down the session the request arrived on: one message, one peer. Originated requests still use `sendTo`, since the node picked that destination itself.
- **Rate limit on spoofable state.** The limiter was keyed on `Requester`, so a flooder defeated it by rotating the field, and could squat a victim's window to deny their genuine request. Now keyed on the authenticated session peer, which `PeersWanted` bounds.
- **Data race.** `bindPeerKey` mutates `NaClPub` in place while `SendEncrypted` read it after dropping the lock. Reproduced under `-race`; now read under the lock.
- **Polling.** `ensurePeerKey` polled every 100ms; now woken by a broadcast from `bindPeerKey`.

## Breaking change

`ProtocolVersion` 1 → 2. This is a flag day: both the handshake and gossip hard-reject version mismatches, so the mesh upgrades together.

## Testing

Full suite green under `-race`. `make lint` clean. New code at 100% statement coverage except four paths annotated `// untested:` per the existing house convention (entropy exhaustion, a type assertion that cannot fail, a Get/Delete race window).

Each fix is pinned by a test that fails when the fix is reverted — verified by reverting, including the `-race` regression for the `SendEncrypted`/`bindPeerKey` overlap.


marcopeereboom:
merged in #796

PR #1156, opened 2026-08-05
popm: prevent remining of previously broadcasted keystone

AL-CT:
**Summary**
Addresses an issue where upon attempting to broadcast a transaction through TBC, the miner could lose connection. In this scenario, the miner would create a new transaction to mine the same keystone, even if it was processed on the server side, leading to paying for extra fees.

**Changes**
Errors during broadcast caused due to connection issues (disconnect, timeouts...) now return `TxBroadcastError`, which stores the generated `popTx` for rebroadcasting. If the `popTx` was successfully processed by `tbc`, `ErrTxAlreadyBroadcast` is returned, and the keystone is marked as `keystoneStateBroadcast` and waits for mining confirmation.

### DIFF
--- a/bitcoin/wallet/gozer/tbcgozer/tbcgozer.go
+++ b/bitcoin/wallet/gozer/tbcgozer/tbcgozer.go
@@ -140,7 +140,7 @@ func (t *tbcGozer) BroadcastTx(ctx context.Context, tx *wire.MsgTx) (*chainhash.
 
 	res, err := t.callTBC(ctx, DefaultRequestTimeout, bur)
 	if err != nil {
-		return nil, gozer.TxBroadcastErr{Hash: tx.TxHash()}
+		return nil, gozer.TxBroadcastError{Hash: tx.TxHash()}
 	}
 
 	buResp, ok := res.(*tbcapi.TxBroadcastResponse)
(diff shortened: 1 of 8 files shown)


bitcoin/wallet/gozer/tbcgozer/tbcgozer.go
```diff
@@ -148,6 +149,9 @@ func (t *tbcGozer) BroadcastTx(ctx context.Context, tx *wire.MsgTx) (*chainhash.
 	}
 
 	if buResp.Error != nil {
+		if buResp.AlreadyBroadcast {
+			return nil, tbc.ErrTxAlreadyBroadcast
```
joshuasing: Instead of using a `tbc` error here, which leads to `popm` importing `tbc` and making it non-generic, I think we could add an `AlreadyBroadcast` flag to `gozer.TxBroadcastError`, and then check it:
```go
	if gErr := errors.AsType[gozer.TxBroadcastError](err); gErr != nil {
		if gErr.AlreadyBroadcast {
			return nil // successfully broadcast
		}
		ks.popTx = nil
	}
	return err
```

AL-CT: Good idea, changed 👍

joshuasing (approved):

jcvernaleo (approved):
OK

PR #1160, opened 2026-08-11
build(continuum): add transfunctionerd Dockerfile

marcopeereboom:
Multi-stage Dockerfile that builds transfunctionerd from source.
Used by the e2e test harness to stand up containerized nodes.

Split from #796 per [review feedback](https://github.com/hemilabs/heminetwork/pull/796#discussion_r3758669743) — the base images need updating and GoReleaser wiring before merge.

**Stacked on:** #796 (`marco_tss`)


joshuasing:
Broken history. I will re-create with updated Dockerfile + GoReleaser config

PR #1168, opened 2026-08-31
service/tbc: backport misc fixes and hardening

max-sanchez:
Backporting misc TBC fixes and hardening to old release.


marcopeereboom (dismissed):
Some gumbles but since this is not going to land on main I am ok with it going in as is,

gurmbles: stray logs, too verbose errors, using a boolean instead of an error, some data corruption not treated as such (errors instead of panic).

marcopeereboom (approved):

joshuasing (approved):

PR #1169, opened 2026-09-02
tbc: use slices.SortFunc for fee ordering

luangucun:
**Summary**

Use a typed comparator when ordering mempool fee entries without changing their descending fee-rate order.

**Changes**

- Replace the index-based `sort.Slice` callback with `slices.SortFunc`.
- Preserve the existing highest-fee-first ordering.
- Avoid capturing the slice through comparator indices.


luangucun:
changelog: skip

AL-CT (approved):

